PR #6, opened 2023-03-17
Adding monai tool seg  model and SSD model

jin-nvidia:


applications/monai_endoscopic_tool_seg/scripts/graph_surgeon_tool_seg.py
jjomier, line 1: @jin-nvidia This file is missing the standard copyright header with Apache 2 license.

jin-nvidia: Just added header to this file

PR #15, opened 2023-04-05
Update SSD app and MONAI app for v0.5

jin-nvidia:
Updated metadata for both apps and the Dockerfile for the SSD app


jjomier (approved):
That looks good.
It is strange that you get the full history  of all the commits. Might be easier to  start a new branch from main next time.

jin-nvidia:
> 

Noted, will do that for the future

PR #19, opened 2023-04-13
Refactoring of V4L2 extension/operators/apps

brudfors:
* Renamed v4l2_plus app to usb_hdmi_video_capture
* Renamed GXF extension files+folder from v4l2_plus_source to v4l2_video_capture
* Renamed operators files+folders from v4l2_plus_source to v4l2_video_capture
* Renamed GXF extension class from V4L2PlusSource to V4L2VideoCapture
* Renamed operator class from V4L2PlusSourceOp to V4L2VideoCaptureOp
* Remove ffmpeg from README
* Replace UnboundedAllocator with BlockMemoryPool for better FPS: HDMI IN now runs on 60 Hz
* Use four V4L2 buffers by default (gives >FPS for HDMI IN)
* Holoviz uses resolution of video source
* Use HDMI IN by default, and show commented example of using USB
* Add CMakeLists.txt for Python app

### DIFF
--- a/applications/usb_hdmi_video_capture/cpp/main.cpp
+++ b/applications/usb_hdmi_video_capture/cpp/main.cpp
@@ -18,23 +18,37 @@
 #include <holoscan/holoscan.hpp>
 #include <v4l2_video_capture.hpp>
 #include <holoscan/operators/holoviz/holoviz.hpp>
-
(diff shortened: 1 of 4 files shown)


agirault:
Why removing the v4l2 name?

PR #29, opened 2023-05-01
Added pulses/sec metric

cliffburdick:
Added pulses/sec metric to cpp and changed metric in Python

### DIFF
--- a/applications/simple_radar_pipeline/python/simple_radar_pipeline.py
+++ b/applications/simple_radar_pipeline/python/simple_radar_pipeline.py
@@ -237,4 +237,4 @@ if __name__ == "__main__":
 
     duration = (iterations * num_pulses * num_channels) / (tstop - tstart)
 
-    print(f"{duration:0.3f} pulses/channel/sec")
+    print(f"{duration:0.3f} pulses/sec")
--- a/applications/simple_radar_pipeline/cpp/README.md
+++ b/applications/simple_radar_pipeline/cpp/README.md
@@ -8,7 +8,7 @@ This demonstration walks the developer through building a simple radar signal pr
 
 While this example generates 'offline' complex-valued data, it could be extended to accept streaming data from a phased array system or simulation via modification of the `SignalGeneratorOperator`.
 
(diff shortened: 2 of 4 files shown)


applications/simple_radar_pipeline/python/simple_radar_pipeline.py
```diff
@@ -237,4 +237,4 @@ def compose(self):
 
     duration = (iterations * num_pulses * num_channels) / (tstop - tstart)
 
-    print(f"{duration:0.3f} pulses/channel/sec")
+    print(f"{duration:0.3f} pulses/sec")
```
MaggieXJZhang, line 240: The readme says "The output of this demonstration is a measure of the number of pulses per channel per second processed on GPU"

Why is the per channel removed?

cliffburdick: Updated the readme. I wanted to show aggregate since the throughput is much higher. The previous printout was a bug since it was not per channel.

wendell-hom:
Cliff, could you help to fix the lint errors?

cliffburdick:
> Cliff, could you help to fix the lint errors?

Where can I see those? Is there a clang-tidy file?

wendell-hom:
> Where can I see those? Is there a clang-tidy file?
From the root directory, you can run `./run lint`.  
If needed, you can run `./run install_lint_deps` to install the necessary tooling

cliffburdick:
Thanks @wendell-hom . Can these be put into a clang-tidy file or do they have to be manually fixed?

cliffburdick:
These should all be addressed

applications/simple_radar_pipeline/python/simple_radar_pipeline.py
MaggieXJZhang: Is this only done for python? I still see the ` 35648 pulses/channel/sec` in the cpp version.

cliffburdick: This should be fixed now. Thanks for catching that!

PR #40, opened 2023-06-09
Add the support of YUAN capture card

DavidSu-Yuan:
I add a new gxf extension and operator to support our capture card (YUAN).
Some applications have modify to let developer can switch to our capture card as AJA card.

Those patch still need our SDK and driver to work (capture video from our card).
But don't cause build break or crash when user select to use AJA card or video replayer.
Even select YUAN card without our SDK and driver, applications still running.
QCAP_SOURCE (our gxf extension) will send out a "No QCAP SDK" screen to hint user.

And we will provide more document on how to use our card and how to get the support from us.


jjomier:
Refactored with the MR: https://github.com/nvidia-holoscan/holohub/pull/105

PR #41, opened 2023-06-23
Port fm_asr and holosar apps to Holohub

wendell-hom:
Port apps to public holohub


applications/fm_asr/metadata.json
```diff
@@ -0,0 +1,43 @@
+{
+	"application": {
+		"name": "FM-ASR",
+		"language": "Python",
```
jjomier, line 4: We need the authors

wendell-hom: thanks!  just updated

PR #46, opened 2023-07-13
Jjomier/fix metadata

jjomier:
Fixing metadata.json and added script to check the validity of the metadata


wendell-hom (approved):
Looks good!

PR #49, opened 2023-07-20
Add data directory and conversion scripts

danpcampbell:
Added data conversion scripts that did not make it into the initial release of the SAR application


wendell-hom (commented):
@danpcampbell please add copyright info to the top of the script files

wendell-hom (approved):

PR #64, opened 2023-08-23
Add installation of clang-format with lint_install_deps, add section …

wendell-hom:
May possibly want to add `run lint --fix` to fix lint issues that can automatically be fixed.

### DIFF
--- a/run
+++ b/run
@@ -260,7 +260,7 @@ install_lint_deps_desc() { echo 'Install lint dependencies
 This command will install the dependencies required to run the linting tools.
 
 - Python linting: ruff, isort, black, mypy
-- C++ linting: cpplint
+- C++ linting: cpplint, clang-format
 - Spelling: codespell
 '
 }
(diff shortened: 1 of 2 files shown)


jjomier:
@wendell-hom thanks for taking care of that! I agree if we can add the `./run lint --fix` that would be great. 
Can you add in the MR?

run
```diff
@@ -311,16 +311,62 @@ Spelling: codespell
...
+      else
+        DIR_TO_RUN=$arg
+      fi
+    done
+
     local exit_code=0
 
+    if [ $FIX == true ]; then
+      echo "Running linting fix on modified files in the repository"
+      modified_files=$(git diff --name-only)
```
wendell-hom, line 342: @jjomier git diff will only work on files that's currently modified in the local workspace, so it's going to miss files that have been staged, or that's been committed into the user's repo.

jjomier: @wendell-hom what's the best way to do it? check all the files?

wendell-hom: For now, we can just run the fixes for all the lint tools besides cpplint, e.g.,

    run_command ruff --fix --ignore E712 $DIR_TO_RUN
   
Afterwards we can deal with cpplint by running cpplint once to see which files it complain about and then use clang to fix each of those files.

wendell-hom: @jjomier I've just added the changes to do the above, can you please check?

jjomier: @wendell-hom great! seems to works well. I just added some more comments.

jjomier (approved):

run
wendell-hom: thanks @jjomier !

PR #65, opened 2023-08-23
Fix CMake policy warning for fetch_content timestamp

jjomier:


operators/volume_renderer/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ set(CLARA_VIZ_WITH_EXAMPLES OFF CACHE INTERNAL "")
 # we are downloading large content, it's better to see the progress
 set(FETCHCONTENT_QUIET OFF)
 
+# enable CMP0135 policy to use timestamp when fetching content
```
AndreasHeumann, line 29: Suggest
```
# Set CMP0135 policy to NEW to use time of extraction for files extracted by
# FetchContent/ExternalProject_Add.
# This ensures that anything that depends on the extracted contents will be rebuilt whenever
# the URL changes.
```

operators/volume_renderer/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ set(CLARA_VIZ_WITH_EXAMPLES OFF CACHE INTERNAL "")
...
+cmake_policy(SET CMP0135 NEW)
+
```
AndreasHeumann, line 31: I also noticed that there is a new warning with cmake 3.27
```
CMake Warning (dev) at build/_deps/claraviz-src/src/claraviz/CMakeLists.txt:100 (find_package):
  Policy CMP0146 is not set: The FindCUDA module is removed.  Run "cmake
  --help-policy CMP0146" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

This warning is for project developers.  Use -Wno-dev to suppress it.
```
Can be suppressed by adding
```
# set CMP0148 policy to OLD to allow FindCUDA to be used by ClaraViz
cmake_policy(SET CMP0146 OLD)
set(CMAKE_POLICY_DEFAULT_CMP0146 OLD)
```

operators/volume_renderer/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ set(CLARA_VIZ_WITH_EXAMPLES OFF CACHE INTERNAL "")
 # we are downloading large content, it's better to see the progress
 set(FETCHCONTENT_QUIET OFF)
 
+# enable CMP0135 policy to use timestamp when fetching content
+cmake_policy(SET CMP0135 NEW)
```
AndreasHeumann, line 30: I'm also seeing this `CMP0135` warning with the volume_loader op when fetching `nifti_clib`. Probably needs to be added there as well.

AndreasHeumann (changes requested):

operators/volume_renderer/CMakeLists.txt
jjomier: I don't see it, but added as well.

AndreasHeumann (approved):
Thanks, looks good to me!

PR #69, opened 2023-08-30
Enabling autocomplete for Holohub run command

jjomier:
Adding autocomplete for HoloHub `run` command


wendell-hom:
This is great!  Thanks for adding autocomplete

wendell-hom (approved):

PR #96, opened 2023-09-28
Update Local Llama llama.cpp

NigelNelson:
## Context
The Local Llama tutorial uses the [Llama.cpp](https://github.com/ggerganov/llama.cpp) repo, and previously [breaking changes](https://github.com/ggerganov/llama.cpp/pull/2398) were introduced. In these changes, the model format was changed from GGML -> GGUF. These changes are now stable and there is enough GGUF support that the tutorial should be updated to stay relevant.
## Description
Updates:
- Remove hack needed for ARM64 support (Updated llama.cpp includes this fix now)
- Correct all mentions and uses of GGML and use the GGUF equivalent.
- Update the `./server` output snippet

### DIFF
--- a/tutorials/local-llama/README.md
+++ b/tutorials/local-llama/README.md
@@ -33,34 +33,20 @@
 git clone https://github.com/ggerganov/llama.cpp.git
 ```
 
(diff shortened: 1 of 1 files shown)


jjomier (approved):
LGTM!

PR #102, opened 2023-10-05
GPUDirect TX

cliffburdick:
Adds GPUDirect support to transmit side


operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -57,22 +57,25 @@ PYBIND11_MODULE(_advanced_network_common, m) {
...
-    m.def("adv_net_set_cpu_udp_payload",
-        [](AdvNetBurstParams *burst, int idx, long int data, int len) {
-                return adv_net_set_cpu_udp_payload(burst, idx,
-                        reinterpret_cast<void*>(data), len); },
-                "Set UDP header parameters and copy payload");
-    m.def("adv_net_set_cpu_udp_payload",
-        [](std::shared_ptr<AdvNetBurstParams> burst, int idx, long int data, int len) {
-            return adv_net_set_cpu_udp_payload(burst, idx, reinterpret_cast<void*>(data), len); },
-            "Set UDP header parameters and copy payload");
+    // m.def("adv_net_set_cpu_udp_payload",
```
jjomier, line 69: Should these be removed instead of commented out?

cliffburdick: I left them in because they still exist, but the python bindings don't work yet due to an issue in Holoscan. This is just a placeholder until they're working.

PR #104, opened 2023-10-10
Refactor video capture

jjomier:


applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
```diff
@@ -21,18 +21,48 @@ extensions:
   - libgxf_serialization.so
   - libgxf_stream_playback.so
   - gxf_extensions/lstm_tensor_rt_inference/libgxf_lstm_tensor_rt_inference.so
+  - gxf_extensions/deltacast_videomaster/libgxf_videomaster.so
```
wendell-hom, line 24: @jjomier if user builds without using `--with deltacast_videomaster`, then it looks like we'll get some additional warnings when application starts:

 [warning] Unable to load extension from 'gxf_extensions/deltacast_videomaster/libgxf_videomaster.so' (error: gxf_extensions/deltacast_videomaster/libgxf_videomaster.so: cannot open shared object file: No such file or directory)

run
```diff
@@ -462,8 +463,13 @@ build() {
...
          echo "Using Holscan SDK in ${sdk_path}"
          skipnext=1
+      elif [ "$arg" = "--with" ]; then
+         operator="${ARGS[i+1]}"
+         configure_args="${configure_args} -DHOLOHUB_BUILD_OPERATORS=${operator}"
+         echo "Building with operator ${operator}"
+         skipnext=1
       elif [ "$arg" = "--configure-args" ]; then
-         configure_args="${ARGS[i+1]}"
+         configure_args="${configure_args} ${ARGS[i+1]}"
```
wendell-hom, line 472: Looks like this should now allow multiple --configure-args options to be used in the command line

applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
jjomier: That's the drawback of combining the applications. That said, you are right, let's comment this line out and document that it should be added when using Deltacast. I'll push a fix.

PR #112, opened 2023-10-18
Map OptiX denoiser weights into container.

AndreasHeumann:
To enable usage of OptiX denoiser by volume renderer and avoid error message `failed to create OptiX denoiser` when running the app.


agirault:
I filed this: https://github.com/NVIDIA/nvidia-container-toolkit/issues/127

PR #113, opened 2023-10-18
Removed aarch64 specific model

jjomier:


wendell-hom:
@jjomier what is the reason for this change?

jjomier:
> @jjomier what is the reason for this change?

We are not distributing the model anymore so that it's easier for user to generate and not have to worry about the naming of the model.

wendell-hom (approved):

PR #114, opened 2023-10-19
CUDA MPS Tutorial

sohamm17:
This MR adds a CUDA MPS tutorial to show how to enable CUDA MPS for concurrently running Holoscan applications, and points out a couple of benefits of using it.


tutorials/cuda_mps/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+For different applications, one may want to set different limits on the number of GPU threads
+available to each of them. This can be done by setting the `CUDA_MPS_ACTIVE_THREAD_PERCENTAGE`
+environment variable separately for each application. It is elaborated in details [here](https://docs.nvidia.com/deploy/mps/index.html#topic_5_2_5).
+
+There are other customizations available in CUDA MPS as well. Please refer to the CUDA MPS
+[documentation](https://docs.nvidia.com/deploy/mps/index.html#topic_5_1_1) to know more about them.
+
+## Performance Benefits
```
jjomier, line 52: Is there any limitations to using MPS? If application are hitting the GPU memory for instance?
Should we add a section on this?

sohamm17: Good point, @jjomier ! Added a cautionary note there about hitting GPU memory limits.

PR #125, opened 2023-11-02
Fix common external video source parameters

ibstewart:
The AJA and Deltacast operators share some common parameters but they were not using the same names and were not being forwarded to the operator constructor properly. This change ensures that both operators use parameters with the 'rdma' and 'enable_overlay' names, and then appends the common parameters to the argument list to construct the operator.


jjomier (approved):
LGTM

PR #126, opened 2023-11-02
Respect the `storage_type` argument of the provided GXF `create_out_message_with_tensor` utility

grlee77:
I noticed that the `create_out_message_with_tensor` utility function provided in the CV-CUDA C++ application was ignoring the provided `storage_type` argument and always allocating a device array. Device tensors are what is wanted for the CV-CUDA application, but the `gxf_utils.hpp` was intended to be general for GXF tensors and so should also support being able to allocate system memory or page-locked host memory as well. That update is made in this PR.

This PR does not modify the behavior of the existing application, just addresses this shortcoming in the provided utility function.

### DIFF
--- a/applications/cvcuda_basic/cpp/gxf_utils.cpp
+++ b/applications/cvcuda_basic/cpp/gxf_utils.cpp
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 #include <cuda_runtime.h>
+#include <fmt/format.h>
 
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


applications/cvcuda_basic/cpp/gxf_utils.cpp
```diff
@@ -28,16 +29,65 @@ std::pair<nvidia::gxf::Entity, std::shared_ptr<void*>> create_out_message_with_t
...
+      // Create a shared pointer for the CUDA memory with a custom deleter.
+      pointer = std::shared_ptr<void*>(new void*, [](void** pointer) {
+        if (pointer != nullptr) {
+          if (*pointer != nullptr) { cudaFree(*pointer); }
+          delete pointer;
+        }
+      });
+
+      // Allocate the CUDA memory (don't need to explicitly initialize)
+      const cudaError_t err = cudaMalloc(pointer.get(), nbytes);
```
grlee77, line 66: This `kDevice` case is the code path that was previously hard-coded

jjomier (approved):
LGTM

PR #134, opened 2023-11-13
Fix build qcap_source

DavidSu-Yuan:
Need some modifications to build and run the qcap_source operator.
1. Install depended library in docker for running QCAP SDK.
2. Map /opt/yuan/qcap/include and lib into docker for building qcap_source.
3. Add the define BUILD_WITH_QCAP_SDK to the file, qcap_source.cpp.
    The define BUILD_WITH_QCAP_SDK to control qcap_source has capture function or just a fake operator. 
    Original intent is avoid build breaking for the user who don't have QCAP SDK.

### DIFF
--- a/gxf_extensions/yuan_qcap/qcap_source.cpp
+++ b/gxf_extensions/yuan_qcap/qcap_source.cpp
@@ -16,11 +16,9 @@
  */
 #include "qcap_source.hpp"
 
-#ifdef BUILD_WITH_QCAP_SDK
 #include <qcap.common.h>
 #include <qcap.h>
 #include <qcap.linux.h>
-#endif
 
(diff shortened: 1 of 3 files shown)


gxf_extensions/yuan_qcap/qcap_source.cpp
```diff
@@ -16,6 +16,7 @@
  */
 #include "qcap_source.hpp"
 
+#define BUILD_WITH_QCAP_SDK
 #ifdef BUILD_WITH_QCAP_SDK
```
jjomier, line 20: We probably should remove this ifdef instead.

DavidSu-Yuan:
OK, let's remove the define.

PR #145, opened 2023-11-27
Added TURN server support to webRTC server example app

MaggieXJZhang:
This PR contains the minimum changes needed for the webRTC video server example application to support TURN server configurations, which enables Kubernetes deployment & Docker deployment of the application without host networking.

Here are the changes:
1. The `WebRTCServerApp`  is modified to support a flag `[--ice-server ICE_SERVER]` which can be specified multiple times if the user need to pass in more than one TURN/STUN server.

2. An `iceServer` REST endpoint is added to the `WebAppThread` so that javascript can obtain the list of specified ice servers before creating the RTCPeerConnection.
3. `client.js` has modifications to fetch the iceServers
4. The `negotiate()` function in `client.js` has gone through a slight re-ordering to align with webRTC and ICE protocol. (See details below) 
5. README has also been updated with TURN server instructions.

**Note on reordering of `negotiate()` function:**

Before this PR, the call order inside negotiate function is: createOffer -> setLocalDescription -> waitForICEGathering -> sendOffer -> receiveAnswer -> setRemoteDescription.

The way webRTC works is that the two peers must first exchange SDP via the signaling server before exchanging ICE candidates needed for data to flow through. In our case the two peers are the client.js application and webRTC Server operator, and the signalling server is the webAppServer which is able to pass on messages between the two peers.  (More on webRTC signalling can be seen https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Signaling_and_video_calling#signaling_transaction_flow)

Therefore, the proper call order following the webRTC protocol is: createOffer -> setLocalDescription ->  sendOffer -> receiveAnswer -> setRemoteDescription -> waitForICEGathering, which this PR hopes to address.

This change of ordering in call stack leads to a a few seconds of speed up for the latency from "clicking start" to seeing the frame. The speed up was most prominent in Firefox when accessed from remote machine. 

| Browser | Accessed via | Avg Latency Before Reorder (s) | Avg Latency Before Reorder (s) | Speed up (s)
| ------------- | ------------- | ------------- | ------------- |------------- |
| Firefox | localhost  | 6.97 | 6.0 | 0.97|
| Firefox | remote machine | 17.47  | 5.59 | 11.88|
| Chrome | localhost  | 6.76 |  5.62 | 1.14|
| Chrome | remote machine | 6.60 |  5.56 | 1.04|

Raw data, collected over 5 trials.

| Browser | Accessed via  | 5 Trials Before Reorder (s)  | 5 Trials After Reorder(s)  |
| ------------- | ------------- | ------------- | ------------- |
| Firefox | localhost  |  7.05, 7.30, 6.81, 6.78, 6.91 | 5.80, 7.41, 5.47, 5.68, 5.63 |
| Firefox | remote machine | 17.84, 16.99, 16.93, 17.77,  17.80 | 5.52, 5.54, 5.56,  5.64,  5.67|
| Chrome | localhost  | 6.80, 6.83, 6.68, 6.68,  6.80 | 5.65, 5.62,  5.65, 5.54,  5.62 |
| Chrome | remote machine | 5.78, 6.79, 6.73, 6.81, 6.88  |  5.52, 5.56, 5.63, 5.57, 5.54 |

### DIFF
--- a/applications/webrtc_video_server/README.md
+++ b/applications/webrtc_video_server/README.md
@@ -45,7 +45,7 @@ Press the `Start` button. Video frames are displayed. To stop, press the `Stop` 
 ### Command Line Arguments
 
 ```
-usage: webrtc_server.py [-h] [--cert-file CERT_FILE] [--key-file KEY_FILE] [--host HOST] [--port PORT] [--verbose VERBOSE]
+usage: webrtc_server.py [-h] [--cert-file CERT_FILE] [--key-file KEY_FILE] [--host HOST] [--port PORT] [--verbose] [--ice-server ICE_SERVER]
 
 optional arguments:
   -h, --help            show this help message and exit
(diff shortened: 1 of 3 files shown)


MaggieXJZhang:
I've only added docker instructions. Please let me know if you'd like to see K8S instructions (i.e. helm chart, deployment steps) in this PR as well or not.

AndreasHeumann (approved):
Looks good to me. Thanks for adding this!

slbryson:
Minor comment: In the first performance table, should the fourth column be "..After Reorder.."

PR #148, opened 2023-11-28
Updated webrtc_holoviz example with TURN server and small bug fix on webrtc_video example

MaggieXJZhang:
Continuation of https://github.com/nvidia-holoscan/holohub/pull/145

This MR adds the same logic of MR!145 to the webrtc_holoviz_server example application.

It also fixes a small bug in the previous MR where if the user clicks start and stop multiple times, duplicate TURN servers gets added.


AndreasHeumann (approved):
Looks good to me. Thanks!

PR #171, opened 2024-01-29
Endoscopy Benchmarking into CI/CD

sohamm17:
- Modified the benchmarking ctest
- Added the python package requirements for benchmarking into Dockerfile


applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
```diff
@@ -19,7 +19,6 @@ extensions:
   - libgxf_cuda.so
   - libgxf_multimedia.so
   - libgxf_serialization.so
-  - libgxf_stream_playback.so
```
jjomier, line 22: @sohamm17 we need to put this line back - this would be addressed in another MR.
Note that it should work for the 1.0 release

sohamm17: Added back.

PR #175, opened 2024-02-09
Fix h264 end dec apps

jjomier:
Updated H264 support for Holoscan 1.0


dev_container
```diff
@@ -475,6 +475,11 @@ launch() {
         conditional_opt+=" -v /opt/yuan/qcap/lib:/opt/yuan/qcap/lib"
     fi
 
+    # add tegra directory
+    if [ -d  /usr/lib/aarch64-linux-gnu/tegra ]; then
```
wendell-hom, line 479: Which libraries are we depending on from here, are they not mounted with `runtime=nvidida`?

applications/orsi/lib/CMakeLists.txt
```diff
@@ -19,21 +19,21 @@ project(orsi_app_lib CXX)
 find_package(holoscan 0.6 REQUIRED CONFIG
              PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")
 
-add_library(${PROJECT_NAME}
+add_library(orsi_app_lib EXCLUDE_FROM_ALL
```
wendell-hom, line 22: With these changes, do the orsi lib no longer get compiled if we are not building the orsi app?

dev_container
jjomier: These are the multimedia libraries

applications/orsi/lib/CMakeLists.txt
jjomier: This is correct. This was an issue that it was built by default so building the h264 will build the orsi lib.
Could be put in another MR but good to put here.

dev_container
agirault: Is this an issue on dGPU only? the nvidia runtime should handle things

jjomier: the nvidia runtime does not add this directory

applications/orsi/lib/CMakeLists.txt
wendell-hom: fine with having this here

wendell-hom (approved):

PR #179, opened 2024-02-15
CVCUDA-Holoscan Interoperability Operators

sohamm17:
- Adds two operators to convert a Holoscan tensor to CVCUDA tensor and vice-versa
- Refactors the cvcuda_basic example application to use those two operarors


dev_container
```diff
@@ -435,10 +435,11 @@ launch() {
     # Allow connecting from docker. This is not needed for WSL2 (`SI:localuser:wslg` is added by default)
     run_command xhost +local:docker
 
+    for video_dev in $(find /dev -regex '/dev/video[0-9]+'); do
```
jjomier, line 438: Can we do the same for AJA?

sohamm17: done.

grlee77 (approved):
Thanks, Soham!

PR #181, opened 2024-02-20
[CVCUDA/Holoscan Interop] Set pointer to null after freeing them

sohamm17:
As Stroustrup mentions, `delete`ing/`free`ing a pointer would not necessarily set it to null. It is good idea to set such  a pointer to null.


jjomier (approved):
LGTM

sohamm17:
Stroustrup's comment: https://www.stroustrup.com/bs_faq2.html#delete-zero

PR #196, opened 2024-02-27
Added argument to disable interactive mode

jjomier:


dev_container
```diff
@@ -400,6 +401,10 @@ launch() {
            use_tini=1
            shift
         fi
+        if [ "$arg" = "--noit" ]; then
```
tbirdso, line 404: Please update `launch_desc` as well

tbirdso (changes requested):
One request to update docs, otherwise LGTM

tbirdso (approved):
LGTM!

PR #206, opened 2024-03-04
Add pre-commit hook to summarize subprojects

tbirdso:
Changes:
- Adds `PROJECTS.md` containing a markdown table of existing HoloHub
  projects and high-level metadata details such as supported platform
  and tested Holoscan versions;
- Adds pre-commit hook to automatically run the HoloHub summary utility
  and commit any updates to `PROJECTS.md` with each commit.
- References `PROJECTS.md` and GitHub Pages in README

### DIFF
--- a/PROJECTS.md
+++ b/PROJECTS.md
@@ -1,0 +1,53 @@
+|    | name                                                         | project_type   | platforms          | language   |   ranking | holoscan_sdk.tested_versions   |
+|---:|:-------------------------------------------------------------|:---------------|:-------------------|:-----------|----------:|:-------------------------------|
+|  0 | Basic PDW Pipeline                                           | application    | ['amd64', 'arm64'] | C++        |         4 | ['0.5.0']                      |
+|  1 | Colonoscopy segmentation                                     | application    | ['amd64', 'arm64'] | Python     |         1 | ['0.5.0']                      |
+|  2 | Endoscopy Depth Estimation                                   | application    | ['amd64', 'arm64'] | Python     |         2 | ['0.6.0']                      |
+|  3 | Endoscopy Out of Body Detection Pipeline in C++              | application    | ['amd64', 'arm64'] | C++        |         1 | ['0.5.0']                      |
+|  4 | Endoscopy Tool Segmentation from MONAI Model Zoo Application | application    | ['amd64', 'arm64'] | Python     |         2 | ['0.6.0']                      |
+|  5 | FM-ASR                                                       | application    | ['amd64']          | Python     |         3 | ['0.4.1', '0.5.0']             |
+|  6 | HoloChat-local                                               | application    | ['amd64', 'arm64'] | Python     |         4 | ['0.6.0']                      |
+|  7 | Hyperspectral image segmentation                             | application    | ['amd64', 'arm64'] | Python     |         2 | ['0.6.0']                      |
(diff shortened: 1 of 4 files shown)


PROJECTS.md
```diff
@@ -0,0 +1,53 @@
+|    | name                                                         | project_type   | platforms          | language   |   ranking | holoscan_sdk.tested_versions   |
```
jjomier, line 1: Can we prettify this:
project_type -> project type
holoscan_sdk.tested_versions -> HSDK tested version

PROJECTS.md
```diff
@@ -0,0 +1,53 @@
...
+|---:|:-------------------------------------------------------------|:---------------|:-------------------|:-----------|----------:|:-------------------------------|
+|  0 | Basic PDW Pipeline                                           | application    | ['amd64', 'arm64'] | C++        |         4 | ['0.5.0']                      |
```
jjomier, line 3: Can we change ['amd64', 'arm64'] to amd64, arm64?

jjomier: Same for the other arrays

PROJECTS.md
```diff
@@ -0,0 +1,53 @@
...
+| 34 | emergent_source                                              | operator       | ['amd64', 'arm64'] | nan        |         1 | ['0.5.0']                      |
+| 35 | lstm_tensor_rt_inference                                     | operator       | ['amd64', 'arm64'] | nan        |         1 | ['0.5.0']                      |
+| 36 | npp_filter                                                   | operator       | ['amd64', 'arm64'] | C++        |         1 | ['0.6.0']                      |
+| 37 | openigtlink                                                  | operator       | ['amd64', 'arm64'] | C++        |         2 | ['1.0.3']                      |
+| 38 | prohawk_video_processing                                     | operator       | ['arm64']          | C++        |         4 | ['0.5.1', '0.6.0']             |
+| 39 | qt_video                                                     | operator       | ['amd64', 'arm64'] | C++        |         1 | ['0.6.0']                      |
+| 40 | tensor_to_video_buffer                                       | operator       | ['amd64', 'arm64'] | nan        |         1 | ['0.5.0']                      |
```
jjomier, line 43: what does nan means?

PROJECTS.md
jjomier, line 1: Can we order by alphabetical order?

PROJECTS.md
tbirdso: Sure, I'll look into cleaning up the table as requested here and in other comments below

PROJECTS.md
tbirdso: Currently the summary script orders first by project type (application, GXF extension, operator) and then alphabetically by project name. Personal preference is to keep project types together, since some applications, extensions, and operators share the same name and it's otherwise not immediately clear at a glance what the difference among those rows is.

Do you think the ordering would be clearer if I moved the project type to be the first column?

jjomier: Maybe we should separate in multiple tables, that would make things easier to visualize.

jjomier: I mean multiple tables in the same PROJECTS.md file

jjomier:
@tbirdso do you want to close this MR?

jjomier:
Ok to keep open, but maybe we should change the title of the MR of we plan follow up.

PR #210, opened 2024-03-05
Application graph with performance numbers along with live update

sohamm17:


jjomier:
@sohamm17 can you fix the linting issues?

PR #213, opened 2024-03-06
Use VideoMaster API Helper lib

Nathoufresh:


gxf_extensions/deltacast_videomaster/FindVideoMaster.cmake
laurent-radoux, line 1: Let's prepare for new upcoming VideoMaster release with native find_package support.

gxf_extensions/deltacast_videomaster/videomaster_base.hpp
```diff
@@ -62,13 +65,13 @@ class VideoMasterBase : public gxf::Codelet {
 
   bool signal_present();
   bool set_loopback_state(bool state);
-  std::unordered_map<ULONG, ULONG> get_detected_input_information(uint32_t channel_index);
-  std::unordered_map<ULONG, ULONG> get_input_information();
 
-  bool api_call_success(ULONG api_error_code, std::string error_message);
```
laurent-radoux, line 68: Wouldn't it still be useful?
The code has become a tad bit more verbose with all the ifs after each api call.

laurent-radoux (changes requested):

PR #214, opened 2024-03-07
Infra: Add `--as_root` launch option to `dev_container` script

tbirdso:
Adds an `--as_root` launch option to allow running the HoloHub container as the root user.

Running as root can often be helpful for development and debugging. Some apps in the future may also require running as root, such as for SSO login.


tbirdso:
Approved internally

PR #216, opened 2024-03-08
Launch: Append to PYTHONPATH instead of replacing

tbirdso:
Appends launch PYTHONPATH to the environment value instead of replacing the value of PYTHONPATH.

This intends to address the scenario where a custom Dockerfile provides a PYTHONPATH to reference Python libraries in the container, but then `./run launch` cannot be used because it overwrites PYTHONPATH.


tbirdso:
Approved internally

PR #230, opened 2024-03-15
Fix bug in Holoscan 2.0 code path of TensorToVideoBufferOp

grlee77:
The primary purpose of this MR is to fix a broken code path for Holoscan 2.0 within TensorToVideoBufferOp

Specifically, there was a location where a `Tensor` was created instead of the expected `shared_ptr<Tensor>`.

This MR also fixes some mistakes in the h264 application README files.

I confirmed that compilation of `h264_endoscopy_tool_tracking ` is fixed by this change.


jjomier:
Thanks for fixing this Greg!

PR #233, opened 2024-03-20
Doc: Add GitHub badges with project statistics and workflow automation

tbirdso:
Changes:

- Checks in and updates the `gather_metadata.py` utility as a standard source of "truth" to collect metadata from HoloHub subprojects. Updates the `summarize_metadata.py` utility to use those results, fixing an issue where that utility previously failed to reflect some metadata from nested folders.
- Adds badges to the HoloHub README to reflect the number of applications and operators available in HoloHub. We can add more badges later to reflect other statistics, such as tutorial counts or latest HSDK supported versions.
- Adds workflow to automatically update project statistics badges in the HoloHub README. If a branch adds a new project, the README will be updated in a new commit appended to the branch co-authored by a GitHub bot.

README after this change:

image

### DIFF
--- a/.github/workflows/update_readme.yml
+++ b/.github/workflows/update_readme.yml
@@ -1,0 +1,34 @@
+name: Update HoloHub README Statistics
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  update-readme:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - name: Update README
+        run: |
+          python ./utilities/gather_metadata.py --output aggregate_metadata.json
+          app_count=$(cat aggregate_metadata.json | grep "\"source_folder\": \"applications\"" | wc -l)
+          sed -i -E "s/Applications-([0-9]+)/Applications-$app_count/" README.md
+          ops_count=$(cat aggregate_metadata.json | grep "\"source_folder\": \"operators\"" | wc -l)
+          sed -i -E "s/Operators-([0-9]+)/Operators-$ops_count/" README.md
+      
+      - name: Auto-Commit README Changes
+        uses: EndBug/add-and-commit@v9
+        with:
+          add: 'README.md'
+          commit: '--signoff'
+          message: Update README Project Statistics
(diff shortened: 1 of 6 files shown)


tbirdso:
Note that we have three options available for how to commit README changes:
1. (favored) In the contributor's PR branch: Best visibility to contributors and maintainers. Could cause confusion where the upstream branch has one more commit than local, requiring either a `pull` prior to more development or force-push to override bot commit.
2. Run workflow in `main` branch after PR merge and add commit directly to `main`: May be less confusing for contributors, but more difficult to debug. Risks automatically updating documentation with improper statistics without a human in the loop.
3. Run workflow in `main` and create a new PR: May be more confusing for maintainers, requires extra action to keep documentation up to date

@jjomier Please let me know if you have a preference to pursue a different path among these three.

tbirdso:
It looks like the auto-commit action is not currently permitted to run in `nvidia-holoscan`. @jjomier are we able to add an exception?

jjomier (approved):
Looks great!
Let me look at the permission for the action

tbirdso:
Per request I've updated the proposed README action to use `EndBug/add-and-commit`. As part of that investigation I've learned more about the limitations of automatic commits with PRs from forks, as discussed in [`add-and-commit` docs](https://github.com/EndBug/add-and-commit?tab=readme-ov-file#working-with-prs):
- Running `add-and-commit` on `pull_request` can result in errors if `nvidia-holoscan` does not have write permissions to update the branch in the user's fork (default), as seen in https://github.com/nvidia-holoscan/holohub/actions/runs/8367432453
- It's not clear that running `add-and-commit` on `pull` only would allow us to fully police the action to ensure that the README has been updated, as that would depend entirely on the user repo and README workflow results may not show up in `nvidia-holoscan` PRs. Also, could be confusing for the user if the workflow only runs on their fork.

Based on this, I recommend that we pursue the second option in my [comment](https://github.com/nvidia-holoscan/holohub/pull/233#issuecomment-2009686067) above, where a new commit may be pushed to `main` with README updates immediately after a MR is merged. That will keep edits completely within the scope and control of `nvidia-holoscan/holohub` and out of the user forks, so we never cause unexpected behavior or deal with permissions in the user fork. One caveat of that approach is that we may need additional effort to run CI workflows on the automated commit.

I've tested that in the `main` branch of my user fork as seen here: https://github.com/tbirdso/holohub/commits/main/

TL;DR: Managing automatic commits and PRs from user forks is tricky. Running directly on `main` seems to be the best path forward to minimize both risks and complexity.

jjomier:
I like the idea of the workflow that update main. Seems cleaner and we keep control (make sure also that the statistics are coherent).

PR #235, opened 2024-03-21
Utility: Consolidate and Update Metadata Schema

tbirdso:
## Motivation

HoloHub requires that projects provide a `metadata.json` file adhering to certain structures. This allows us to implement standard processes that collect project information and improve project visibility within HoloHub.

This PR intends to address pain points in the HoloHub metadata schema strategy:
1. Reduce schema maintenance requirements by reducing duplicate definitions;
2. Implement requested fields and schema for richer project descriptions

## Metaschema Updates
- Adds `project.schema.json` metaschema to supply schema elements for
      reuse across HoloHub project types. This reduces the maintenance
      burden for any changes in metadata conventions that should apply
      across HoloHub projects. Existing `metadata.schema.json` files are
      updated to reference metaschema elements rather than duplicating
      definitions.
- Adds ID field to each project schema with versioning information,
      starting at v1. This will allow us to update schema in the future
      without requiring that we immediately enforce the updated schema on
      every HoloHub project.

## Schema enforcement updates
 - Adds regular expression in metaschema to validate and enforce
      version semantics. Regex string is based on semver but relaxed to
      accept existing HoloHub conventions (major.minor or
      major.minor.revision).
 - Adds optional "description" field
 - Adds optional application "run command" field
 - Adds optional tutorial schema

## Additional Notes
The only immediate changes impacting existing HoloHub projects are as follows:
- Version fields must now match `<major>.<minor>` or `<major>.<minor>.<revision>` regex (no violations observed);
- If a tutorial supplies a `metadata.json` it must match the provided schema

Following alignment on this change, we will discuss strategies and milestones for if/when any future field requirements can be introduced and enforced.

### DIFF
--- a/.github/workflows/check_metadata.yml
+++ b/.github/workflows/check_metadata.yml
@@ -10,5 +10,14 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v4
 
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
+      - name: Install Utility Dependencies
+        run: |
+          python -m pip install -r ./utilities/requirements.txt
+
       - name: Run metadata_validator.py
         run: python utilities/metadata_validator.py
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,7 +86,7 @@ We do not require that community members conduct formal Software Quality Assuran
 Every application and operator should have an associated *metadata.json* file which describes the features
 and dependencies.
 
-`metadata.json` schemas differ slightly for [applications](./applications/metadata.schema.json), [GXF extensions](./gxf_extensions/metadata.schema.json), and [operators](./operators/metadata.schema.json), but generally follow the convention below:
+`metadata.json` schemas differ slightly for [applications](./applications/metadata.schema.json), [GXF extensions](./gxf_extensions/metadata.schema.json), [operators](./operators/metadata.schema.json), and [tutorials](./tutorials/metadata.schema.json), but generally follow the convention below:
 
 ```json
 // Main json definition for application or operator
@@ -228,9 +228,10 @@ function. This keyword should only be used for sample applications that are main
 
 Add each tutorial in its own subdirectory under the [```tutorials```](./tutorials) directory. The subdirectory should contain:
 - A README file summarizing the application's purpose and architecture;
+- A *metadata.json* file which describes its specifications and requirements in accordance with the [tutorial metadata.json schema](./tutorials/metadata.schema.json) (optional);
 - A LICENSE file governing use (optional).
 
-There are no project-wide metadata or build requirements for tutorials.
+There are no project-wide build requirements for tutorials.
 
 ### License Guidelines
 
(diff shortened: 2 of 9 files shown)


jjomier:
We should force the tutorials to have a metadata information and add the files - thoughts?

tbirdso:
> We should force the tutorials to have a metadata information and add the files - thoughts?

@jjomier I agree that we should move in that direction, but as a general strategy I'd like for us to move away from having to immediately reflect any schema updates across all HoloHub projects.

If we require that all projects match the latest metadata schema, it quickly becomes painful to add any new required fields to any schema. For instance, if we have 100 applications and want to introduce a new field requirement, we would either need to update all 100 application metadata files ourselves, wait for all 100 original contributors to update, or abandon the field.

One of the two major goals of this PR is to begin to introduce versioning infrastructure that will allow us to incrementally introduce schema updates without immediately requiring that all applications update. In this case, we should begin the practice of incremental updates by currently making schema changes optional, and only then starting to enforce them as required sometime in the (near) future.

In future cases, we would ideally have a "v1" schema that is currently enforced, a "v2" schema in the proposal phase, a date/milestone for when the "v2" schema will no longer accept new proposal updates, and a following date/milestone for when we will begin to enforce the "v2" schema.

That schedule is beyond the scope of this PR, but I suggest that we start to think along those lines by making proposed metadata changes optional for the moment and setting a rough date (say April 1) when we will start to enforce the tutorial schema. Does that make sense to you?

jjomier:
That makes sense. However, the github page is using this metadata information to generate the pages so if we do not have a consistent schema this is not going to be maintainable. I agree that this could be done in different PR, but ideally we should modify the existing metadata to reflect the new schema.

tbirdso:
> if we do not have a consistent schema this is not going to be maintainable

In a sense I would argue the opposite -- requiring all metadata immediately be updated to match an evolving schema is not going to be maintainable.

More to your point though, in my view when we consider the needs of developers, maintainers, and automation in the HoloHub monorepo, the most maintainable approach would be to have backwards compatibility in infrastructure for exactly two schema versions, those being 1) the schema we are moving away from and 2) the schema we are moving towards. We shouldn't need to immediately update 1) every project's `metadata.json` and 2) our GitHub Pages infrastructure each time we propose a small schema change.

> I agree that this could be done in different PR, but ideally we should modify the existing metadata to reflect the new schema.

Yes, I agree that we should set this as a goal to work towards over a fixed timeline. Let's follow up with tutorial metadata in one or several subsequent PRs. I can take on that task.

jjomier:
I think we should almost never modify the schema - unless to add new fields potentially (which would make the other format backward compatible). Have a schema that is too complex will be difficult for authors and it should be kept to a minimum IMHO.

tbirdso:
> I think we should almost never modify the schema - unless to add new fields potentially (which would make the other format backward compatible).

I agree in concept, but in practice I'd like to have a strategy in place that is able to handle breakage without requiring a simultaneous update across every HoloHub project. I think it's reasonable to continue to augment the "v1" version with new, optional fields for now with the option to eventually pursue a "v2" schema with breaking changes that will require updates across the board over some clearly communicated timeline.

> Have a schema that is too complex will be difficult for authors and it should be kept to a minimum IMHO.

Completely fair. Let's aim to add a metadata template alongside README templates in the near future to minimize complexity for most users.

Moving to merge following offline approval from @jjomier.

PR #236, opened 2024-03-21
Add --persistent option to dev_container

jjomier:
This new -`-persistent` option disables `docker --rm` command and keep the container data intact after launch.


tbirdso (approved):
That was fast! LGTM

PR #237, opened 2024-03-21
Fix volume_rendering application's instructions

jjomier:
Change the instructions to use the run script instead of individual command.


jjomier:
Yes tested baremetal and it works as expected.

tbirdso (approved):

PR #238, opened 2024-03-21
Adding keywords definitions for different locations

jjomier:


applications/colonoscopy_segmentation/README.md
```diff
@@ -19,6 +19,13 @@ If you want to manually convert the video data, please refer to the instructions
 To run this application, you'll need to configure your PYTHONPATH environment variable to locate the
 necessary python libraries based on your Holoscan SDK installation type.
 
+The following terms are used to define specific locations:
```
tbirdso, line 22: How would you feel about adding a reference back to definitions in `CONTRIBUTING.md` rather than copy/paste in app READMEs?

This is OK to merge as is but I worry about maintainability as more apps are added or as definitions go out of spec.

jjomier: Great point - let me do that. I'll add a section in the CONTRIBUTING.md.
Is `glossary` a good term for that or you have a better suggestion?

tbirdso: "Glossary" sounds good to me 👍

PR #241, opened 2024-03-21
bump CV-CUDA to v0.6.0-beta and simplify the cvcuda_basic Dockerfile

grlee77:
This MR bumps the CV-CUDA version used by the `cvcuda_basic` app to v0.6.0-beta and simplifies the Dockerfile in a couple of ways
- a CUDA version patch is no longer needed (v0.6.0 can be built with CUDA>=11.4)
- `apt install git-lfs` prior to checkout from GitHub so manually downloading stub files is not needed

The generated `.deb` filenames change. I install all of them in the Dockerfile with the exception of `cvcuda-tests` as that one adds an additional dependency on `python3-pytest`.

There is no change to the application itself or its behavior

### DIFF
--- a/applications/cvcuda_basic/Dockerfile
+++ b/applications/cvcuda_basic/Dockerfile
@@ -29,38 +29,25 @@ ARG GPU_TYPE
 FROM ${BASE_IMAGE} as cvcuda-downloader
 
 ARG GCC_VERSION=11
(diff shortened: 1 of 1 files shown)


jjomier:
@grlee77 LGTM. Is this backward compatible?

grlee77:
Yes, it should be backwards compatible.

PR #245, opened 2024-03-25
Infra: Default to first available application language at launch

tbirdso:
Changes:
- Adds `run launch --verbose` option to aid in development
- If a language is not specified for the `./run launch` command,
      use the output of `./run list` to select the first available
      application language implementation for the requested app name
- Fix issue where positional language argument was used in place of
    updated local variable
- Fixes `list_desc` function name so that `./run list --help` prints as
      expected

Test with `multiai_ultrasound`:
```sh
$ ./dev_container build && ./dev_container launch && ./run build multiai_ultrasound
>>>$ ./dev_container launch multiai_ultrasound
Default language for multiai_ultrasound selected: cpp
...
```

Or:
```sh
$ ./dev_container build_and_run multiai_ultrasound
```

Apps that do not provide or require a language are unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,17 +146,22 @@ To list all available applications you can run the following command:
   ./run list
 ```
 
-Then you can run the application using
+Then you can run the application using the command:
 ```bash
-  ./run launch <application> <language>
+  ./run launch <application>
 ```
 
-For example, to run the tool tracking endoscopy application in C++
+For example, to run the tool tracking endoscopy application:
 ```bash
-  ./run launch endoscopy_tool_tracking cpp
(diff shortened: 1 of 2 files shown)


jjomier (approved):
Looks great - thanks for adding this!

Should we document the behavior in the README? So that users are not surprised if a language is not provided that the first one will be used?

tbirdso:
> Looks great - thanks for adding this!
> 
> Should we document the behavior in the README? So that users are not surprised if a language is not provided that the first one will be used?

Yes, good idea. Let me update that before I merge.

PR #249, opened 2024-03-25
Infra: Do not show empty `./run launch` default language

tbirdso:
Addresses issue where an application that does not support a language specifier, such as colonoscopy_segmentation, would print out an empty language string from its language detection attempt. That printout could be confusing for users.

For example, under previous behavior the following line was printed to the console:
"Default language for colonoscopy segmentation selected: "

Under updated behavior the line is not printed.

Applications that do accept a language specifier are not impacted by this change and will continue to print the default selection.


jjomier (approved):
LGTM

PR #250, opened 2024-03-26
Utility: Add tutorial template and update validator tool

tbirdso:
Changes:
- Adds `template/` stubs to `tutorial` folder for new projects to use and updates metadata validator script accordingly
- Fixes issue in metadata validator script where a JSON syntax error in a project or schema would error out without reporting the file origin of the error


jjomier (approved):
Looks great.

PR #259, opened 2024-03-27
Remove references to add_operator in tutorial creating-multi-node-applications

jin-nvidia:


tutorials/creating-multi-node-applications/README.md
```diff
@@ -164,7 +163,7 @@ class MultiAIDetectionSegmentation(Application):
 ```
 
 ### Changes in scenario 1 - Adding `add_operator` in fragments
-When composing a non-distributed application, in the `compose()` method we need to create the operators and `add_flow` between pairs of operators. When composing fragments, in the `compose()` method there is an extra step, after creating the operators, we will need to do `add_operator` for each operator, and then `add_flow`.
+When composing a non-distributed application, in the `compose()` method we need to create the operators and `add_flow` between pairs of operators. By default, `add_flow` already adds each operator to the app graph. When composing fragments, in the `compose()` method if there are operators not added to the application graph by `add_flow`, we will need to do `add_operator` to add each operator to the app graph.
```
agirault, line 166:
```suggestion
When composing a non-distributed application, operators are created in the `compose()` method, then added to the app graph one of two ways:
1. For applications with a single operator (rare), `add_operator()` should be called.
1. for applications with multiple operators, using `add_flow()` will take care of adding each operator to the app graph on top of connecting them.

This applies to distributed applications as well: when composing multiple fragments, each of them are responsible for adding all their operators to the app graph in their `compose()` method. Calling `add_flow()` in the `compose()` method of the distributed application (connecting fragments) does not add the inner operators to the graph. This is often relevant when breaking down a single fragment application in a multi fragments application for distributed use cases, as some fragments might end up owning a single operator, and the removal of `add_flow()` in that fragment should come with the addition of `add_operator()` instead.
```

agirault (approved):
LGTM!

jin-nvidia:
Thanks! @agirault could you please help merge the PR? (I don't have write access.)

PR #261, opened 2024-03-31
Add vtk_renderer operator

vicentebolea:
Hello HoloHub team, this PR proposes adding a VTK based renderer operator as a proof of concept that aims to show how VTK can be used as an alternative as holoviz.

VTK can be a useful addition to holohub stack since VTK is a industry leading visualization toolkit. Showing how to use VTK with holoscan sdk in holohub can possibly help holoscan sdk users and developer to expand their capabilities when it comes to data visualization. Some of relevant features that might benefit the user are: applying filters, add further annotations, export data to other formats, leverage computation in HPC grade vtk-m filter implementations, and many other features.

On the drawback side, this renderer operator needs to copy the input from device memory to host due to limitations of VTK. I might be wrong but it appears that `holoviz` does not need to copy its input to host memory. While this is a strong limitation for VTK we believe that VTK can still be a good addition and VTK is an evolving project, perhaps in the future we could overcome this limitation.

While VTK will remain as an optional operator in `Holohub` a dockerfile to build VTK based on the developer docker image is also provided in this PR.

I have also modified the `endoscopy_tool_tracking` cpp version to optionally use `vtk_renderer` if the user specifies so in its yaml conf file.

Here are the instructions to build the `vtk_renderer` operator and enable it in the `endoscopy_tool_tracking` application.

### Screencast of the application in action

<details>

<summary>Attention, the video includes footage of a surgery</summary>

[Peek 2024-03-30 20-37.webm](https://github.com/nvidia-holoscan/holohub/assets/939798/1aa44f7e-5335-43e2-a634-55aa5d5ac4e9)

</details>

### How to build the tool tracking application with VTK

Build the HoloHub container as described at the root [README.md](../../README.md)

You need to create a docker image which includes VTK with the provided
`vtk.Dockerfile`:

```bash
docker build -t vtk:latest -f vtk.Dockerfile .
```
Set vtk as the visualizer by:

* Using a vtk_renderer instead of holoviz
    ```bash
    sed -i -e 's#^visualizer:.*#visualizer: "vtk"#' applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
    applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking --data <data_dir>/endoscopy
    ```
Then, you can build the tool tracking application with the provided
`Dockerfile`:

```bash
./dev_container launch --img vtk:latest
```

Inside the container you can build the holohub suite with:

```bash
./run build --with vtk_renderer
./run build
```

Now you can run the tool tracking application with:

```bash
./run launch endoscopy_tool_tracking cpp
```

Please let us know if you need us to make any change.

@agirault 
@tbirdso


vtk.Dockerfile
```diff
@@ -0,0 +1,43 @@
+# syntax=docker/dockerfile:1
```
tbirdso, line 1: Would prefer to move this file from the top level directory to the `operators/vtk_renderer` folder to which it relates and update reference docs accordingly.

operators/vtk_renderer/vtk_renderer.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+  spec.param(this->in_annotations,
+             "in_annotations",
+             "Input Annotations",
+             "Input Annotations",
+             &in_annotations);
+  spec.param(this->in_videostream,
+             "in_videostream",
+             "Input Videostream",
+             "Input Annotations",
```
tbirdso, line 103:
```suggestion
             "Input video stream",
```

operators/vtk_renderer/vtk_renderer.cpp
```diff
@@ -0,0 +1,188 @@
...
+             &in_videostream);
+
+  spec.param(width, "width", "Window width", "Window width in px", 640u);
+  spec.param(height, "height", "Window height", "Window height in px", 480u);
+  spec.param(labels, "labels", "tools labels", "tools labels");
+}
+
+void VtkRendererOp::start() {
+  this->rendererWindow->SetWindowName("VTK (Kitware)");
```
tbirdso, line 112: Suggest parameterizing the window name so that other users can update for their application needs.

operators/vtk_renderer/metadata.json
```diff
@@ -0,0 +1,31 @@
...
+		],
+		"version": "1.0",
+		"changelog": {
+			"1.0": "Initial Release"
+		},
+		"holoscan_sdk": {
+			"minimum_required_version": "0.5.0",
+			"tested_versions": [
+				"0.5.0"
+			]
```
tbirdso, line 18: It seems like the tested Holoscan SDK version is 1.0.3 based on the Dockerfile accompanying this PR. Can you please confirm?

operators/vtk_renderer/README.md
```diff
@@ -0,0 +1,18 @@
+### vtk_renderer operator
+
+The `vtk_renderer` extension takes the output the source video player and the
```
tbirdso, line 3:
```suggestion
The `vtk_renderer` extension takes the output of the source video player and the
```

operators/vtk_renderer/README.md
```diff
@@ -0,0 +1,18 @@
...
+output of the `tool_tracking_postprocessor` operator and renders the video
+stream with an overlay annotation of the label of the tool being used.
+
```
tbirdso, line 6: Please briefly note performance characteristics of the operator, including how often it performs host/GPU copies and any relevant benchmarking information. You can copy/paste what you've written in the PR description.

tbirdso (commented):
Hi @vicentebolea , thank you for contributing to HoloHub!

We've requested reviews from a couple of key Holoscan members. In the meantime I've added some comments below to begin discussion relating to the proposed VTK operator. Thanks!

vicentebolea:
@tbirdso many thanks for the promptly review, find the comment addressed in the last push. Let me know if you need further changes

vtk.Dockerfile
tbirdso: Looks like the file was copied rather than moved, this file can be deleted

tbirdso:
cc @okutter @AndreasHeumann for review

vtk.Dockerfile
vicentebolea: my bad, resolved in the last push

operators/vtk_renderer/README.md
```diff
@@ -0,0 +1,50 @@
...
+- **`videostream`**: Input channel for the videostream, type `gxf::Tensor`
+  - type: `gxf::Handle<gxf::Receiver>`
+- **`annotations`**: Input channel for the annotations, type `gxf::Tensor`
+  - type: `gxf::Handle<gxf::Receiver>`
+- **`width`**: width of the renderer window.
+  - type: `int`
+- **`height`**: height of the renderer window.
+  - type: `int`
+- **`labels`**: labels to be displayed on the rendered image.
+  - type: `int`
```
AndreasHeumann, line 50:
```suggestion
  - type: `std::vector<std::string>>`
```

operators/vtk_renderer/README.md
```diff
@@ -0,0 +1,50 @@
...
+```bash
+./run build --with vtk_renderer
+./run build
+```
+
+##### Parameters
+
+- **`videostream`**: Input channel for the videostream, type `gxf::Tensor`
+  - type: `gxf::Handle<gxf::Receiver>`
+- **`annotations`**: Input channel for the annotations, type `gxf::Tensor`
```
AndreasHeumann, line 43: Please add the `window_name` parameter as well.

operators/vtk_renderer/vtk_renderer.cpp
```diff
@@ -0,0 +1,189 @@
...
+ * limitations under the License.
+ */
+
+#include "vtk_renderer.hpp"
+
+#include "holoscan/core/execution_context.hpp"
+#include "holoscan/core/gxf/entity.hpp"
+#include "holoscan/core/io_context.hpp"
+#include "holoscan/core/operator_spec.hpp"
+#include "holoscan/core/resources/gxf/cuda_stream_pool.hpp"
```
AndreasHeumann, line 24: Is CudaStreamPool used? I could not find it.

operators/vtk_renderer/vtk_renderer.cpp
```diff
@@ -0,0 +1,189 @@
...
+  spec.param(this->in_annotations,
+             "in_annotations",
+             "Input Annotations",
+             "Input Annotations",
+             &in_annotations);
+  spec.param(this->in_videostream,
+             "in_videostream",
+             "Input Videostream",
+             "Input Videostream",
+             &in_videostream);
```
AndreasHeumann, line 103: The spec.param() here is not needed. I assume this is from the HolovizOp which has the `receivers` input which can be connected to multiple outputs. Since you split the inputs to two you don't need this.
```suggestion
  spec.input<gxf::Entity>("annotations");
  spec.input<gxf::Entity>("videostream");
```

operators/vtk_renderer/vtk_renderer.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // Render the videostream on the background layer.
+  auto videostream = op_input.receive<gxf::Entity>("videostream").value();
+  const auto videostream_tensor = videostream.get<Tensor>("");
+  if (videostream_tensor) {
+    holoscan::gxf::GXFTensor in_tensor_gxf{videostream_tensor->dl_ctx()};
+    auto& shape = in_tensor_gxf.shape();
+    const int x = shape.dimension(0);
+    const int y = shape.dimension(1);
+    const int z = shape.dimension(2);
+
```
AndreasHeumann, line 165: It might be nice to have some checks here that the image is really the expected format (uchar, component count, etc.) to avoid crashes when users try to input images with different formats.

operators/vtk_renderer/vtk_renderer.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+  VtkRendererOp() = default;
+
+  void setup(OperatorSpec& spec) override;
+  void start() override;
+  void compute(InputContext& op_input, OutputContext&, ExecutionContext& context) override;
+
+ private:
+  Parameter<holoscan::IOSpec*> in_annotations;
+  Parameter<holoscan::IOSpec*> in_videostream;
```
AndreasHeumann, line 48: Can be removed.

operators/vtk_renderer/vtk_renderer.hpp
```diff
@@ -0,0 +1,63 @@
...
+ * limitations under the License.
+ */
+
+#ifndef VTK_OPERATOR__HPP
+#define VTK_OPERATOR__HPP
+
+#include <string>
+#include <vector>
+
+#include <vtkImageActor.h>
```
AndreasHeumann, line 24: It might be nice to hide these dependency by e.g. using PImpl (https://en.cppreference.com/w/cpp/language/pimpl). Wit the includes in the header for each app using this operator Vtk spills into the app code. Also compile time increases.
But this is just a suggestion.

AndreasHeumann (commented):
Thanks for the contribution. Looks pretty good, it would be great to have some performance comparison.
I added some comments, but nothing major.

operators/vtk_renderer/vtk_renderer.hpp
agirault: @vicentebolea on top of slower compilation, the general concern with exposing these interfaces is that end users need them as well. For most projects packages, one can choose for dev vs non-dev packages, so that dof interest. For VTK though, I'm not sure there is a distribution mechanism without the development files?

agirault:
@vicentebolea thanks for the great contribution!

That does not need to block the MR, but as @AndreasHeumann said, having some benchmarkings compared to holoviz would be beneficial. VTK brings a huge convenience given the breadth of capabilities and the existing user base. At the same time, it would be great to also ensure we're setting it up for optimal performance as that is Holoscan's mission. Could a VTK pipeline like what is shown here be configured so that the buffers never leave the GPU? If I'm not mistaken, incoming tensors will currently be copied to the CPU before going back to the GPU for rendering with OpenGL (please correct if I'm wrong, apologies if I missed it). In some of the early versions of the SDK, we had an opengl operator implemented by @okutter (since then deprecated in favor of Vulkan/Holoviz) which made use of [CUDA-OpenGL interop](https://docs.nvidia.com/cuda/cuda-runtime-api/index.html). That could be a good place to start to improve performance if needed.

operators/vtk_renderer/vtk_renderer.cpp
vicentebolea: Not used, I removed it

operators/vtk_renderer/vtk_renderer.hpp
vicentebolea: Please find the VTK state of the vtkRendererOp moved to a nested opaque class (a pimpl idiom). Indeed it is unnecessary and inconvenient for our headers to depend in VTK. Thanks for point this out.

vicentebolea:
@agirault thanks for the insightful review.

> Could a VTK pipeline like what is shown here be configured so that the buffers never leave the GPU? If I'm not mistaken, incoming tensors will currently be copied to the CPU before going back to the GPU for rendering with OpenGL 

Yes, in order to pass the data to VTK we need to first copy to host(cpu). AFAIK, there is no easy way of rendering directly from CUDA to the GPU framebuffer with VTK. Perhaps something can be done in the next iterations of this task and with changes in upstream VTK. Perhaps the best course of action is to have this first version of the vtk_renderer operator merged and later explore (if possible) how to further optimize this, this is to make VTK both: Accept GPU memory, Render to the GPU directly.

tbirdso:
@vicentebolea Thank you for addressing review comments! It would be beneficial to make performance benchmarks for the VTK operator available in the near future, but does not need to block this initial work. I have entered https://github.com/nvidia-holoscan/holohub/issues/276 to track, we would appreciate your help in following up there.

This initial implementation looks good and benefits Holoscan users with VTK interoperability, we can move forward with the merge.

tbirdso (approved):

PR #262, opened 2024-04-02
update ToolTrackingPostprocessorOp for compatibility with distributed apps

grlee77:
## Description
This MR makes some updates to the existing `ToolTrackingPostprocessorOp`. The behavior of the operator is the same as before **except** that the coordinate and mask tensors are now emitted on separate ports. This is done to enable compatibility of this operator with connection between fragments in a distributed application. For distributed apps, it is not allowed to transmit a mixture of host and device tensors on the same port.

This is not backwards compatible, but the changes needed on the application side are small (just changing one `add_flow` call as done here for Holohub's apps). There likely may not be any apps outside of Holohub using this operator as it is pretty specific to the tool tracking model used in the endoscopy examples.

## List of changes
* split the former `out` port into separate `out_coords` and `out_mask` ports (to separate CPU and GPU tensors for distributed app compatibility)
* document the input and output ports in the documentation strings
* document parameters in the C++ doxygen string
* clang-format was applied to modified C++ files (most changes in the H264 app are due to that)

## Tests done
I ran both the C++ and Python versions of the endoscopy tool tracking application and they continue to work as before. 

There is not currently a distributed version of this app on Holohub to verify that this fixes use in distributed applications. However, it is known that the UCX API used by `UcxTransmitter` cannot mix host and device tensors. Given that, it is necessary to use separate ports for host vs. device tensors as done here (in Holoscan SDK, each port has its own, independent transmitter).

### DIFF
--- a/applications/h264/h264_endoscopy_tool_tracking/main.cpp
+++ b/applications/h264/h264_endoscopy_tool_tracking/main.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -31,9 +31,7 @@
 
 class App : public holoscan::Application {
  public:
-  void set_datapath(const std::string& path) {
-     datapath = path;
-  }
+  void set_datapath(const std::string& path) { datapath = path; }
 
   void compose() override {
     using namespace holoscan;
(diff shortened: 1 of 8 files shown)


grlee77:
@jjomier, the changes in this MR are ones Victor and I needed to get a 3-fragment distributed version of the endoscopy tool tracking app working.

The changes here are also needed to align with proposed documentation updated to Holoscan SDK for creation of Python bindings of existing C++ operators (`ToolTrackingPostprocessorOp` is used as a concrete example in the proposed docs)

For testing, I verified locally that the existing `endoscopy_tool_tracking` C++ and Python applications continue to work back to Holoscan SDK v0.6 with the changes proposed here.

applications/h264/h264_endoscopy_tool_tracking/main.cpp
```diff
@@ -43,18 +41,17 @@ class App : public holoscan::Application {
     int64_t source_block_size = width * height * 3 * 4;
     int64_t source_num_blocks = 2;
 
-    auto bitstream_reader =
```
grlee77, line 46: I applied clang-format to the modified files which accounts for the style-related changes 

(similarly `black` was applied to Python scripts)

gigony (approved):
Thanks @grlee77 for updating the operator to make it available for the distributed app!

Looks good to me!

tbirdso (commented):
Hi @grlee77 , I see one more instance of `ToolTrackingPostprocessorOp` in a tutorial. Can we update that tutorial for `out_coords`, `out_mask` in this PR? It looks like you caught the other instances throughout HoloHub, thanks.

https://github.com/nvidia-holoscan/holohub/blob/main/tutorials/creating-multi-node-applications/scenario2/endoscopy_distributed_app.py#L187

Code looks good, no issues with moving forward after that fix. Thanks

grlee77:
> Hi @grlee77 , I see one more instance of ToolTrackingPostprocessorOp in a tutorial. Can we update that tutorial for out_coords, out_mask in this PR?

Great catch, Tom! I just pushed a new commit updating it.

tbirdso (approved):
Looks good to me, thanks!

mocsharp (approved):

PR #264, opened 2024-04-02
Tutorial: Add flow benchmarking support for Python applications

tbirdso:
Adds support for benchmarking Python applications in HoloHub as a parameter to the Holoscan Flow Benchmarking tutorial `benchmark.py` script.

Under previous behavior the benchmarking script primarily supported only HoloHub C++ applications. The recommended path to benchmark Python applications was to pass the full command string to "--run-command". This approach disallowed use of the "run" script that provides infrastructure for environment setup and working directory management. As a result, in practice "--run-command" alone required manual setup of the environment and working directory, i.e. several extra steps on the user's part.

This update allows users to launch a Python application with the "run" script to minimize additional user effort. The "--language" argument is forwarded to the "run" script which executes as otherwise recommended in HoloHub instructions. The "--run-command" parameter is left intact.

### DIFF
--- a/tutorials/holoscan_flow_benchmarking/benchmark.py
+++ b/tutorials/holoscan_flow_benchmarking/benchmark.py
@@ -97,9 +97,17 @@ def main():
         "--holohub-application",
         type=str,
         required=False,
-        help="name of HoloHub application to run. It runs the cpp version of an application\n\
-To run a python application, use --run-command option.",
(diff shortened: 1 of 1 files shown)


sohamm17 (commented):
Minor suggestion. Otherwise, LGTM. Thank you for the change!

PR #271, opened 2024-04-09
Infra: Create `benchmarks` folder

tbirdso:
# Motivation

We would like to better highlight existing benchmarking operations in HoloHub and provide a path to encourage development of new performance-related measures to emphasize Holoscan SDK for low latency applications.

## Overview

This PR creates a `benchmarks` folder at the top level to manage HoloHub projects focused on pipeline performance. `holoscan_flow_benchmarking` and `model_benchmarking` projects are migrated to this folder to help new users explore Holoscan benchmarking practices.

## Additional Notes

Additional effort is required to support benchmark applications in the `dev_container` and `run` infrastructure scripts. For now `model_benchmarking` applications may be run directly via their built executables in the Holoscan dev container.


benchmarks/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This directory contains applications, tools, and instructions to help measure
+the performance of pipelines built with Holoscan SDK.
+
+## Contents
+
+Any of the following may be covered under a benchmark project:
+- Holoscan SDK performance;
+- HoloHub application performance;
+- Tools for benchmarking Holoscan SDK applications not maintained in HoloHub.
```
sohamm17, line 11: it says `not maintained in HoloHub`. do you mean it would be maintained outside of Holohub?

sohamm17 (changes requested):
@tbirdso Creating a new directory in benchmarking would also mean that we need to change the paths in many places, for example, in `endoscopy_tool_tracking` where `holoscan_flow_benchmarking` is being used in `cmake` scripts for CI/CD.

sohamm17:
https://github.com/nvidia-holoscan/holohub/blob/f40089268bf6cdd522a77eff19a5e6301cbc79ff/applications/endoscopy_tool_tracking/cpp/CMakeLists.txt#L124

tbirdso:
> @tbirdso Creating a new directory in benchmarking would also mean that we need to change the paths in many places, for example, in `endoscopy_tool_tracking` where `holoscan_flow_benchmarking` is being used in `cmake` scripts for CI/CD.

Thanks, I wasn't aware of that. On review it looks like `endoscopy_tool_tracking` is the main application affected, will update paths in this PR.

benchmarks/README.md
tbirdso: Yes, here I mean to say that developers can use benchmarking tools to profile their own custom applications without necessarily adding those custom applications to HoloHub.

Do you have suggestions on how we can update the wording here to be more clear to new users?

sohamm17:
```suggestion
- Tools for benchmarking non-Holohub Holoscan SDK applications.
```

tbirdso:
@sohamm17 I've updated paths throughout HoloHub to reflect the `holoscan_flow_benchmarking` migration. Please re-review when you have a moment. Thanks!

sohamm17 (approved):

PR #272, opened 2024-04-10
Support passing Condition and Resource arguments as a positional argument to all provided Python operators

grlee77:
Currently there is a limitation in Python operators that wrap an underlying C++ operator that conditions like `CountCondition` or `PeriodicCondition` cannot be passed directly. This is being resolved for Holoscan SDK's built-in operators in a future release (v2.1). We can apply the same update to the operators provided here on Holohub (this approach is compatible with older releases back to at least v0.6).

For example, prior to this change, to have `EmergentSourceOp` call compute only 10 times one could add a `CountCondition` only after the operator was constructed like this:

```py
source = EmergentOp(self, name="emergent_source", **self.kwargs("emergent"))
source.add_arg(CountCondition(self, count=10))
```

but after the change made in this MR any conditions can be passed positionally to the constructor (as is currently already supported for native Python operators)
```py
source = EmergentOp(self, CountCondition(self, count=10), name="emergent_source", **self.kwargs("emergent"))
```

The operators updated in this MR are:
- AdvNetworkOpRx
- AdvNetworkOpTx
- EmergentSourceOp
- LSTMTensorRTInferenceOp
- OpenIGTLinkRxOp
- OpenIGTLinkTxOp
- orsi::FormatConverterOp
- orsi::SegmentationPreprocessorOp
- orsi::SegmentationPostprocessorOp
- orsi::OrsiVisualizationOp
- ProhawkOp
- QCAPSourceOp
- ToolTrackingPostprocessorOp
- VisualizerICardioOp
(This should be all of the Python operators that are wrapping an underlying C++ operator)

### DIFF
--- a/operators/advanced_network/python/adv_network_rx_pybind.cpp
+++ b/operators/advanced_network/python/adv_network_rx_pybind.cpp
@@ -31,7 +31,6 @@
 #include <holoscan/core/operator_spec.hpp>
 #include <holoscan/core/resources/gxf/allocator.hpp>
 
-
 using std::string_literals::operator""s;
 using pybind11::literals::operator""_a;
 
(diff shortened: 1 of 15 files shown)


gigony (approved):
Thanks @grlee77 !
Looks good to me!

tbirdso:
> this approach is compatible with older releases back to at least v0.6

Thanks @grlee77 ! Could you please confirm whether each of the updated operators/apps are tested with at least Holoscan SDK v0.6 per their metadata files?

grlee77:
> > this approach is compatible with older releases back to at least v0.6
> 
> Thanks @grlee77 ! Could you please confirm whether each of the updated operators/apps are tested with at least Holoscan SDK v0.6 per their metadata files?

So far, I only tested a subset of the operators with SDK v1.0.3. It should also work with v0.6 as I don't think this change relies on any APIs that were new to 1.0, but let me check.

grlee77:
For further testing, I built the `endoscopy_tool_tracking`, `multi_ai_ultrasound`, `orsi_segmentation_ar`, `orsi_in_out_body` and `orsi_multi_ai_ar` applications with **v0.6** and they are still working as expected.

PR #275, opened 2024-04-15
Infra: Update projects with enhanced data schema requirements

tbirdso:
Changes:
- Adds enhanced data schema requirement to enforce clearer descriptions in project metadata.
- Updates applications and operators to comply with new "data" field schema requirements.


jjomier (approved):
Looks very good - thanks for taking care of that!

PR #280, opened 2024-04-18
remove explicit data download step from Orsi application run instructions

grlee77:
The [CMakeLists.txt](https://github.com/nvidia-holoscan/holohub/blob/main/applications/orsi/CMakeLists.txt#L42-L56) includes commands to automatically download and extract the data to the expected path at build time. Given that, I don't think we should explicitly list it as an initial step in the instructions within the README


tbirdso (approved):
Good catch!

PR #282, opened 2024-04-19
Fix typo in basic CV-CUDA example

grlee77:
This MR fixes a typo in the `compose` method of the CV-CUDA app where the concrete "app" variable defined in `__main__` was being used instead of `self`. This happened to work, but would lead to the application breaking if the variable  in `__main__` was renamed.

I did a `grep` of other applications and fortunately did not find other instances of this mistake.


tbirdso (approved):
Good catch!

PR #284, opened 2024-04-24
App: Fix for data pollution among volume rendering apps

tbirdso:
## Description

Updates `volume_rendering_xr` to avoid data path conflicts with `volume_rendering`.

Under previous behavior `volume_rendering` and `volume_rendering_xr` both downloaded sample data to the `volume_rendering` data folder. The sample data for each app respectively included mono or stereoscopic view configurations. As a result, when one app was built first, and then the other app was built and run, the wrong configuration could be used, leading to unexpected visual data representations without an obvious immediate cause.

The data folder could not be cleared out with `./run clear_cache` as typically recommended for debugging, and the issue appeared to lie in app behavior when in fact it was due to configuration data pollution.

Under updated behavior `volume_rendering_xr` checks and downloads to its own dedicated folder, which significantly reduces the risk of data pollution and consequent rendering misbehavior.

## Previous Behavior

If `volume_rendering` was built before `volume_rendering_xr`, the latter app would attempt to render the volume (but not the bounding box) for a monoscopic view.

Screenshot from 2024-04-22 18-20-28

If `volume_rendering_xr` was built before `volume_rendering`, the latter app would render stereoscopically with a visual divide among the top and bottom halves of the viewport.

## Updated Behavior

`volume_rendering_xr` data downloads into its own data folder and renders appropriately.

Screenshot from 2024-04-18 12-46-07

### DIFF
--- a/applications/volume_rendering_xr/CMakeLists.txt
+++ b/applications/volume_rendering_xr/CMakeLists.txt
@@ -68,13 +68,13 @@ add_dependencies(volume_rendering_xr volume_rendering_xr_config_yaml)
 option(HOLOHUB_DOWNLOAD_DATASETS "Download datasets" ON)
 if(HOLOHUB_DOWNLOAD_DATASETS)
   include(holoscan_download_data)
-  holoscan_download_data(volume_rendering
+  holoscan_download_data(volume_rendering_xr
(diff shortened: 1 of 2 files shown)


AndreasHeumann (approved):
Thanks for the fix, looks good.

PR #285, opened 2024-04-24
App: Bump `volume_rendering_xr` firmware requirement

tbirdso:
Bump the documented Magic Leap 2 device firmware requirement to v1.6.0.

Follows local testing where the remote rendering application could not connect on firmware v1.5.0, but connected without issue when bumped to the latest firmware (v1.6.0).

Other software requirements are unaffected.


AndreasHeumann (approved):
Looks good, thanks.

PR #305, opened 2024-05-07
Doc: Add build cache discussion

tbirdso:
Adds discussion on build cache cleanup in response to user feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,24 @@ endoscopy application:
 The run script reads the "run" command from the metadata.json file for a given application and runs from the "workdir" directory.
 Make sure you build the application (if applicable) before running it.
 
+# Cleanup
+
+HoloHub applications share a build directory by default. We recommend running the command below to reset your build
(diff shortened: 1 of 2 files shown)


jjomier:
Now that we don't have SAMPLE_APPS grouping, we could change the build directory by default to be build/[appname]. That would avoid any potential issues. There are two caveats:
1. The launch command would have to match the name of the application and if the build directory is set the launch command would not work (unless the build directory is specified - this is a current issue actually since the launch command will not work if the build directory is not named "build")
2. Two applications using the same operator will build the operator multiple times - this is probably OK since some compilation might be different based on the container (for instance).

tbirdso:
> Now that we don't have SAMPLE_APPS grouping...

@jjomier I agree that we should move in that direction. In the meantime I'd like to move forward with this PR to fill a gap in documentation based on existing behavior. What do you think?

jjomier (approved):

PR #309, opened 2024-05-08
Remove the notion of sample applications tag in CMake

jjomier:
Currently Holohub has the notion of SAMPLE_APPS which allow to bundle initial applications into a group and build them together. In practice, users and developers only target one applicaiton at the time.


tbirdso (approved):
LGTM

PR #313, opened 2024-05-09
Switching to separate build directories for each applications

jjomier:
Instead of building all applications in the same `build `directory switching to `build/<app>`

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,8 +172,7 @@ Make sure you build the application (if applicable) before running it.
 
 # Cleanup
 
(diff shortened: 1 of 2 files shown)


tbirdso:
@jjomier You had mentioned compatibility issues with `run launch`. Is that addressed by these changes?

jjomier:
> @jjomier You had mentioned compatibility issues with `run launch`. Is that addressed by these changes?

Yes it is (see modification to the launch command)

tbirdso (approved):

PR #314, opened 2024-05-09
App: Add runtime container procedure for `volume_rendering_xr`

tbirdso:
Adds custom Dockerfile, scripts, and documentation to support releasing
`volume_rendering_xr` in a standalone, self-contained container with a
command line interface.

The primary advantage of packaging a release container is to allow
sharing a built `volume_rendering_xr` application with another machine
or another user with minimum setup. After following the README to build
the container, the release container can then be shared as a compressed
file or from a container registry.

The `volume_rendering_xr` release container is based on the
`nvidia-cuda-runtime` container publicly available from NGC. The release
container is approximately 3GB in size, while the build container is
approximately 10GB in size.

Currently only one volume dataset is currently available for packaging. Other
datasets could be added locally before packaging is carried out.

This work is based on prior development in collaboration between the
NVIDIA Holoscan and Magic Leap teams. Verified on x86_64 and IGX (aarch64) platforms.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ CTestConfig.cmake
 
 # Utilities
 aggregate_metadata.json
+/dist/
+/install/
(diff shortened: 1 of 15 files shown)


jjomier (approved):

AndreasHeumann (approved):
Looks good to me.

PR #321, opened 2024-05-13
App: Update run commands for per-app build folder paths

tbirdso:
Update `body_pose_estimation` and `tao_peoplenet` to use HoloHub pre-defined path keywords instead of relying on relative pathing between the build and source directories.

As of 3d4100ac HoloHub now defines a new build subfolder for each project instead of using one overarching build for the entire project.

Verified with `body_pose_estimation` command:
```
dev_container build_and_run body_pose_estimation
```

EDIT: I checked that the above command launched, but not that the app ran fully. As reported below, there is still an issue with data paths that will be resolved in a subsequent PR.

### DIFF
--- a/applications/body_pose_estimation/metadata.json
+++ b/applications/body_pose_estimation/metadata.json
@@ -42,7 +42,7 @@
 			]
 		},
 		"run": {
-			"command": "python3 ../applications/body_pose_estimation/body_pose_estimation.py",
+			"command": "python3 <holohub_app_source>/body_pose_estimation.py",
 			"workdir": "holohub_bin"
 		}
 	}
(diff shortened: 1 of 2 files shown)


jjomier (approved):

ibstewart:
This still has the issue that the default data path used by applications is incorrect. Many apps are doing this:

```
data = os.environ.get("HOLOSCAN_DATA_PATH", "../data")
```

But the working directory used for `holohub_bin` from the metadata and `run` script is `build/<app_name>`, so the default path above would need to be changed to `../../data`

tbirdso:
> This still has the issue that the default data path used by applications is incorrect.

@ibstewart thanks for catching this. You're right, I agree that issue stems from https://github.com/nvidia-holoscan/holohub/commit/3d4100ac6012302579c6e7b1740b52e3216f17d7. Since it impacts app runtime rather than launch, I'd like to pursue in a separate discussion and PR: https://github.com/nvidia-holoscan/holohub/issues/323

PR #322, opened 2024-05-13
App: Update `cuda_quantum` application for `build_and_run` compatibility

tbirdso:
Changes:
- Updates `cuda_quantum` run command to execute from the app source directory. Under previous behavior the run command failed because `cuda_quantum.py` was not in the working directory.
- Install requirements as part of the run command. We typically prefer to use a dedicated Dockerfile for HoloHub app environments, but in doing so `cuda_quantum` dependencies install with root permissions and are inaccessible when running the container as a non-superuser by default. Instead we can install dependencies each time for demonstration.

Validated on x86_64.

### DIFF
--- a/applications/cuda_quantum/metadata.json
+++ b/applications/cuda_quantum/metadata.json
@@ -32,8 +32,8 @@
             "cuda_quantum": "^0.4.0"
         },
         "run": {
-            "command": "python3 cuda_quantum.py",
-            "workdir": "holohub_bin"
+            "command": "pip install -r requirements.txt && python3 cuda_quantum.py",
+            "workdir": "holohub_app_source"
         }
     }
 }


applications/cuda_quantum/metadata.json
```diff
@@ -32,8 +32,8 @@
             "cuda_quantum": "^0.4.0"
         },
         "run": {
-            "command": "python3 cuda_quantum.py",
-            "workdir": "holohub_bin"
+            "command": "pip install -r requirements.txt && python3 cuda_quantum.py",
```
jjomier, line 35: I don't think we want pip install as part of the run. We should create a separate Dockerfile.

tbirdso: See the PR description -- I also would prefer a Dockerfile, but `cuda_quantum` appears to have a transitive dependency on the [wheel-axle-runtime](https://pypi.org/project/wheel-axle-runtime/) package, which creates a conflict when running the app with different permissions than the installation environment. When I tried using a Dockerfile I saw permissions errors when running as a non-root user inside the container.

We could investigate further, but IMO it's worth making this exception to the rule for now to get the app working with `build_and_run`.

jjomier (approved):

Issue #323, opened 2024-05-13
HoloHub app default data paths are misaligned with source/build folder


tbirdso:
Report: https://github.com/nvidia-holoscan/holohub/pull/321#issuecomment-2108738200

Many apps are doing the following, which may be out of date as of https://github.com/nvidia-holoscan/holohub/commit/3d4100ac6012302579c6e7b1740b52e3216f17d7:
```
data = os.environ.get("HOLOSCAN_DATA_PATH", "../data")
```

tbirdso:
@jjomier I'm surprised that I do not see `HOLOSCAN_DATA_PATH` defined in the default HoloHub container. What is the recommended path for apps to get the default data directory? Is there another environment variable they should use?

tbirdso:
Following internal discussion:

Several HoloHub apps use os.environ.get("HOLOSCAN_DATA_PATH", "../data") to find the common data folder. We are seeing issues because
- HOLOSCAN_DATA_PATH should be HOLOHUB_DATA_PATH;
- HOLOHUB_DATA_PATH isn't actually exported in the run launch command, only used as a keyword replacement for command line args; and
- ../data assumes a relative path relationship between the source root and build root, which is an unstable interface;
-  ../data is not always a valid path, depending on what working directory is set in the metadata.json run command field.

We can fix this as follows:
- HOLOSCAN_DATA_PATH -> HOLOHUB_DATA_PATH
- Export HOLOHUB_DATA_PATH in run launch as we currently do with PYTHONPATH
- Make an educated guess on a per-app, per-working-directory basis as to what the fallback relative data path should be. We typically will not want to use the fallback and will prefer HOLOHUB_DATA_PATH instead.

PR #324, opened 2024-05-13
App: Update data path environment variable

tbirdso:
Updates HoloHub apps as follows:
- Use `HOLOHUB_DATA_PATH` instead of `HOLOSCAN_DATA_PATH`, which is not set in HoloHub
- Update `run` script to set `HOLOHUB_DATA_PATH` as an environment variable at runtime. The script already parses the data path from `CMakeCache.txt` , but previously did not set the corresponding environment variable.

Fixes https://github.com/nvidia-holoscan/holohub/issues/323


jjomier (approved):

tbirdso:
Verified end-to-end with `body_pose_estimation`, able to see webcam overlay again with these changes.

PR #326, opened 2024-05-14
Added import for Holoscan and GXF for interop tutorial

jjomier:
Added missing imports for the interop tutorial for more clarity.


tbirdso (approved):
Good catch. I have not tried running the tutorial.

PR #327, opened 2024-05-14
Update GXF ENC/DEC support to GXF 4.0

apatole:
Holoscan has been upgraded to use GXF 4.0. As a result, we see version mismatch errors while running H.264 sample apps. There was also a bug related to iGPU and dGPU detection which is now fixed with the latest version of GXF ENC / DEC and related latest DeepStream dependency.

This request updates Holoscan H.264 support to use GXF Encode / Decode extension from GXF 4.0. It also updates the DeepStream dependency to the required version.

This change is successfully tested it on IGX with dGPU and x86 with dGPU dev machine.

### DIFF
--- a/operators/video_decoder/video_decoder_context/video_decoder_context.cpp
+++ b/operators/video_decoder/video_decoder_context/video_decoder_context.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2023-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


tbirdso (approved):
The changes look good to me at a glance. Please review linting/signing CI issues.

jjomier (approved):

operators/video_decoder/video_decoder_context/video_decoder_context.cpp
```diff
@@ -24,6 +24,11 @@ void VideoDecoderContext::setup(ComponentSpec& spec) {
              "async_scheduling_term",
              "Asynchronous Scheduling Condition",
              "Asynchronous Scheduling Condition");
+  spec.param(device_id_,
```
AndreasHeumann, line 27: It seems this parameter is never used in the code anywhere?

apatole: Hi @AndreasHeumann , This is newly added parameter which is set to '0' by default and which will work for most of the cases.  This parameter can be used to specify Cuda Device ID and required to be set only incase of multiple Cuda devices and app needs to use specific Cuda device with the ID other than '0'. Hoping it answers your question, let me know otherwise, thanks.

AndreasHeumann: Thanks @apatole. I tried to find the line where `device_id_` de-referenced and used, but I could not. Could you point me to the code line?
I also noticed that that the parameter is not mentioned in the Readme.

PR #337, opened 2024-05-16
Add DDS documentation for containers and arm

ibstewart:
Also adds a note in the body_pose_estimation makefile about why we use the qos_profiles.xml file from the dds_video app.


tbirdso (approved):
Looks great, thanks for adding this @ibstewart !

PR #339, opened 2024-05-17
DDS CMake improvement

jjomier:
Modify CMake processing to avoid finding package holoscan when not using DDS


tbirdso:
@jjomier I see a build error when attempting to build DDS with this branch:

```
fatal error: ../../../operator_util.hpp: No such file or directory
```

It looks like this PR goes the extra mile to align DDS to HoloHub infrastructure, which I agree we should pursue, but does more work than required to address the issue at hand. I'm going to open a companion PR with the minimum fix and then I will help track down the organization issue here.

jjomier:
> @jjomier I see a build error when attempting to build DDS with this branch:
> 
> ```
> fatal error: ../../../operator_util.hpp: No such file or directory
> ```
> 
> It looks like this PR goes the extra mile to align DDS to HoloHub infrastructure, which I agree we should pursue, but does more work than required to address the issue at hand. I'm going to open a companion PR with the minimum fix and then I will help track down the organization issue here.

We need to target include the base operator directory - which was removed. I'm pushing a fix

jjomier:
> > @jjomier I see a build error when attempting to build DDS with this branch:
> > ```
> > fatal error: ../../../operator_util.hpp: No such file or directory
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It looks like this PR goes the extra mile to align DDS to HoloHub infrastructure, which I agree we should pursue, but does more work than required to address the issue at hand. I'm going to open a companion PR with the minimum fix and then I will help track down the organization issue here.
> 
> We need to target include the base operator directory - which was removed. I'm pushing a fix

Actually just need to point to the right path.

tbirdso:
```
fatal error: dds/shapes/subscriber.hpp: No such file or directory
```

operators/dds/video/dds_video_publisher/python/publisher_pydoc.hpp
jjomier, line 1: Should be dds_video_publisher

tbirdso: Ah, I see the "#include" statement I referenced was in fact the issue with respect to naming conventions. Fixed.

jjomier:
LGTM! Thanks for fixing @tbirdso

ibstewart:
It looks good to me, too, and I've tested it on my end without any issues. Thanks!

PR #340, opened 2024-05-17
Op: Fix HoloHub compatibility breakage from DDS operators

tbirdso:
Issue: DDS operators provide custom flags to enable or disable DDS sources, so that DDS operators do not always build with other HoloHub apps unless requested. However, the `find_package(holoscan 2.0)` calls in DDS operator subdirectories previously existed outside of these if/endif blocks. As a result, HoloHub always tried to find holoscan 2.0 when building any application, and failed if it was not found.

Workaround: This small fix moves `find_package(holoscan 2.0)` calls inside of if/endif blocks so that DDS operators do not impact building other HoloHub apps if DDS is not turned on.

A more complete fix will be pursued in
https://github.com/nvidia-holoscan/holohub/pull/339 to align DDS operators to standard HoloHub CMake build infrastructure.

Tested and verified by building and running `dds_video` with HSDK 2.0 (bare metal, x86_64) and by building `endoscopy_tool_tracking` with an HSDK 1.0.3 internal branch.

### DIFF
--- a/operators/dds/base/CMakeLists.txt
+++ b/operators/dds/base/CMakeLists.txt
@@ -13,14 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 3.24)
-project(dds_operator_base)
-
-find_package(holoscan 2.0 REQUIRED CONFIG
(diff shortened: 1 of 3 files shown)


tbirdso:
Closing in favor of #339, which resolves the compatibility issue and also updates DDS operators to comply with HoloHub infrastructure.

PR #346, opened 2024-05-22
Fix markdown link syntax error in multiai readme

aim-nara:


tbirdso (approved):
Thanks for contributing!

PR #348, opened 2024-05-23
Fix version number in README.md

JohnMoon-VTS:
Currently, the README.md mentions the `dev_container` script uses the Holoscan SDK container v1.0, but a recent commit (8894dd4) bumped the version used to v2.0.

Update this reference in the README.md.


tbirdso (approved):
Thanks for finding this @JohnMoon-VTS !

PR #349, opened 2024-05-23
Infra: Fix /dev/snd error when no sound card is present

JohnMoon-VTS:
Currently, if you run `./dev_container launch` on a system without a sound card, you get:

% ./dev_container launch
find: '/dev/snd': No such file or directory

As a sound card shouldn't be required here, let's add a check for `/dev/snd` before running that `find` command.

### DIFF
--- a/dev_container
+++ b/dev_container
@@ -491,13 +491,15 @@ launch() {
         fi
     done
 
-    # Find all audio devices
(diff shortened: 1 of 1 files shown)


jjomier (approved):
LGTM - thanks for the fix!

NigelNelson (approved):
Looks good, thank you for the catch!

PR #350, opened 2024-05-24
Update Body Pose Estimation

wendell-hom:


tbirdso (commented):
Code looks good to me! Please confirm SWIPATs, after that this looks ready for merge.

tbirdso:
Thanks for splitting here and https://github.com/nvidia-holoscan/holohub/pull/354 . Please rebase on `main`

applications/body_pose_estimation/README.md
```diff
@@ -17,6 +17,8 @@ The model is downloaded when building the application.
 
 This application downloads a pre-recorded video from [Pexels](https://www.pexels.com/video/a-woman-showing-her-ballet-skill-in-turning-one-footed-5385885/) when the application is built.
 
+> **_NOTE:_** For each dataset an user elects to use, the user is responsible for checking if the dataset license is fit for the intended purpose.
```
tbirdso, line 20: A couple of questions:
- Is the Pexels "twirl" MP4 the only dataset provided with this app, or are there others?
- Under what license is the Pexels video distributed?
- Can we provide a link to the Pexels video in the README and/or in `metadata.json`?

whom3: @tbirdso 

- Yes, the "twirl" mp4 is the only dataset we are making use of in this application
- good point, I've added a link to the pexels license terms 
- sure, there's already a link to the video in the README.  Just updated the metadata.json to reference the video as well

tbirdso: Thanks! Current language suggests multiple datasets, suggest a small update:

```suggestion
> **_NOTE:_** The user is responsible for checking if the dataset license is fit for the intended purpose.
```

dev_container
```diff
@@ -712,7 +712,7 @@ launch() {
         ${ucx_opt} \
         ${docker_opts} \
         ${img} \
-        "$@"
+        -c "$@"
```
tbirdso, line 715: On review, I'd request that we maintain the existing "$@" behavior. Reasoning:
- The proposed behavior introduces a regression where `dev_container launch` no longer launches into an interactive shell and instead exits immediately.
- `-c` itself is a forwarded argument rather than a part of the `docker run` command. Under typical execution the trailing arguments here are forwarded first to the `/opt/nvidia/nvidia_entrypoint.sh` container script and then on to the bash `exec` function. It makes sense that an outer call to `dev_container launch` should include `-c` among a larger list of arguments to forward to the container.
- On inspection `-c` itself doesn't seem to be required to run a forwarded command with `exec`. If you're seeing issues with multi-arg forwarding, maybe the cause lies elsewhere?

Based the bash [manpage](https://manpages.ubuntu.com/manpages/trusty/en/man1/bash.1.html) I don't anticipate we have a strong use case for alternative arguments to `exec`, but it would be good to avoid changing default `dev_container launch` interactive behavior. Are you seeing in your testing that multi-arg forwarding fails without this update?

tbirdso (changes requested):

dev_container
wendell-hom: @tbirdso good catch for the `dev_container launch` issue, thank you!  

I've reverted that change and moved it inside `dev_container build_and_run`. 
Yes, some change is needed in order to get multi-arg to be forwarded correctly after recent changes to how the args were being forwarded.  The current proposed changes in dev_container now seem to work both interactively and with multi-arg forwarding.

tbirdso: Verified the update works on my system, thanks!

tbirdso (approved):
Looks good to me!

PR #351, opened 2024-05-24
Infra: Rename `build_and_run` `--extra_args` to `--run_args`

tbirdso:
Rename the `build_and_run` CLI option from `--extra_args` to `--run_args` following internal discussion.

This change disambiguates the option to make it clear that extra arguments passed in will be applied to the last infrastructure command that runs the application, i.e. `./run launch`.

Tested with `./dev_container build_and_run body_pose_estimation --run_args -s v4l2`


wendell-hom:
@tbirdso I've added the renaming of the extra args in this PR https://github.com/nvidia-holoscan/holohub/pull/350
Also needed to put quotes around it to make sure that we can pass more than one extra run arg to the `./run launch` command, e.g., when user needs to run the command like:
`./dev_container build_and_run --run_args "--source v4l2 --video_device /dev/video1"`

tbirdso:
Closing in favor of #350

PR #357, opened 2024-05-29
HoloChat Update

NigelNelson:
This MR makes the following changes:

- Convert local LLM to use mock OpenAI server so the LLM class can use the standardized OpenAI API
- Integrate HoloChat with the HoloHub container and the dev_container script
- Add the option to run with the remote NIM API for Llama-3-70b
- Update the version of the Holoscan User Guide used


tbirdso (approved):
Updates look great!

PR #358, opened 2024-05-29
Infra: Relax select `dev_container` arg parse restrictions

tbirdso:
The recent commit 4b4243f0b tightened restrictions on argument parsing in `dev_container` functions with the goal of highlighting erroneous script input. This change relaxes those restrictions on a small subset of arguments to allow argument forwarding to proceed without errors.

Changes:
- Forward trailing arguments to `dev_container launch` after `--` to the container shell. This mimics behavior prior to 4b4243f0b where trailing arguments were forwarded to the container, but with the added restriction that `--` must be explicitly specified to prevent unintended argument forwarding.
- Allow values following `--docker_opts` and `--run_args` flags to start with `--`. It is reasonable for those values to contain additional flags that will be forwarded to the appropriate `docker run` or `run launch` command.
- Fix an issue with string parsing for commands forwarded to the container with `dev_container build_and_run`.

Validated with HoloHub apps:
- `dev_container build_and_run body_pose_estimation`
- `dev_container build_and_run volume_rendering`

### DIFF
--- a/dev_container
+++ b/dev_container
@@ -390,6 +390,7 @@ Usage: ./dev_container launch [options]
     --add-volume : Mount additional volume
     --as_root  : Run the container with root permissions
     --docker_opts : Additional options to pass to the Docker launch
+    -- : Any trailing arguments after "--" are forwarded to `docker run`
     '
 }
 launch() {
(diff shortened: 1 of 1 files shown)


NigelNelson:
Checked with HoloChat using `dev_container build_and_run holochat --local` and it worked.

NigelNelson (approved):
LGTM!

PR #362, opened 2024-05-30
App: Fixups for Holochat local run

tbirdso:
Address minor Holochat issues encountered when running locally.

Fixes issues encountered on first HoloChat local run:

```
python3: can't open file '/workspace/holohub/build/holochat/build_holoscan_db.py': [Errno 2] No such file or directory
```

```
NameError: name 'SimpleNamespace' is not defined
```

With these fixes applied the DB build stage appears to complete. However, I can't fully confirm whether the local run is successful after that stage as my GPU does not meet hardware requirements.

### DIFF
--- a/applications/holochat/build_holoscan_db.py
+++ b/applications/holochat/build_holoscan_db.py
@@ -15,6 +15,7 @@
 
 import os
 import re
+from types import SimpleNamespace
 
 import yaml
 from langchain.docstore.document import Document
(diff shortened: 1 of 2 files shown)


NigelNelson:
Thank you for the catches. I deleted my local builds and am building from scratch to test this now.

NigelNelson:
The changes work from a clean build. However, I noticed `--local` is hardcoded in the metadata.json, could you remove this as well?

NigelNelson (approved):
Thanks for the catches, I tested locally and it works!

tbirdso:
>I noticed --local is hardcoded in the metadata.json

Huh, I thought we removed that in the last PR. Thanks for that fix and for verifying!

PR #364, opened 2024-05-30
App: Vista3D NIM Sample App

mocsharp:
Sample application that connects to the Vista-3D NVIDIA Inference Microservice (NIM). 

This PR downloads the sample dataset and calls the Vista 3D NIM API hosted by NVIDIA (build.nvidia.com).
The API is expected to return an NRRD file. The user may use 3D Slicer to display the dataset and the segmentation volume as described in the README.


operators/volume_loader/nrrd_loader.cpp
```diff
@@ -148,56 +148,96 @@ bool load_nrrd(const std::string& file_name, Volume& volume) {
...
+        holoscan::log_error("NRRD failed to uncompress {}, inflate failed with error code {}",
+                            data_file_name,
+                            result);
+        return false;
+      }
+    } else {
+      holoscan::log_error("NRRD unsupported file format");
     }
-  } else {
-    file.read(reinterpret_cast<char*>(data.get()), data_size);
```
AndreasHeumann, line 200: I think this path for uncompressed data is missing from the new code.

operators/volume_loader/nrrd_loader.cpp
```diff
@@ -148,56 +148,96 @@ bool load_nrrd(const std::string& file_name, Volume& volume) {
...
+    file.open(file_name, std::ios::in);
+    if (!file.is_open()) {
+      holoscan::log_error("NRRD could not open {}", file_name);
+      return false;
+    }
+    std::string line;
+    while (std::getline(file, line)) {
+      if (file.tellg() != -1) { byte_skip = file.tellg(); }
+    }
+  }
```
AndreasHeumann, line 178: Could we integrate this code into the loop above? Opening the file again and reading the header to get the header size seems to be inefficient.

operators/volume_loader/nrrd_loader.cpp
```diff
@@ -148,56 +148,96 @@ bool load_nrrd(const std::string& file_name, Volume& volume) {
...
+    }
+    std::ifstream file;
+    file.open(file_name, std::ios::in | std::ios::binary);
+    if (!file.is_open()) {
+      holoscan::log_error("NRRD could not open {}", data_file_name);
+      return false;
+    }
 
-  std::ifstream file;
+    file.seekg(byte_skip, std::ios_base::beg);
```
AndreasHeumann, line 195: Could we open the file only once? Reopening seems to be inefficient.

operators/volume_loader/nrrd_loader.hpp
```diff
@@ -28,6 +29,11 @@ bool is_nrrd(const std::string& file_name);
 
 bool load_nrrd(const std::string& file_name, Volume& volume);
 
+std::vector<double> parse_vector(std::string str);
+
+void trim(std::string& s);
+
+std::vector<std::string> split_string_by_space(std::string str);
```
AndreasHeumann, line 36: These helper functions don't need to be exposed in the header.

operators/volume_loader/volume.hpp
```diff
@@ -54,6 +54,10 @@ class Volume {
   std::array<bool, 3> flip_axes_{false, false, false};
   /// frame duration
   std::chrono::duration<float> frame_duration_;
+  /// space origin
+  std::array<double, 3> space_origin_{0.0, 0.0, 0.0};
+  /// space directions
+  std::vector<std::array<double, 3>> space_directions_;
```
AndreasHeumann, line 60: It seems these are not used but the volume renderer. Do you plan to parse these in the renderer?

operators/volume_loader/nrrd_loader.cpp
```diff
@@ -234,4 +274,40 @@ bool load_nrrd(const std::string& file_name, Volume& volume) {
...
+
+void trim(std::string& str) {
+  str.erase(str.begin(), std::find_if(str.begin(), str.end(), [](unsigned char ch) {
+              return !std::isspace(ch);
+            }));
+  str.erase(
+      std::find_if(str.rbegin(), str.rend(), [](unsigned char ch) { return !std::isspace(ch); })
+          .base(),
+      str.end());
+}
```
AndreasHeumann, line 298: Aren't spaces already removed in line 67?

operators/volume_loader/nrrd_loader.cpp
```diff
@@ -234,4 +274,40 @@ bool load_nrrd(const std::string& file_name, Volume& volume) {
   return true;
 }
 
+std::vector<double> parse_vector(std::string str) {
+  std::vector<double> result;
+  trim(str);
+  // ensures the string is surround by parenthesis and then remove them
+  if ((str[0] != '(') || (str[str.length() - 1] != ')')) return result;
```
AndreasHeumann, line 281: Should this throw an error if there are no parenthesis?

operators/volume_loader/dual_type_receiver.hpp
```diff
@@ -0,0 +1,69 @@
...
+template <typename T>
+struct is_shared_ptr<std::shared_ptr<T>> : std::true_type {};
+
+/*
+ * Any object emitted from a Python operator on a port (except Tensor) is tracked by the
+ * GILGuardedObject. With that we can detect if the object received by the C++ operator
+ * is coming from Python and do the conversion to a C++ type.
+ * This needs to be moved to Holoscan SDK.
+ * See discussion https://nvidia.slack.com/archives/C03E643SK89/p1701875573021209.
+ */
```
AndreasHeumann, line 37: This is actually a hack I implemented. I don't think this should be added here. @grlee77 implemented support for this in Holoscan SDK 2.1. We should require .2.1 for the app and remove this and the `gil_guarded_object.hpp` file.

operators/volume_loader/volume_loader.cpp
```diff
@@ -56,14 +77,20 @@ void VolumeLoaderOp::compute(InputContext& input, OutputContext& output,
...
       static_cast<nvidia::gxf::Entity&>(entity).add<nvidia::gxf::Tensor>("volume").value();
 
-  if (is_nifty(file_name_.get())) {
-    load_nifty(file_name_.get(), volume);
-  } else if (is_mhd(file_name_.get())) {
-    load_mhd(file_name_.get(), volume);
-  } else if (is_nrrd(file_name_.get())) {
-    load_nrrd(file_name_.get(), volume);
+  if (is_nifty(file_name)) {
+    if (!load_nifty(file_name, volume)) {
```
AndreasHeumann, line 81: Thanks for adding this check, totally forgot it!

applications/nvidia_nim/nvidia_nim_imaging/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+#     FetchContent_Populate(ClaraViz)
+#     set(CLARA_VIZ_PUBLIC_CMAKE_TOOLS_DIR "${claraviz_SOURCE_DIR}/cmake")
+#     add_subdirectory(${claraviz_SOURCE_DIR}/thirdparty ${claraviz_BINARY_DIR}/thirdpary)
+#     add_subdirectory(${claraviz_SOURCE_DIR}/src ${claraviz_BINARY_DIR}/src)
+# endif()
+# set(FETCHCONTENT_QUIET ON)
+
+# set(CLARA_VIZ_PUBLIC_CMAKE_TOOLS_DIR "${claraviz_SOURCE_DIR}/cmake")
+
+# find_package(clara_viz_renderer REQUIRED HINTS ${claraviz_SOURCE_DIR}/cmake)
```
AndreasHeumann, line 66: Commented out code, please remove.

applications/nvidia_nim/nvidia_nim_imaging/app.py
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        session = requests.Session()
+        response = session.post(self.base_url, headers=headers, json=payload)
+
+        response.raise_for_status()
+        seg_filename = os.path.abspath(f"{sample}_seg.nrrd")
+        with tempfile.TemporaryDirectory() as temp_dir:
+            z = zipfile.ZipFile(io.BytesIO(response.content))
+            z.extractall(temp_dir)
```
AndreasHeumann, line 117: If we add zip support to `nrrd` files we could avoid uncompressing here. Also other apps would benefit form it.

applications/nvidia_nim/nvidia_nim_imaging/app.py
```diff
@@ -0,0 +1,290 @@
...
+    headers = {
+        "Authorization": f"Bearer {api_key}",
+    }
+    response = requests.get(payload["image"], headers=headers)
+    response.raise_for_status()
+    nifti_filename = os.path.abspath("sample.nii")
+    buffer = io.BytesIO(response.content)
+    with gzip.GzipFile(fileobj=buffer) as gz_file:
+        with open(nifti_filename, "wb") as f:
+            f.write(gz_file.read())
```
AndreasHeumann, line 228: I think NIFTI supports compressed files, we could avoid this step.

applications/nvidia_nim/nvidia_nim_imaging/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+  include(holoscan_download_data)
+  holoscan_download_data(nvidia_nim_imaging
+    URL https://api.ngc.nvidia.com/v2/resources/nvidia/clara-holoscan/holoscan_volume_rendering_sample_data/versions/20230628/zip
+    DOWNLOAD_NAME holoscan_volume_rendering_sample_data_20230628.zip
+    URL_MD5 129a15d848b83c8f54988d1ee985edca
+    DOWNLOAD_DIR ${HOLOHUB_DATA_DIR}
+    ALL
+  )
+  add_dependencies(nvidia_nim_imaging nvidia_nim_imaging_data)
+endif()
```
AndreasHeumann, line 85: You probably need to create your own config file for the renderer. You can use the config.json file from the volume rendering sample as a reference. Keep the lighting and the camera and other settings. And for each segmented organ (index) add an entry to the `components` array of the `TransferFucntion` structure. `activeRegions` is the index of the organ in the segmentation file. 0 is usually the background, so the first organ could be 1 (depends on the segmentation algorithm). `diffuseColorStart` and `diffuseColorEnd` is the color, assign one color for each organ.
```        {
                "activeRegions": [
                    1
                ],
                "diffuseEnd": {
                    "x": 1,
                    "y": 0.0,
                    "z": 0.0
                },
                "diffuseStart": {
                    "x": 1,
                    "y": 0.0,
                    "z": 0.0
                },
                "specularEnd": {
                    "x": 1,
                    "y": 1,
                    "z": 1
                },
                "specularStart": {
                    "x": 1,
                    "y": 1,
                    "z": 1
                },
                "emissiveEnd": {
                    "x": 0,
                    "y": 0,
                    "z": 0
                },
                "emissiveStart": {
                    "x": 0,
                    "y": 0,
                    "z": 0
                },
                "emissiveStrength": 0,
                "opacity": 1,
                "opacityProfile": "SQUARE",
                "opacityTransition": 0.0,
                "range": {
                    "min": 0.0,
                    "max": 1.0
                },
                "roughness": 1.
            },
 ```

There is a section at the end of `TransferFunction`
```
        "densityScale": 999.9999999999998,
        "globalOpacity": 1,
        "gradientScale": 10,
        "shadingProfile": "BRDF ONLY",
        "blendingProfile": "MAXIMUM_OPACITY"
```
If the boundaries of the organs are too sharp or too blurry, decrease or increase `densityScale`.

**Review by AndreasHeumann — COMMENTED** [review]
Nice work! Left some comments.

File: operators/volume_loader/dual_type_receiver.hpp (hunk 8 above)
Comment (tbirdso) [reply operators/volume_loader/dual_type_receiver.hpp:37]: Note that we do currently require the `volume_rendering` application is backwards compatible with Holoscan SDK v1.0.3 for NVIDIA AI Enterprise related reasons. Maybe there is a way to default to the 2.1 functionality with include guards for backwards compatibility? Happy to chat further internally.

File: applications/nvidia_nim/nvidia_nim_imaging/CMakeLists.txt (hunk 10 above)
Comment (mocsharp) [reply applications/nvidia_nim/nvidia_nim_imaging/CMakeLists.txt:66]: Removed. Thank you @AndreasHeumann. So we do not need to include ClaraViz at all correct?

File: applications/nvidia_nim/nvidia_nim_imaging/app.py (hunk 12 above)
Comment (mocsharp) [reply applications/nvidia_nim/nvidia_nim_imaging/app.py:228]: Awesome! Thank you!

File: operators/volume_loader/dual_type_receiver.hpp (hunk 8 above)
Comment (mocsharp) [reply operators/volume_loader/dual_type_receiver.hpp:37]: Thank you @AndreasHeumann! I've removed `dual_type_receiver` and `gil_guarded_pyobject`.

File: operators/volume_loader/nrrd_loader.cpp (hunk 1 above)
Comment (mocsharp) [reply operators/volume_loader/nrrd_loader.cpp:200]: Thank you.

File: operators/volume_loader/nrrd_loader.cpp (hunk 7 above)
Comment (mocsharp) [reply operators/volume_loader/nrrd_loader.cpp:281]: Thank you!

File: operators/volume_loader/nrrd_loader.cpp (hunk 6 above)
Comment (mocsharp) [reply operators/volume_loader/nrrd_loader.cpp:298]: Line 67 removes all spaces for the parameter names (keys), not the values, right?

File: operators/volume_loader/nrrd_loader.hpp (hunk 4 above)
Comment (mocsharp) [reply operators/volume_loader/nrrd_loader.hpp:36]: Thank you!

File: operators/volume_loader/volume.hpp (hunk 5 above)
Comment (mocsharp) [reply operators/volume_loader/volume.hpp:60]: I need your help with that as I have no experience with it.

File: operators/volume_loader/nrrd_loader.cpp (hunk 6 above)
Comment (mocsharp) [reply operators/volume_loader/nrrd_loader.cpp:298]: I missed that. Thanks.

File: operators/volume_loader/nrrd_loader.cpp (hunk 2 above)
Comment (mocsharp) [reply operators/volume_loader/nrrd_loader.cpp]: Please review the changes; I also refactored the function into smaller ones. Thank you!

File: applications/nvidia_nim/nvidia_nim_imaging/CMakeLists.txt (hunk 10 above)
Comment (AndreasHeumann) [reply applications/nvidia_nim/nvidia_nim_imaging/CMakeLists.txt]: Correct, ClaraViz is a dependency of the volume renderer operator, no need to include it in the app.

File: applications/nvidia_nim/nvidia_nim_imaging/app.py (hunk 11 above)
Comment (mocsharp) [reply applications/nvidia_nim/nvidia_nim_imaging/app.py:117]: In that case, we must always assume there is a supported nrrd/nhdr file in the zip file.  I would rather create an operator that does decompression and lets the user emit the files they want for a downstream operator.

File: applications/nvidia_nim/nvidia_nim_imaging/README.md [hunk 14]
```diff
@@ -0,0 +1,65 @@
...
+```
+
+## Display the Results
+
+In this section we will show how to view the sample data and segmentation results returned from Vista-3D.
+
+1. Decompress the sample data volume:
+   ```
+   gzip -d build/nvidia_nim_imaging/applications/nvidia_nim/nvidia_nim_imaging/sample.nii.gz
+   ```
```
tbirdso, line 43: Can you confirm whether this step is necessary? `.nii.gz` is a standard medical image format, I would expect you can load `sample.nii.gz` into 3D Slicer and then 3D Slicer will handle the decompression.

applications/nvidia_nim/nvidia_nim_imaging/nvidia_nim.yaml
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+extensions:
+
+# For a self-hosted NIM instance, uncomment the following lines and replace the base_url
+# with the URL of your NIM instance.
+# nim:
+#   base_url: http://0.0.0.0:8000/v1/medicalimaging/nvidia/vista-3d
```
tbirdso, line 23: Would uncommenting as-is work or would the `base_url` on line 28 need to be replaced?

applications/nvidia_nim/nvidia_nim_imaging/metadata.json
```diff
@@ -0,0 +1,36 @@
...
+		"version": "1.0",
+		"changelog": {
+			"1.0": "Initial Release"
+		},
+		"holoscan_sdk": {
+			"minimum_required_version": "1.0.0",
+			"tested_versions": [
+				"1.0.0"
+			]
+		},
```
tbirdso, line 20: Probably `2.1.0` based on `requirements.txt`?

```suggestion
			"minimum_required_version": "2.1.0",
			"tested_versions": [
				"2.1.0"
			]
		},
```

applications/nvidia_nim/nvidia_nim_imaging/app.py
```diff
@@ -0,0 +1,216 @@
...
+    def __init__(self, model_params: Dict, user_input: str):
+        self.model_params = model_params
+        self.user_input = user_input
+
+    def to_json(self):
+        return json.dumps(self.__dict__)
+
+    @classmethod
+    def from_json(cls, json_text):
+        return MessageBody(**json.loads(json_text))
```
tbirdso, line 65: Unused class?

applications/nvidia_nim/nvidia_nim_imaging/app.py
```diff
@@ -0,0 +1,216 @@
...
+        self.add_flow(nim_op, unzip_op, {("file_name", "zip_file_bytes")})
+        self.add_flow(unzip_op, print_files_op, {("matching_files", "data")})
+
+    def set_nifti_file_path(self, file):
+        self._nifti_file_path = file
+        logger.info(f"NIFTI file: {self._nifti_file_path}")
+
+
+def _download_dataset(api_key, validate_file_checksum):
+    logger.info("Downloading sample image data from NVIDIA NGC...")
```
tbirdso, line 143: I'm not clear, could you please clarify what is the difference between this dataset downloaded at init and the files fetched by `NIMOperator` at app runtime?

tbirdso (commented):
Thanks @mocsharp ! One question about the datasets being retrieved, otherwise comments are mostly nitpicks. Seems reasonable to rely on 3D Slicer for visualization for now.

applications/nvidia_nim/nvidia_nim_imaging/README.md
mocsharp: You were right, this step is not required.

applications/nvidia_nim/nvidia_nim_imaging/nvidia_nim.yaml
mocsharp: We ship the configuration file to use the NVIDIA-hosted Vista 3D API by default. This line is for the users who want to use a self-hosted version of the API.

applications/nvidia_nim/nvidia_nim_imaging/metadata.json
mocsharp: I updated `requirements.txt` to use 1.0.3

applications/nvidia_nim/nvidia_nim_imaging/app.py
mocsharp: Correct.

applications/nvidia_nim/nvidia_nim_imaging/app.py
mocsharp: This downloads the input dataset.  The operator returns the result from the Vista3D API call.

applications/nvidia_nim/nvidia_nim_imaging/app.py
mocsharp: Removed! Thank you!

applications/nvidia_nim/nvidia_nim_imaging/nvidia_nim.yaml
tbirdso: Ah I understand now, thanks!

applications/nvidia_nim/nvidia_nim_imaging/app.py
tbirdso: Understood!

tbirdso (approved):
Great work @mocsharp ! Thanks for making those updates.

PR #365, opened 2024-05-30
Update benchmarking script and base Dockerfile to enable start_gui.py

wendell-hom:
Adding some infrastructure changes to make it easier to call `app_perf_graph.py` from another script and enable it to be easier to add `start_gui.py` script in subsequent checkins.  

The `start_gui.py` script will add a GUI interface so that users can 
- modify application operator parameter's while the application is running
- have a live latency graph that shows the elapsed time of each node

### DIFF
--- a/benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
+++ b/benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
@@ -16,6 +16,7 @@
 import argparse
 import os
 import queue
+import sys
 import time
 
 import pydot
@@ -140,11 +141,23 @@ def create_graph(
     edge_avg_latencies,
     edge_max_latencies,
     num_samples,
+    highlight,
 ):
+    # Go through all of the operator's average latencies and find the average
+    if len(operator_avg_latencies) > 0:
+        values = operator_avg_latencies.values()
+        total_sum = sum(values)
+        avg_val = total_sum / len(values)
+
     graph = pydot.Dot(graph_type="digraph")
     for op, latency in operator_avg_latencies.items():
+        color = "red" if highlight and (latency > avg_val) else "black"
+        penwidth = "2.0" if highlight and (latency > avg_val) else "1.0"
         node = pydot.Node(
-            op, label="{}\navg: {:.2f}\nmax: {:.2f}".format(op, latency, operator_max_latencies[op])
+            op,
+            color=color,
+            penwidth=penwidth,
+            label="{}\navg: {:.2f}\nmax: {:.2f}".format(op, latency, operator_max_latencies[op]),
         )
         graph.add_node(node)
     for edge, latency in edge_avg_latencies.items():
@@ -172,10 +185,16 @@ def add_graph_labels(graph, num_samples):
     return graph
 
 
-def main():
+def parse_arguments():
     parser = argparse.ArgumentParser(description="Show the application graph with latency data.")
     parser.add_argument(
         "filenames", nargs="+", help="The files to plot. In live mode, provide a folder."
+    )
+    parser.add_argument(
+        "-h",
+        "--highlight",
+        action="store_true",
+        help="highlight nodes in the graph whose avg. latency is higher than the average",
     )
     parser.add_argument(
         "-l", "--live", action="store_true", help="live mode: keep updating the graph with new data"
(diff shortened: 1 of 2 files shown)


benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
```diff
@@ -199,14 +218,56 @@ def main():
...
+    return args
+
+
+def print_message(message, silent):
+    if silent:
+        return
+    print(message)
+
+
+def create_dot_file(filenames, output_filename, live_graph, verbose, highlight, silent=False):
```
sohamm17, line 268: We can keep using `verbose` instead of introducing another variable `silent`.

sohamm17 (commented):
Thanks for the nice improvements! Left some comments.

dev_container
```diff
@@ -540,10 +540,12 @@ launch() {
     if [ -d  "$local_sdk_root" ] ; then
         local_sdk_opt+=" -v $local_sdk_root:/workspace/holoscan-sdk"
          #   Define paths needed by the python applications
-        local_sdk_opt+=" -e PYTHONPATH=/workspace/holoscan-sdk/build/python/lib"
+        local_sdk_opt+=" -e PYTHONPATH=/workspace/holoscan-sdk/build/python/lib:/workspace/holohub/benchmarks/holoscan_flow_benchmarking"
```
tbirdso, line 543: Would be nice to maintain the path in only one place, with something like:

```bash
local container_pythonpath="/workspace/holohub/benchmarks/holoscan_flow_benchmarking"
if [ ... ]; then
  container_pythonpath="/workspace/holoscan-sdk/build/python/lib:${container_pythonpath}"
else
  container_pythonpath="/opt/nvidia/holoscan/python/lib:${container_pythonpath}"
fi
local_sdk_opt+=" -e PYTHONPATH=${container_pythonpath}
```

benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
whom3: @sohamm17 are we ok with updating some of the messages to print out only if verbose is true?  I was trying to keep the existing behavior but it makes more sense to not introduce the extra silent if that's not a concern

Dockerfile
```diff
@@ -53,7 +53,10 @@ RUN apt update \
         libgstreamer1.0-0 \
         libgstreamer-plugins-base1.0-0 \
         libgles2 \
-        libopengl0
+        libopengl0 \
+        qtbase5-dev \
```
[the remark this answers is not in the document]
whom3: @tbirdso that's right, this was meant to make it easier for users to add GUI for other apps in Holohub w/o having to introduce/modify Dockerfile for their app

benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
sohamm17: @whom3 Yes, I think it is fine to skip printing a few things when `verbose` is not `true`.

Dockerfile
```diff
@@ -53,7 +53,10 @@ RUN apt update \
...
+        xdot \
```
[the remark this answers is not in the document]
whom3: ok, thanks.  Looks like this wasn't necessary since it's covered by the benchmarking packages you mentioned

dev_container
whom3: thanks @tbirdso , updated

Dockerfile
```diff
@@ -64,8 +65,17 @@ RUN apt update \
     libgtk-3-dev \
     libcanberra-gtk-module \
     graphviz
+
+# Install Qt6
+RUN apt update \
+    && apt install --no-install-recommends -y \
+        qt6-base-dev \
+        libxcb-cursor0 \
+    && rm -rf /var/lib/apt/lists/*
```
tbirdso, line 74: Do we also want to find an alternative for Qt C++? Or only for Qt Python?

whom3: We just need an alternative for PyQt.  Both PyQt and PySide are python bindings for C++ Qt which @AndreasHeumann has filed SWIPAT for in the past https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=4227741 

PyQt is not developed by Qt and so has additional licensing on top of it, so switching to PySide now which is from Qt.

tbirdso: Okay, so we will recommend that users use Qt for C++ and PySide for Python apps, and that will satisfy licensing due to both being available under LGPL? That makes sense to me, thanks for clarifying.

benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
```diff
@@ -140,11 +141,23 @@ def create_graph(
     edge_avg_latencies,
     edge_max_latencies,
     num_samples,
+    highlight,
 ):
+    # Go through all of the operator's average latencies and find the average
+    if len(operator_avg_latencies) > 0:
+        values = operator_avg_latencies.values()
+        total_sum = sum(values)
+        avg_val = total_sum / len(values)
```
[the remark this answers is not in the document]
wendell-hom: updated

benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
```diff
@@ -140,11 +141,23 @@ def create_graph(
     edge_avg_latencies,
     edge_max_latencies,
     num_samples,
+    highlight,
 ):
+    # Go through all of the operator's average latencies and find the average
+    if len(operator_avg_latencies) > 0:
```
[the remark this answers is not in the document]
wendell-hom: updated

benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
```diff
@@ -140,11 +141,23 @@ def create_graph(
...
+        values = operator_avg_latencies.values()
+        total_sum = sum(values)
+        avg_val = total_sum / len(values)
+
     graph = pydot.Dot(graph_type="digraph")
     for op, latency in operator_avg_latencies.items():
+        color = "red" if highlight and (latency > avg_val) else "black"
+        penwidth = "2.0" if highlight and (latency > avg_val) else "1.0"
```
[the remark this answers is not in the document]
wendell-hom: updated

benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
```diff
@@ -172,11 +185,17 @@ def add_graph_labels(graph, num_samples):
...
+def parse_arguments():
     parser = argparse.ArgumentParser(description="Show the application graph with latency data.")
     parser.add_argument(
         "filenames", nargs="+", help="The files to plot. In live mode, provide a folder."
     )
+    parser.add_argument(
+        "-h",
+        "--highlight",
+        action="store_true",
+        help="highlight nodes in the graph whose avg. latency is higher than the average",
```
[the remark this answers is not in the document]
wendell-hom: updated

benchmarks/holoscan_flow_benchmarking/app_perf_graph.py
wendell-hom: updated

wendell-hom:
@tbirdso after discussing with @jjomier, I'm modified this PR to remove changes to the base Dockerfile to avoid increasing the image size.  Instead I'll add instructions in the planned tutorial to mention adding a Dockerfile to their app if needed or to modify existing one to add the required dependencies

PR #370, opened 2024-06-04
Fixing tests on CAGX

jjomier:
- CAGX takes a little bit longer to generate TRT files -> Increase the timeout to 900 (it takes a little bit more than 12 minutes).
- The LD_LIBRARY_PATH environment needs to be set for h264 applications

### DIFF
--- a/applications/h264/h264_endoscopy_tool_tracking/CMakeLists.txt
+++ b/applications/h264/h264_endoscopy_tool_tracking/CMakeLists.txt
@@ -67,4 +67,10 @@ if(BUILD_TESTING)
   set_tests_properties(h264_endoscopy_tool_tracking_cpp_test PROPERTIES
                        PASS_REGULAR_EXPRESSION "Deactivating Graph"
                        FAIL_REGULAR_EXPRESSION "[^a-z]Error;ERROR;Failed")
+
+  # For aarch64 LD_LIBRARY_PATH needs to be set
+  if(CMAKE_SYSTEM_PROCESSOR STREQUAL aarch64 OR CMAKE_SYSTEM_PROCESSOR STREQUAL arm64)
+    set_tests_properties(h264_endoscopy_tool_tracking_cpp_test PROPERTIES ENVIRONMENT
+                    "LD_LIBRARY_PATH=/usr/lib/aarch64-linux-gnu/tegra/")
+  endif()
 endif()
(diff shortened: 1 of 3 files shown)


applications/h264/h264_endoscopy_tool_tracking/CMakeLists.txt
```diff
@@ -67,4 +67,10 @@ if(BUILD_TESTING)
   set_tests_properties(h264_endoscopy_tool_tracking_cpp_test PROPERTIES
                        PASS_REGULAR_EXPRESSION "Deactivating Graph"
                        FAIL_REGULAR_EXPRESSION "[^a-z]Error;ERROR;Failed")
+
+  # For aarch64 LD_LIBRARY_PATH needs to be set
+  if(CMAKE_SYSTEM_PROCESSOR STREQUAL aarch64 OR CMAKE_SYSTEM_PROCESSOR STREQUAL arm64)
+    set_tests_properties(h264_endoscopy_tool_tracking_cpp_test PROPERTIES ENVIRONMENT
+                    "LD_LIBRARY_PATH=/usr/lib/aarch64-linux-gnu/tegra/")
+  endif()
```
tbirdso, line 75: Is this for container testing only? Why is `LD_LIBRARY_PATH` not set correctly in the container environment?

Is `LD_LIBRARY_PATH` already empty such that no paths are populated by default, or is this replacing other environment paths?

Should we set this in `run` instead, as we do with `PYTHONPATH`?

applications/ultrasound_segmentation/cpp/CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ if(BUILD_TESTING)
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
   set_tests_properties(ultrasound_segmentation_cpp_test PROPERTIES
                        ENVIRONMENT "LD_LIBRARY_PATH=$ENV{LD_LIBRARY_PATH}:${CMAKE_BINARY_DIR}/gxf_extensions"
+                       TIMEOUT 900
```
tbirdso, line 98: Is this different from the 1-hour TRT generation issue previously reported?

jjomier: This is the only test that fails currently on CAGX so probably unrelated

applications/h264/h264_endoscopy_tool_tracking/CMakeLists.txt
jjomier: This is just for H264 on arm64 (documented on the readme). I'm not sure about the other path for LD_LIBRARY_PATH.

tbirdso (approved):

PR #372, opened 2024-06-06
Fixing CAGX tests

jjomier:
- Increasing ultrasound test timeout to 1500s to allow for generation of TRT engine file
- Fix environment variables for H264


jjomier:
> Changes look good. @jjomier have you confirmed whether these tests pass locally on CAGX with these updates?

I have not tested the h264 but if that fails we might need to do something else.

PR #373, opened 2024-06-06
Update volume renderer operator to use ClaraViz 0.4.0

AndreasHeumann:
This fixes a compile warning and makes OptiX and NvEnc optional to support ruinng on iGPU.


tbirdso (approved):
Looks good!

PR #375, opened 2024-06-06
Modify data path to be based on build path

NigelNelson:
The run script has been updated so that the data path environment variable is based on the build_path variable.

This makes sense to me as I am developing in a container where I don't have control of the volume mount, which is a perfect use case for the `--buildpath` arg. Intuitively, it would make sense to have that data dir be adjusted by this variable as well. That way, both your build and data will be persisted when developing in containers with different volume mount locations.

### DIFF
--- a/run
+++ b/run
@@ -658,6 +658,9 @@ build() {
     build_path="build/$1"
   fi
 
+  # Create data path outside build directory
+  data_path="${build_path}/../data"
+
   echo "Building $1 application"
   application="-DAPP_$1=1"
 
@@ -665,7 +668,7 @@ build() {
   cmake_extra_args="--no-warn-unused-cli -DPython3_EXECUTABLE=${HOLOHUB_PY_EXE} -DPython3_ROOT_DIR=${HOLOHUB_PY_LIB}"
 
   # We set the data directory to be outside the build directory
-  cmake_extra_args="$cmake_extra_args $configure_args -DHOLOHUB_DATA_DIR:PATH=${SCRIPT_DIR}/data"
+  cmake_extra_args="$cmake_extra_args $configure_args -DHOLOHUB_DATA_DIR:PATH=${data_path}"
 
   # Sets the default path for cuda
   export PATH=$PATH:/usr/local/cuda/bin


run
```diff
@@ -658,14 +658,17 @@ build() {
     build_path="build/$1"
   fi
 
+  # Create data path outside build directory
+  data_path="${build_path}/../data"
```
tbirdso, line 662: As currently written, it looks like this would change HoloHub behavior so that `data` is a subdirectory of `build` rather than at the top level. I.e., `data_path=/workspace/holohub/build/myapp/../data`. I'm not necessarily opposed to that change in behavior, just want to make sure it's known.

I do suggest we run `mkdir` to error out if the relative path does not exist or encounters permissions issues:
```suggestion
  data_path="${build_path}/../data"
  mkdir -p data_path
```

NigelNelson: Ahh, that is not necessarily what I wanted to do, I wanted to try to have this change be a no-op if a build_path isn't passed. So `data_path="${build_path}/../../data"` is actually the behavior I wanted. Do you see any issues with this?

tbirdso: One thing I'm not clear on is whether the data path is required to be outside of the build directory or not. Maybe @jjomier or @wendell-hom could weigh in on what they had in mind with "--buildpath" and build structure conventions?

I'm wondering if it would be less complicated to add a "--datapath" override specifier than to set our own convention for how it relates to a custom "--buildpath"?

jjomier: The data path can be anywhere but we do set it to a default when using the run script.
@NigelNelson what is the use case you are trying to solve here?

tbirdso: Following up here to move forward, to solve for the general case while avoiding too many implicit conventions, I suggest that we continue to use the default data path and add a `--datapath` CLI override argument to allow the user to specify a different argument for their specific needs.

tbirdso:
Hi @NigelNelson , is this PR still in progress or should we close it?

jjomier:
@NigelNelson Can you confirm if can close this MR and revisit later?

tbirdso:
Closing for now but open to revisiting the proposal.

PR #382, opened 2024-06-13
Streamline tao peoplenet, add support for replayer and testing

wendell-hom:


jjomier (approved):
Works well!

tbirdso (approved):
Code looks good to me!

PR #383, opened 2024-06-14
Fix cropping issue

wendell-hom:
Found a cropping issue during hackathon that affects tao_peoplenet and body_pose_estimation apps.  
Jonathan found the fix by specifying `in_dtype` for the preprocessor.

The image that was going into Inference Op was missing 25% of the right side of the picture, so the body and face detection is not working properly.  With this fix, the bounding boxes for the person on the right is working as expected.

**Before**
before

**After**
after

### DIFF
--- a/applications/body_pose_estimation/body_pose_estimation.py
+++ b/applications/body_pose_estimation/body_pose_estimation.py
@@ -351,6 +351,9 @@ class BodyPoseEstimationApp(Application):
     def compose(self):
         pool = UnboundedAllocator(self, name="pool")
 
+        # Input data type of preprocessor
+        in_dtype = "rgb888"
+
         # Determine if the DDS Publisher is enabled.
         dds_common_args = self.kwargs("dds_common")
         dds_publisher_args = dds_common_args | self.kwargs("dds_publisher")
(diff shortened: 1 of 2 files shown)


tbirdso (commented):
@wendell-hom could you comment on how the fix works? Not sure I understand the relation between the datatype and image width here?

Are all input sources already streaming or compatible with uint8 RGBA data? Do you anticipate any input/output compatibility issues with the new datatype?

wendell-hom:
> @wendell-hom could you comment on how the fix works? Not sure I understand the relation between the datatype and image width here?
The output of v4l2 operator is always `RGBA8888` as mentioned [here](https://gitlab-master.nvidia.com/holoscan/holoscan-sdk/-/blob/main/public/examples/v4l2_camera/README.md?ref_type=heads&plain=1#L7).  If we don't explicitly tell format converter that the data type is `RGBA8888` it seems to be treating it as something else possibly something like `RGB888` and as a result we are not capturing the entire image.

> Are all input sources already streaming or compatible with uint8 RGBA data? Do you anticipate any input/output compatibility issues with the new datatype?
That's a good point, I'll should probably only update the in_dtype to be `RGBA8888` if the input is v4l2.

tbirdso (approved):
Looks good, thanks!

PR #402, opened 2024-06-26
Consecutive similar lines are avoided for statistics in analyze

sohamm17:
If two consecutive lines are similar, then they are ignored. This can happen when there are two outputs from one operator, going to the same downstream operator. We will then log two lines, but their paths are same. Therefore, two latency numbers, same or similar to each other, will be represented in the latency list. This can have weird impact into statistical numbers such as max latency, 99.9, 99 percentile latencies. So, we ignore the second line in this case. Below two lines are considered same, so that same latency is not overrepresneted in the list of observed latencies.

Example:

```
(replayer,1718165307330350,1718165307332187) -> (format_converter,1718165307332241,1718165307332354) -> (lstm_inferer,1718165307332391,1718165307333036) -> (tool_tracking_postprocessor,1718165307333095,1718165307335577) -> (holoviz,1718165307335664,1718165307343867) 
(replayer,1718165307330350,1718165307332187) -> (format_converter,1718165307332241,1718165307332354) -> (lstm_inferer,1718165307332391,1718165307333036) -> (tool_tracking_postprocessor,1718165307333095,1718165307335550) -> (holoviz,1718165307335653,1718165307343867) 
```

### DIFF
--- a/benchmarks/holoscan_flow_benchmarking/log_parser.py
+++ b/benchmarks/holoscan_flow_benchmarking/log_parser.py
@@ -58,6 +58,7 @@ def is_same_path(line1, line2):
     # if none of the above is true, then they are same path
     return True
 
+
 # parse the log file and return all the latencies for each path
 # The format is (Operator1, receive timestamp, publish timestsamp) -> (Operator2, receive timestamp,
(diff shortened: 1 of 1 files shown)


jjomier (approved):
LGTM

PR #403, opened 2024-06-27
Fix h.264 endoscopy tool tracking python test

mocsharp:
The test for h.264 Endoscopy Tool Tracking Python app was not correctly configured. Fixing the CMakeLists.txt with corrected test commands.

### DIFF
--- a/applications/h264/h264_endoscopy_tool_tracking/python/CMakeLists.txt
+++ b/applications/h264/h264_endoscopy_tool_tracking/python/CMakeLists.txt
@@ -13,52 +13,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Enable the operators
+add_library(nvidia_nim_imaging INTERFACE)
+target_link_libraries(nvidia_nim_imaging INTERFACE holoscan::core holoscan::ops::holoviz holoscan::ops::lstm_tensor_rt_inference holoscan::ops::tensor_to_video_buffer holoscan::ops::tool_tracking_postprocessor)
+
 # Add testing
 if(BUILD_TESTING)
-  # To get the environment path
-  find_package(holoscan 2.1.0 REQUIRED CONFIG PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")
-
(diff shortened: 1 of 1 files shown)


run
```diff
@@ -493,7 +493,7 @@ lint() {
     pushd ${SCRIPT_DIR} > /dev/null
 
     echo "Linting Python"
-    run_command ruff $DIR_TO_RUN --ignore E712 || exit_code=1
+    run_command ruff check $DIR_TO_RUN --ignore E712 || exit_code=1
```
wendell-hom, line 496: @mocsharp were you seeing an error without `check`?

mocsharp: https://github.com/nvidia-holoscan/holohub/actions/runs/9703101172/job/26780463857#step:4:7

```
Linting Python
[command] ruff . --ignore E712
error: `ruff <path>` has been removed. Use `ruff check <path>` instead.
```

jjomier: Which version of ruff are you using?
I guess adding check does not hurt for previous versions...

applications/h264/h264_endoscopy_tool_tracking/python/CMakeLists.txt
```diff
@@ -13,52 +13,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
wendell-hom, line 15:
```suggestion
cmake_minimum_required(VERSION 3.20)

find_package(holoscan 2.1.0 REQUIRED CONFIG
             PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")

```

wendell-hom: Need to add the above, otherwise ctest will fail to find `holoscan` module

run
wendell-hom: Yea, looks fine to update this.

mocsharp: @jjomier it's coming from GitHub actions. Doesn't look like we pin to any specific version in the run script so likely the latest.

jjomier: That makes sense - thanks!

jjomier (approved):

PR #404, opened 2024-06-28
Support data export in out of body detection app

apatole:
This patch modifies out of body detection application to support exporting application output to the CSV files. This output can later be used by analytics applications.

This patch adds a new flag `enable_analytics` in config file which can be used to toggle exporting output to the CSV file. It is by default set to false.


jjomier (approved):
LGTM!

apatole:
Thanks @jjomier for the review!

Mentioning just for information, the lint error was unrelated so merged the change, thanks.

PR #414, opened 2024-07-03
Volume Rendering Python Sample

mocsharp:
This PR adds the Python version of the Volume Rendering application:
- Adds Python bindings for the `volume_renderer` and the `volume_loader` operators
- Updates the NRRD loader with support for loading attached-header format
- Updates README for both C++ and Python versions of the Volume Rendering application to use the new `build_and_run` command
- Updates the `run` script to print error and exit upon build failures

### DIFF
--- a/operators/volume_loader/CMakeLists.txt
+++ b/operators/volume_loader/CMakeLists.txt
@@ -59,6 +59,7 @@ target_link_libraries(volume_loader
   PRIVATE
     holoscan::core
     NIFTI::nifti2
+    pybind11::embed
   )
 
 target_include_directories(volume_loader INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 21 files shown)


operators/volume_loader/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ target_link_libraries(volume_loader
   PRIVATE
     holoscan::core
     NIFTI::nifti2
+    pybind11::embed
```
AndreasHeumann, line 62: I think this is no longer needed, I assume this was used for pre 2.1 not supporting Python type conversions for C++ operators.

AndreasHeumann (commented):
Looks good to me. Just one minor comment.

AndreasHeumann (approved):
Looks good, thanks!

PR #428, opened 2024-07-15
Update ANO README indentation

JohnMoon-VTS:
Currently, the ANO README renders strangely on GitHub due to an indentation issue in the source markdown. Fix this indentation.

For example, here's what it looks like currently:
image

And after this patch:
image


tbirdso (approved):
Thank you for fixing this @JohnMoon-VTS !

PR #432, opened 2024-07-17
Benchmark: Fix masked `-h` argument in flow tracking graph tool

tbirdso:
Fixes a small issue in `app_perf_graph.py` where the `-h` short option referred to either `--highlight` or the built-in `--help`, which prevented the tool from launching.

`app_perf_graph.py` is part of the `holoscan_flow_tracking` benchmarking tools/tutorial and is used for translating flow tracking log output into operator flow graphviz visualizations. The `-h` option is removed, but the `--highlight` long flag can still be used to launch the tool with highlighting.

Under previous behavior launching the script would return an error:
```
$ python3 app_perf_graph.py
...
argparse.ArgumentError: argument -h/--highlight: conflicting option string: -h
```


jjomier (approved):
Thanks for fixing this.

PR #435, opened 2024-07-18
percentile csv files are added by default

sohamm17:
Percentile csv files are added by default even though `--save-csv` is not true. The check was not added previously.


tbirdso (approved):
Code looks good. Please sign the commit before merge to resolve the CI failure (`git commit --amend --no-edit -s`)

tbirdso:
Thanks @sohamm17 !

PR #436, opened 2024-07-18
basic_network Python app - handle unavailable parameter values in HSDK 2.1

mocsharp:
Variables defined by `spec.param` are not ready in `initialize()` due to a change in HSDK 2.1.

This fix calls the original initialize logic when in `compute()` once if the connection is yet to be initialized.

Updates the README with instructions to run the receiver and the transmitter along with CMAKE to copy config files to the correct locations.

### DIFF
--- a/applications/basic_networking_ping/cpp/CMakeLists.txt
+++ b/applications/basic_networking_ping/cpp/CMakeLists.txt
@@ -31,11 +31,11 @@ target_link_libraries(basic_networking_ping
 
 # Copy config file
 add_custom_target(basic_networking_ping_rx_yaml
-  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_rx.yaml" ${CMAKE_BINARY_DIR}
+  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_rx.yaml" ${CMAKE_CURRENT_BINARY_DIR}
   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_rx.yaml"
 )
 add_custom_target(basic_networking_ping_tx_yaml
-  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_tx.yaml" ${CMAKE_BINARY_DIR}
+  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_tx.yaml" ${CMAKE_CURRENT_BINARY_DIR}
   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_tx.yaml"
 )
 
--- a/applications/basic_networking_ping/cpp/CMakeLists.txt
+++ b/applications/basic_networking_ping/cpp/CMakeLists.txt
@@ -31,11 +31,11 @@ target_link_libraries(basic_networking_ping
 
 # Copy config file
 add_custom_target(basic_networking_ping_rx_yaml
-  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_rx.yaml" ${CMAKE_CURRENT_BINARY_DIR}
+  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_rx.yaml" ${CMAKE_BINARY_DIR}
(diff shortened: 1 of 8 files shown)


applications/basic_networking_ping/cpp/CMakeLists.txt
```diff
@@ -31,11 +31,11 @@ target_link_libraries(basic_networking_ping
 
 # Copy config file
 add_custom_target(basic_networking_ping_rx_yaml
-  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_rx.yaml" ${CMAKE_CURRENT_BINARY_DIR}
+  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_rx.yaml" ${CMAKE_BINARY_DIR}
```
gigony, line 34: Is there any reason that `CMAKE_BINARY_DIR` makes sense here?

Many apps in Holohub is still using `CMAKE_CURRENT_BINARY_DIR`. Should we revisit other CMakeLists.txt files?

applications/basic_networking_ping/cpp/CMakeLists.txt
```diff
@@ -31,11 +31,11 @@ target_link_libraries(basic_networking_ping
...
   DEPENDS "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_rx.yaml"
 )
 add_custom_target(basic_networking_ping_tx_yaml
-  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_tx.yaml" ${CMAKE_CURRENT_BINARY_DIR}
+  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/../basic_networking_ping_tx.yaml" ${CMAKE_BINARY_DIR}
```
gigony, line 38: Same for this.

gigony (commented):
Thanks @mocsharp for this change!
Looks good to me. I have one question in this PR.

I think we can restore this change once the bug in 2.1 is fixed.

applications/basic_networking_ping/cpp/CMakeLists.txt
mocsharp: This change allows the user to launch the application using `./devcontainer build_and_run` and specify the config file as put in the README.

applications/basic_networking_ping/cpp/CMakeLists.txt
mocsharp: This change allows the user to launch the application using `./devcontainer build_and_run` and specify the config file as put in the README.

applications/basic_networking_ping/cpp/CMakeLists.txt
tbirdso: @mocsharp Could we please instead update the default `run` command in the app [`metadata.json`](https://github.com/nvidia-holoscan/holohub/blob/main/applications/basic_networking_ping/cpp/metadata.json#L42) to launch relative to `holohub_app_bin`, aka `CMAKE_CURRENT_BINARY_DIR`? I agreed with @gigony that `CMAKE_CURRENT_BINARY_DIR` would be better aligned with the Python app version along with other HoloHub apps.

mocsharp: Thank you @tbirdso for the tip. I have updated and tested the app.

applications/basic_networking_ping/cpp/CMakeLists.txt
mocsharp: Fixed as per the recommendation and updated `metadata.json` to use `holohub_app_bin` so it can pick up the config file correctly.

applications/basic_networking_ping/python/basic_networking_ping.py
```diff
@@ -65,8 +66,8 @@ class App(Application):
     def compose(self):
         # Define the tx and rx operators, allowing the tx operator to execute 10 times
         if len(self.kwargs("network_tx")) > 0:
-            tx = BasicNetworkPingTxOp(self, CountCondition(self, NUM_MSGS), name="tx")
             basic_net_tx = BasicNetworkOpTx(self, name="basic_net_tx", **self.kwargs("network_tx"))
+            tx = BasicNetworkPingTxOp(self, CountCondition(self, NUM_MSGS), name="tx")
```
tbirdso, line 70: Does order matter here? If so, why?

mocsharp: I moved it while I was debugging the issue, but apparently, the order doesn't matter. 
I kept the change so the both Rx and Tx define the operators in the same order.

tbirdso: Great, thanks for confirming 👍

tbirdso (approved):
Looks good to me, thanks @mocsharp !

PR #438, opened 2024-07-19
Remove arguments and use env var for data input

mocsharp:
This PR fixes an issue where in distributed mode, passing arguments such as `--worker` will be taken by the argument handler.

- Remove handling of arguments and use `HOLOSCAN_INPUT_PATH` as data input
- Add instructions to run the application in distributed mode


gigony (approved):
Thanks @mocsharp for the change!

It looks good to me.

Issue #440, opened 2024-07-23
vulkansdk Resource Not Found When Building and Running xr_hello_holoscan and volume_rendering_xr


codenML:
```
 > [vulkansdk-builder 2/4] RUN wget -nv --show-progress --progress=bar:force:noscroll     https://sdk.lunarg.com/sdk/download/1.3.216.0/linux/vulkansdk-linux-x86_64-1.3.216.0.tar.gz:                                                          
0.681 https://sdk.lunarg.com/sdk/download/1.3.216.0/linux/vulkansdk-linux-x86_64-1.3.216.0.tar.gz:                                                              
0.682 2024-07-23 14:23:56 ERROR 404: Not Found.
```

**Steps to Reproduce**

1. Set up per [README](https://github.com/nvidia-holoscan/holohub/tree/main/applications/volume_rendering_xr#prerequisites)
2. Run either of the following commands in freshly cloned repo:
`./dev_container build_and_run xr_hello_holoscan`
`./dev_container build_and_run volume_rendering_xr`

[build_and_run_log.txt](https://github.com/user-attachments/files/16350365/build_and_run_log.txt)

tbirdso:
Thanks for reporting @codenML . This issue shares a root cause with https://github.com/nvidia-holoscan/holoscan-sdk/issues/30 that was also identified today. I'll update the `volume_rendering_xr` Dockerfile with a similar resolution, and work with the Holoscan team to identify other potential failure points in HoloHub.

PR #441, opened 2024-07-23
App: Use Vulkan SDK packages in `volume_rendering_xr`

tbirdso:
Updates the `volume_rendering_xr` application Dockerfile to fetch Vulkan SDK libraries as Debian packages.

Addresses an issue where the Vulkan SDK installer v1.3.216.0 is no longer available for download as of 2024.07.23. Instead we use Debian packages for the closest version (1.3.204.1).

The container build previously fetched and built Vulkan SDK libraries from source on arm64. This was a reflection of previous availability where  Vulkan SDK arm64 Jammy Debian packages were previously not available from apt.

Resolves https://github.com/nvidia-holoscan/holohub/issues/440.

Tested on x86_64 with the Windrunner XR simulator, will validate on IGX as well before merge.

### DIFF
--- a/applications/volume_rendering_xr/Dockerfile
+++ b/applications/volume_rendering_xr/Dockerfile
@@ -28,36 +28,6 @@ COPY applications/volume_rendering_xr/thirdparty/magicleap/MagicLeapRemoteRender
 RUN /tmp/base-setup.sh
 
 ############################################################
-# Vulkan SDK
-#
-# Use the SDK because we need the newer Vulkan headers and the newer shader compiler than provided
-# by the Ubuntu deb packages. These are compile time dependencies, we still use the Vulkan loaded
-# and the Vulkan validation layer as runtime components provided by Ubuntu packages because that's
-# what the user will have on their installations.
-############################################################
-FROM base as vulkansdk-builder
-ARG VULKAN_SDK_VERSION
-
-WORKDIR /opt/vulkansdk
-ENV CMAKE_GENERATOR=Ninja
-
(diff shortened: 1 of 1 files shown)


applications/volume_rendering_xr/Dockerfile
```diff
@@ -27,36 +27,6 @@ COPY applications/volume_rendering_xr/thirdparty/magicleap/MagicLeapRemoteRender
     /usr/local/share/keyrings/magicleap/MagicLeapRemoteRendering.gpg
 RUN /tmp/base-setup.sh
 
-############################################################
```
AndreasHeumann, line 30: Could you remove the `ARG VULKAN_SDK_VERSION=1.3.216.0` above as well. That's no longer needed.

AndreasHeumann (approved):
Looks good,  just one minor issue.

applications/volume_rendering_xr/Dockerfile
```diff
@@ -77,23 +47,10 @@ RUN apt update \
         libxrandr-dev \
         libxxf86vm-dev \
         mesa-common-dev \
+        libvulkan-dev="1.3.204.1-*" \
+        glslang-tools="11.8.0+1.3.204.0-*" \
```
AndreasHeumann, line 51: It might be possible that `glslang-tools` is not needed as well. I don't think the example needs to compile GLSL shaders.

tbirdso: Good point, verified that we can remove this package 👍

applications/volume_rendering_xr/Dockerfile
tbirdso: Removed

PR #442, opened 2024-07-23
Fix Distributed Endoscopy Tool Tracking Test Failures

mocsharp:
This PR implements two options to provide input data and config file and fixes the test failures:

- Allows users to provide input data and config file via 
  - arguments: `--data` and `--config`, or
  - environment variables: `HOLOSCAN_INPUT_PATH` and `HOLOSCAN_CONFIG_PATH` (as prescribed by [HAP specifications](https://docs.nvidia.com/holoscan/sdk-user-guide/cli/hap.html))
  - fix test cases and add additional ones to cover both use cases
  - remove existing frame validation coverage since this app does not support recording

### DIFF
--- a/applications/endoscopy_tool_tracking_distributed/Dockerfile
+++ b/applications/endoscopy_tool_tracking_distributed/Dockerfile
@@ -1,0 +1,67 @@
+# syntax=docker/dockerfile:1
+
+# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+############################################################
+# Base image
+############################################################
+
+ARG BASE_IMAGE
+ARG GPU_TYPE
+
+FROM ${BASE_IMAGE} AS base
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Holohub Environment Variables
+ENV HOLOSCAN_INPUT_PATH=/workspace/holohub/data/endoscopy
+
+# --------------------------------------------------------------------------
+#
+# Holohub run setup
+#
+
+RUN mkdir -p /tmp/scripts
+COPY run /tmp/scripts/
+RUN mkdir -p /tmp/scripts/utilities
+COPY utilities/holohub_autocomplete /tmp/scripts/utilities/
+RUN chmod +x /tmp/scripts/run
+RUN /tmp/scripts/run setup
+
+# Enable autocomplete
+RUN echo ". /etc/bash_completion.d/holohub_autocomplete" >> /etc/bash.bashrc
+
+# - This variable is consumed by all dependencies below as an environment variable (CMake 3.22+)
+# - We use ARG to only set it at docker build time, so it does not affect cmake builds
+#   performed at docker run time in case users want to use a different BUILD_TYPE
+ARG CMAKE_BUILD_TYPE=Release
+
+
+# For benchmarking
+RUN apt update \
+    && apt install --no-install-recommends -y \
+    libcairo2-dev \
+    libgirepository1.0-dev \
+    gobject-introspection \
+    libgtk-3-dev \
+    libcanberra-gtk-module \
+    graphviz
+RUN if ! grep -q "VERSION_ID=\"22.04\"" /etc/os-release; then \
+        pip install setuptools; \
+    fi
+COPY benchmarks/holoscan_flow_benchmarking/requirements.txt /tmp/benchmarking_requirements.txt
+RUN pip install -r /tmp/benchmarking_requirements.txt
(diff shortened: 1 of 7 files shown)


applications/endoscopy_tool_tracking_distributed/cpp/CMakeLists.txt
```diff
@@ -58,43 +58,21 @@ add_dependencies(endoscopy_tool_tracking_distributed endoscopy_tool_tracking_dis
...
-      --validation_frames_dir ${VALIDATION_FRAMES_DIR}
-      WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-  )
-
-  set_tests_properties(endoscopy_tool_tracking_distributed_cpp_render_test PROPERTIES
-  DEPENDS endoscopy_tool_tracking_distributed_cpp_test
-  PASS_REGULAR_EXPRESSION "Valid video output!"
-  )
-
 endif()
```
gigony, line 78: nit: need a blank line at the end.

applications/endoscopy_tool_tracking_distributed/cpp/metadata.json
```diff
@@ -32,7 +32,7 @@
 			]
 		},
 		"run": {
-			"command": "HOLOSCAN_INPUT_PATH=<holohub_data_dir>/endoscopy <holohub_app_bin>/endoscopy_tool_tracking_distributed",
+			"command": "<holohub_app_bin>/endoscopy_tool_tracking_distributed",
```
gigony, line 35: Does the user need to set the environment variable (`HOLOSCAN_INPUT_PATH`) before executing this command on the host system (without a Docker container)? How will the user know about the requirement for the environment variable if it is not specified in this command?

gigony (approved):
Thanks @mocsharp !
Looks good to me with a question on the example usage.

applications/endoscopy_tool_tracking_distributed/cpp/metadata.json
jjomier: We should add/use the `--data` from the command line. This is what we have for most of the applications.
See https://github.com/nvidia-holoscan/holohub/blob/main/applications/endoscopy_tool_tracking/cpp/metadata.json#L35 for reference

mocsharp: Okay. Gigon confirms that arguments still work; here's an excerpt from the `application.hpp` source code:

>    * This constructor parses the command line for flags that are recognized by App Driver/Worker,
>   * and removes all recognized flags so users can use the remaining flags for their own purposes.

Let me update the app to use `--data`

jjomier: Awesome - thanks for looking into this!

gigony (approved):
It looks great to me!
Thank you @mocsharp for fixing the issue!

applications/endoscopy_tool_tracking_distributed/cpp/CMakeLists.txt
```diff
@@ -58,43 +58,29 @@ add_dependencies(endoscopy_tool_tracking_distributed endoscopy_tool_tracking_dis
...
            WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
-  set_tests_properties(endoscopy_tool_tracking_distributed_cpp_test PROPERTIES
+
+  set_tests_properties(endoscopy_tool_tracking_distributed_cpp_test_args PROPERTIES
                        PASS_REGULAR_EXPRESSION "Reach end of file or playback count reaches to the limit. Stop ticking."
                        FAIL_REGULAR_EXPRESSION "[^a-z]Error;ERROR;Failed")
 
-  # Add a test to check the validity of the frames
-  add_test(NAME endoscopy_tool_tracking_distributed_cpp_render_test
-      COMMAND python3 ${CMAKE_SOURCE_DIR}/utilities/video_validation.py
```
tbirdso, line 87: Same question, why remove frame tests?

tbirdso (commented):
This looks great! A couple of questions about test coverage.

applications/endoscopy_tool_tracking_distributed/python/CMakeLists.txt
```diff
@@ -18,48 +18,33 @@ if(BUILD_TESTING)
...
-  set_tests_properties(endoscopy_tool_tracking_distributed_python_test PROPERTIES
-                PASS_REGULAR_EXPRESSION "Reach end of file or playback count reaches to the limit. Stop ticking.;"
-                FAIL_REGULAR_EXPRESSION "[^a-z]Error;ERROR;Failed")
+  add_test(NAME endoscopy_tool_tracking_distributed_python_test_env_vars
+          COMMAND python3 ${CMAKE_CURRENT_SOURCE_DIR}/endoscopy_tool_tracking.py --data "${HOLOHUB_DATA_DIR}/endoscopy" --config "${CMAKE_CURRENT_BINARY_DIR}/endoscopy_tool_tracking_testing.yaml"
+          WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 
-  # Add a test to check the validity of the frames
-  add_test(NAME endoscopy_tool_tracking_distributed_python_render_test
-    COMMAND python3 ${CMAKE_SOURCE_DIR}/utilities/video_validation.py
```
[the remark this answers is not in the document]
mocsharp: Yes, this app does not support recording like the non-distributed version of the endoscopy tool tracking app.

applications/endoscopy_tool_tracking_distributed/Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+
+ARG BASE_IMAGE
+ARG GPU_TYPE
+
+FROM ${BASE_IMAGE} AS base
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Holohub Environment Variables
+ENV HOLOSCAN_INPUT_PATH=/workspace/holohub/data/endoscopy
```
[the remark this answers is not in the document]
mocsharp: I have updated the Python test with `HOLOSCAN_INPUT_PATH` and remove the Dockerfile. 

I kept it thinking to have a way to demo the use of HAP.

PR #447, opened 2024-07-28
add --benchmark option in build

sohamm17:
This MR adds a `--benchmark` option in `./run build` so that an application is built with `--configure-args -DCMAKE_CXX_FLAGS=-I$PWD/benchmarks/holoscan_flow_benchmarking`

### DIFF
--- a/run
+++ b/run
@@ -565,6 +565,7 @@ build_desc() {
   echo "   --buildpath <build_directory>        : Change the build path."
   echo "                                          Default: build"
   echo "                                          Associated environment variable: CMAKE_BUILD_PATH"
+  echo "   --benchmark                          : Build for Holoscan Flow Benchmarking."
   echo "   --configure-args <extra_args>        : Additional configuration arguments"
   echo "                                          multiple arguments can be passed between quotes"
   echo "                                          or using several --configure-args in the command line"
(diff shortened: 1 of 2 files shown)


run
```diff
@@ -565,6 +565,9 @@ build_desc() {
   echo "   --buildpath <build_directory>        : Change the build path."
   echo "                                          Default: build"
   echo "                                          Associated environment variable: CMAKE_BUILD_PATH"
+  echo "   --benchmark                          : Build for Holoscan Flow Benchmarking."
+  echo "                                          Make sure the application is patched before"
+  echo "                                          building with this option."
```
tbirdso, line 570: Could we automate this step to reduce dev responsibilities? Can we run `patch_application.sh` automatically prior to the CMake build when `--benchmark` is specified?

sohamm17: Thanks for the suggestion! Automated it.

run
```diff
@@ -693,6 +705,11 @@ build() {
   ret=$?
   check_exit_code $ret "Error building application."
 
+  if [ $benchmark == true ]; then
+    app_source_root_path=$(get_app_source_root_dir $1)
+    run_command benchmarks/holoscan_flow_benchmarking/restore_application.sh ${app_source_root_path}
```
tbirdso, line 710: It's not clear to me whether restoring the application after building is the right action or not. The source won't be an exact reflection of the built app, and I could see it being confusing for debugging. But then, interactive debugging doesn't necessarily make sense during benchmarking anyway.

I think it's fine to leave in for now and revisit if we encounter any issues.

tbirdso (approved):
Looks good to me!

Please rebsae on `main` and sign commits (`git commit -s`) to merge

PR #453, opened 2024-07-31
Update RTI Connext version to 7.3.0

ibstewart:
- Updates documentation and metadata to point to the latest RTI Connext 7.3.0 LTS release.
 - Adds CONNEXTDDS_ARCH environment variable to build container launch command.
 - Installs Java (and sets JREHOME environment variable) in the dev container, which is required by the RTI build time code generation tools.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,3 +69,9 @@ RUN if ! grep -q "VERSION_ID=\"22.04\"" /etc/os-release; then \
     fi
 COPY benchmarks/holoscan_flow_benchmarking/requirements.txt /tmp/benchmarking_requirements.txt
 RUN pip install -r /tmp/benchmarking_requirements.txt
+
+# For RTI Connext DDS
+RUN apt update \
+    && apt install --no-install-recommends -y \
+        openjdk-21-jre
+RUN echo 'export JREHOME=$(readlink /etc/alternatives/java | sed -e "s/\/bin\/java//")' >> /etc/bash.bashrc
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -69,3 +69,9 @@ RUN if ! grep -q "VERSION_ID=\"22.04\"" /etc/os-release; then \
     fi
 COPY benchmarks/holoscan_flow_benchmarking/requirements.txt /tmp/benchmarking_requirements.txt
 RUN pip install -r /tmp/benchmarking_requirements.txt
+
+# For RTI Connext DDS
+RUN apt update \
+    && apt install --no-install-recommends -y \
+        openjdk-21-jre
+RUN echo 'export JREHOME=$(readlink /etc/alternatives/java | sed -e "s/\/bin\/java//")' >> /etc/bash.bashrc
```
tbirdso, line 77: To confirm, is this valid for arm64 as well as x86?

tbirdso (commented):
Looks great! One platform question.

Dockerfile
ibstewart: The RTI host package for x86 does include its own prebuilt java runtime that can be used by the code generation tools, so it's not actually required to install another runtime into the container, but it also doesn't break anything and the install in the container is pretty minimal so I figured it would be easiest to just install it for both rather than have it be an arch-specific conditional install. I also wouldn't' object to make it arch-specific if that would be preferred though.

tbirdso: Great, this is good since it works on both systems. Thanks!

tbirdso (approved):
Looks good to me. Please sign the commit to merge: `git commit --amend --no-edit -s`

PR #455, opened 2024-08-01
Fixing readme for endoscopy_depth_estimation

jjomier:
Switch to run_and_build command and specifies that application is not supported on iGPU


tbirdso (approved):
Looks good to me

PR #456, opened 2024-08-01
Updating ultrasound_segmenation AVI file since index seemed broken

jjomier:


tbirdso (commented):
Looks like the dataset is also referenced in `openigtlink_3dslicer`, can we please update there as well? cc @brudfors

jjomier:
Thanks updated now. This does not change anything to the data BTW.

tbirdso (approved):
Thanks!

PR #466, opened 2024-08-12
Enable packaging of the applications using Holoscan CLI

mocsharp:
Add scripts to package the Endoscopy Tool Tracking application and the Object Detection Torch application.

The script copies necessary files from the build output and generates commands to use with the Holoscan CLI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,9 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/deps)
 include(HoloHubConfigHelpers)
 
 # Set install directory
-set(CMAKE_INSTALL_PREFIX "${CMAKE_SOURCE_DIR}/install")
+if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+  set_property(CACHE CMAKE_INSTALL_PREFIX PROPERTY VALUE "${CMAKE_SOURCE_DIR}/install")
+endif()
 
 # Enable Testing
 include(CTest)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,9 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/deps)
 # To configure which apps/operators to build
 include(HoloHubConfigHelpers)
 
+# Set install directory
+set(CMAKE_INSTALL_PREFIX "${CMAKE_SOURCE_DIR}/install")
+
 # Enable Testing
 include(CTest)
 
--- a/applications/endoscopy_tool_tracking/README.md
+++ b/applications/endoscopy_tool_tracking/README.md
@@ -109,3 +109,7 @@ There are a two launch profiles configured for this application:
 
 1. **(debugpy) endoscopy_tool_tracking/python**: This launch profile enables debugging of Python code.
 2. **(pythoncpp) endoscopy_tool_tracking/python**: This launch profile enables debugging of Python and C++ code.
+
+## Containerize the application
+
+To containerize the application, first build the application, run the `package-app.sh` script in the [cpp](./cpp/package-app.sh) or the [python](./python/package-app.sh) directory and then follow the generated output to package and run the application.
--- a/applications/endoscopy_tool_tracking/README.md
+++ b/applications/endoscopy_tool_tracking/README.md
@@ -112,4 +112,6 @@ There are a two launch profiles configured for this application:
 
 ## Containerize the application
 
-To containerize the application, first build the application, run the `package-app.sh` script in the [cpp](./cpp/package-app.sh) or the [python](./python/package-app.sh) directory and then follow the generated output to package and run the application.
+To containerize the application using [Holoscan CLI](https://docs.nvidia.com/holoscan/sdk-user-guide/cli/cli.html), first build the application, run the `package-app.sh` script in the [cpp](./cpp/package-app.sh) or the [python](./python/package-app.sh) directory and then follow the generated output to package and run the application.
+
+Refer to the [Packaging Holoscan Applications](https://docs.nvidia.com/holoscan/sdk-user-guide/holoscan_packager.html) section of the [Holoscan User Guide](https://docs.nvidia.com/holoscan/sdk-user-guide/) to learn more about installing the Holoscan CLI or packaging your application using Holoscan CLI.
--- a/applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
+++ b/applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
@@ -159,20 +159,25 @@ install(
   COMPONENT "holohub-apps"
 )
 
-file(READ "${endoscopy_tool_tracking_BINARY_DIR}/endoscopy_tool_tracking.yaml" APP_YAML_FILE)
+file(READ "${CMAKE_CURRENT_SOURCE_DIR}/endoscopy_tool_tracking.yaml" APP_YAML_FILE)
 string(REPLACE "gxf_extensions/lstm_tensor_rt_inference/" "" APP_YAML_FILE "${APP_YAML_FILE}")
-file(WRITE "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp/endoscopy_tool_tracking.yaml" "${APP_YAML_FILE}")
+file(WRITE "${CMAKE_BINARY_DIR}/packaging/cpp/endoscopy_tool_tracking.yaml" "${APP_YAML_FILE}")
+
+install(
+  FILES ${CMAKE_BINARY_DIR}/packaging/cpp/endoscopy_tool_tracking.yaml
+  DESTINATION endoscopy_tool_tracking_cpp
+  COMPONENT "holohub-apps"
+)
 
 install(
   DIRECTORY ${gxf_lstm_tensor_rt_inference_BINARY_DIR}/
-  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp"
+  DESTINATION endoscopy_tool_tracking_cpp
   FILES_MATCHING PATTERN "*.so"
   PATTERN "CMakeFiles" EXCLUDE)
 
 install(
   FILES
-    ${lstm_tensor_rt_inference_BINARY_DIR}/liblstm_tensor_rt_inference.so
-    ${tool_tracking_postprocessor_BINARY_DIR}/libtool_tracking_postprocessor.so
-  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp"
-  COMPONENT "holohub-apps"
-)
+  ${lstm_tensor_rt_inference_BINARY_DIR}/liblstm_tensor_rt_inference.so
+  ${tool_tracking_postprocessor_BINARY_DIR}/libtool_tracking_postprocessor.so
+  DESTINATION endoscopy_tool_tracking_cpp
+  COMPONENT "holohub-apps")
--- a/applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
+++ b/applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
@@ -151,3 +151,28 @@ if(FLOW_BENCHMARKING)
   # Set test timeout to 25 minutes
   set_tests_properties(endoscopy_tool_tracking_cpp_benchmark_test PROPERTIES TIMEOUT 1500)
 endif()
+
+# Install application and dependencies into the install/ directory for packaging
+install(
+  TARGETS endoscopy_tool_tracking
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp"
+  COMPONENT "holohub-apps"
+)
+
+file(READ "${endoscopy_tool_tracking_BINARY_DIR}/endoscopy_tool_tracking.yaml" APP_YAML_FILE)
+string(REPLACE "gxf_extensions/lstm_tensor_rt_inference/" "" APP_YAML_FILE "${APP_YAML_FILE}")
+file(WRITE "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp/endoscopy_tool_tracking.yaml" "${APP_YAML_FILE}")
+
+install(
+  DIRECTORY ${gxf_lstm_tensor_rt_inference_BINARY_DIR}/
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp"
+  FILES_MATCHING PATTERN "*.so"
+  PATTERN "CMakeFiles" EXCLUDE)
+
+install(
+  FILES
+    ${lstm_tensor_rt_inference_BINARY_DIR}/liblstm_tensor_rt_inference.so
+    ${tool_tracking_postprocessor_BINARY_DIR}/libtool_tracking_postprocessor.so
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp"
+  COMPONENT "holohub-apps"
+)
--- a/applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
+++ b/applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
@@ -14,6 +14,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+# Required fields for packaging the application
+# See https://docs.nvidia.com/holoscan/sdk-user-guide/cli/run_config.html#
 application:
   title: Holohub - Endoscopy Tool Tracking
   version: 1.0
--- a/applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
+++ b/applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
@@ -14,6 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+application:
+  title: Holohub - Endoscopy Tool Tracking
+  version: 1.0
+  inputFormats: []
+  outputFormats: ["screen"]
+
+resources:
+  cpu: 2
+  gpu: 1
+  memory: 1Gi
+  gpuMemory: 1Gi
+
 extensions:
   - gxf_extensions/lstm_tensor_rt_inference/libgxf_lstm_tensor_rt_inference.so
 # Uncomment the following extension when using deltacast as a source
--- a/applications/endoscopy_tool_tracking/cpp/package-app.sh
+++ b/applications/endoscopy_tool_tracking/cpp/package-app.sh
@@ -16,50 +16,15 @@
 set -e
 
 GIT_ROOT=$(readlink -f ./$(git rev-parse --show-cdup))
+APP_PATH="$GIT_ROOT/install/endoscopy_tool_tracking_cpp"
 
-# Utilities
+. $GIT_ROOT/utilities/bash_utils.sh
 
-YELLOW="\e[1;33m"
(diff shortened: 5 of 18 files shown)


applications/endoscopy_tool_tracking/README.md
```diff
@@ -109,3 +109,7 @@ There are a two launch profiles configured for this application:
 
 1. **(debugpy) endoscopy_tool_tracking/python**: This launch profile enables debugging of Python code.
 2. **(pythoncpp) endoscopy_tool_tracking/python**: This launch profile enables debugging of Python and C++ code.
+
+## Containerize the application
```
tbirdso, line 113: Neat! Could you please add another sentence or two explaining why someone might want to containerize their application? Build on one machine and distribute to another of the same architecture?

run
```diff
@@ -856,8 +856,8 @@ for k, v in obj[project_type]["run"].items():
       workdir="cd ${holohub_build_dir}"
    fi
 
-   local environment="export PYTHONPATH=\${PYTHONPATH}:${holoscan_sdk_install}/../../../python/lib:${holohub_build_dir}/python/lib:${SCRIPT_DIR} && export HOLOHUB_DATA_PATH=${holohub_data_dir}"
-   local reset_environment="export PYTHONPATH=${PYTHONPATH} && export HOLOHUB_DATA_PATH=\"${HOLOHUB_DATA_PATH}\""
+   local environment="export PYTHONPATH=\${PYTHONPATH}:${holoscan_sdk_install}/../../../python/lib:${holohub_build_dir}/python/lib:${SCRIPT_DIR} && export HOLOHUB_DATA_PATH=${holohub_data_dir} && export HOLOSCAN_INPUT_PATH=${holohub_data_dir}"
```
tbirdso, line 859: Is this required by the CLI packager?

applications/object_detection_torch/package-app.sh
```diff
@@ -0,0 +1,71 @@
...
+cp -f $GIT_ROOT/build/object_detection_torch/applications/object_detection_torch/object_detection_torch.yaml $APP_PATH
+
+PLATFORM=x64-workstation
+GPU=$(get_host_gpu)
+if [ $(get_host_arch) == "aarch64" ]; then
+    PLATFORM=igx-orin-devkit
+fi
+
+echo -e "done\n"
+echo -e Use the following commands to package and run the Object Detection Torch application:
```
tbirdso, line 65: Why print these commands rather than running them directly in the bash script?

applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
```diff
@@ -14,6 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
+application:
```
tbirdso, line 17: Could you please add a README reference clarifying that these fields are specified by the CLI app runner? https://docs.nvidia.com/holoscan/sdk-user-guide/cli/run_config.html#

applications/endoscopy_tool_tracking/cpp/package-app.sh
```diff
@@ -0,0 +1,77 @@
...
+
+APP_PATH="$GIT_ROOT/endoscopy_tool_tracking_cpp"
+echo Creating application directory $APP_PATH...
+mkdir -p $APP_PATH
+echo Copying application files to $APP_PATH...
+cp -f $GIT_ROOT/build/endoscopy_tool_tracking/applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking $APP_PATH
+cp -f $GIT_ROOT/build/endoscopy_tool_tracking/applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml $APP_PATH
+cp -f $GIT_ROOT/build/endoscopy_tool_tracking/gxf_extensions/lstm_tensor_rt_inference/*.so $APP_PATH
+cp -f $GIT_ROOT/build/endoscopy_tool_tracking/operators/lstm_tensor_rt_inference/liblstm_tensor_rt_inference.so $APP_PATH
+cp -f $GIT_ROOT/build/endoscopy_tool_tracking/operators/tool_tracking_postprocessor/libtool_tracking_postprocessor.so $APP_PATH
```
tbirdso, line 59: Since this starts to touch the build system, it would be good to standardize with CMake `install()` and `cmake --install` rather than directly copying built files. We can leverage CMake to 1) ensure files exist before we try to copy them, and 2) automatically build files if they do not exist before installing them.

Could you please add `install()` commands to the related application and operator targets in place of these lines? Similar to:
```
install(
  TARGETS endoscopy_tool_tracking
  COMPONENT holoscan-apps
)
```

Refer to [volume_rendering_xr](https://github.com/nvidia-holoscan/holohub/blob/main/applications/volume_rendering_xr/CMakeLists.txt#L86) for an example of installing files in HoloHub (without CLI).

cc @jjomier for a second opinion on criticality.

applications/endoscopy_tool_tracking/cpp/package-app.sh
```diff
@@ -0,0 +1,77 @@
...
+
+echo Updating application configuration...
+sed -e s!gxf_extensions/lstm_tensor_rt_inference/!!g -i $APP_PATH/endoscopy_tool_tracking.yaml
```
tbirdso, line 62: Like above, would be nice to move this to CMake as part of the build and installation process.

applications/endoscopy_tool_tracking/cpp/package-app.sh
```diff
@@ -0,0 +1,77 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -e
+
+GIT_ROOT=$(readlink -f ./$(git rev-parse --show-cdup))
+
+# Utilities
```
tbirdso, line 20: Could we reduce boilerplate code among `package-app.sh` by moving some of these functions into a script in `holohub/utilities`?

applications/endoscopy_tool_tracking/cpp/package-app.sh
```diff
@@ -0,0 +1,77 @@
...
+sed -e s!gxf_extensions/lstm_tensor_rt_inference/!!g -i $APP_PATH/endoscopy_tool_tracking.yaml
+
+PLATFORM=x64-workstation
+GPU=$(get_host_gpu)
+if [ $(get_host_arch) == "aarch64" ]; then
+    PLATFORM=igx-orin-devkit
+fi
+
+echo -e "done\n"
+echo -e Use the following commands to package and run the Endoscopy Tool Tracking application:
```
tbirdso, line 71: Could we consolidate `package-app.sh` scripts into a general command in the `run` script? Or are there too many specifics to the given apps?

applications/endoscopy_tool_tracking/python/endoscopy_tool_tracking.py
```diff
@@ -303,7 +303,7 @@ def compose(self):
     parser.add_argument(
         "-d",
         "--data",
-        default="none",
+        default=os.environ.get("HOLOSCAN_INPUT_PATH", None),
```
tbirdso, line 306: Seems like this should be the data path?

```suggestion
        default=os.environ.get("HOLOSCAN_DATA_PATH", None),
```

run
mocsharp: The `HOLOSCAN_INPUT_PATH` environment variable is required by [HAP](https://docs.nvidia.com/holoscan/sdk-user-guide/cli/hap.html) and the CLI.

applications/object_detection_torch/package-app.sh
mocsharp: When users install the PyPI version of the HSDK in a virtual environment, the CLI may not be accessible in a Holohub repo. Therefore, instead of running it automatically, the commands are printed.

tbirdso: I'm not familiar, what additional step would the user need to have access to the CLI? Could we include those steps in the printout? Under current behavior it's not clear that those commands may not be immediately valid.

applications/endoscopy_tool_tracking/python/endoscopy_tool_tracking.py
mocsharp: It is `HOLOSCAN_INPUT_PATH` per [HAP specification](https://docs.nvidia.com/holoscan/sdk-user-guide/cli/hap.html#table-of-environment-variables)

run
tbirdso: Got it, thanks for the link.

applications/endoscopy_tool_tracking/cpp/package-app.sh
mocsharp: Not at the moment, since each app has different dependencies that need to be copied into the container.
Plus, I'm not sure if plan to containerize all applications.

tbirdso: Mentioned above, I suggest that per-app dependencies should be managed in CMake and the script should be responsible for copying some `install-myapp` folder into the container.

Common infrastructure does not need to target all applications out of the box.

mocsharp: I've updated CMake to "install" the files required for packaging. Please review.

applications/endoscopy_tool_tracking/README.md
mocsharp: I added a few links to reference CLI sections in the user guide instead. Please let me know if that's fine.

applications/object_detection_torch/package-app.sh
mocsharp: I have updated the message with a link to get more info.

CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/deps)
 # To configure which apps/operators to build
 include(HoloHubConfigHelpers)
 
+# Set install directory
+set(CMAKE_INSTALL_PREFIX "${CMAKE_SOURCE_DIR}/install")
```
tbirdso, line 28: Let's use `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` to allow the user to override our default setting:
```suggestion
# Set install directory
if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
  set_property(CACHE CMAKE_INSTALL_PREFIX PROPERTY VALUE "${CMAKE_SOURCE_DIR}/install")
endif()
```
Ref: https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html#variable:CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT

applications/endoscopy_tool_tracking/README.md
tbirdso: Looks good, thanks!

applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
```diff
@@ -151,3 +151,28 @@ if(FLOW_BENCHMARKING)
...
+  COMPONENT "holohub-apps"
+)
+
+file(READ "${endoscopy_tool_tracking_BINARY_DIR}/endoscopy_tool_tracking.yaml" APP_YAML_FILE)
+string(REPLACE "gxf_extensions/lstm_tensor_rt_inference/" "" APP_YAML_FILE "${APP_YAML_FILE}")
+file(WRITE "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp/endoscopy_tool_tracking.yaml" "${APP_YAML_FILE}")
+
+install(
+  DIRECTORY ${gxf_lstm_tensor_rt_inference_BINARY_DIR}/
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp"
```
tbirdso, line 168: Destination is relative to `CMAKE_INSTALL_PREFIX`:

```suggestion
  DESTINATION "endoscopy_tool_tracking_cpp"
```

https://cmake.org/cmake/help/v3.24/command/install.html

applications/endoscopy_tool_tracking/python/CMakeLists.txt
```diff
@@ -63,3 +63,32 @@ if(BUILD_TESTING)
   )
 
 endif()
+
+# Install application and dependencies into the install/ directory for packaging
+install(
+  FILES ${CMAKE_CURRENT_SOURCE_DIR}/endoscopy_tool_tracking.py
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_python"
```
tbirdso, line 70:
```suggestion
  FILES endoscopy_tool_tracking.py
  DESTINATION "endoscopy_tool_tracking_python"
```

applications/endoscopy_tool_tracking/python/CMakeLists.txt
```diff
@@ -63,3 +63,32 @@ if(BUILD_TESTING)
...
+  COMPONENT "holohub-apps"
+)
+
+file(READ "${CMAKE_CURRENT_SOURCE_DIR}/endoscopy_tool_tracking.yaml" APP_YAML_FILE)
+string(REPLACE "gxf_extensions/lstm_tensor_rt_inference/" "" APP_YAML_FILE "${APP_YAML_FILE}")
+file(WRITE "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_python/endoscopy_tool_tracking.yaml" "${APP_YAML_FILE}")
```
tbirdso, line 76: Best to do this in two stages:
1. Write files generated at config time to the build directory, to avoid polluting sources in git
2. Reference the installation folder only via `install` in order to respect a user `--prefix` argument to `cmake install`

```suggestion
file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/packaging/endoscopy_tool_tracking.yaml" "${APP_YAML_FILE}")
install(
  FILE endoscopy_tool_tracking.yaml
  DESTINATION endoscopy_tool_tracking_python
  COMPONENT "holohub-apps"
)
```

applications/endoscopy_tool_tracking/python/CMakeLists.txt
```diff
@@ -63,3 +63,32 @@ if(BUILD_TESTING)
...
+
+install(
+  DIRECTORY ${gxf_lstm_tensor_rt_inference_BINARY_DIR}/
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_python"
```
tbirdso, line 80:
```suggestion
  DESTINATION "endoscopy_tool_tracking_python"
```

applications/endoscopy_tool_tracking/python/CMakeLists.txt
```diff
@@ -63,3 +63,32 @@ if(BUILD_TESTING)
...
+  FILES_MATCHING PATTERN "*.so"
+  PATTERN "CMakeFiles" EXCLUDE)
+
+install(
+  FILES
+    ${lstm_tensor_rt_inference_BINARY_DIR}/liblstm_tensor_rt_inference.so
+    ${tool_tracking_postprocessor_BINARY_DIR}/libtool_tracking_postprocessor.so
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_python"
```
tbirdso, line 88:
```suggestion
  DESTINATION "endoscopy_tool_tracking_python"
```

applications/endoscopy_tool_tracking/python/CMakeLists.txt
```diff
@@ -63,3 +63,32 @@ if(BUILD_TESTING)
...
+
+file(READ "${CMAKE_CURRENT_SOURCE_DIR}/endoscopy_tool_tracking.yaml" APP_YAML_FILE)
+string(REPLACE "gxf_extensions/lstm_tensor_rt_inference/" "" APP_YAML_FILE "${APP_YAML_FILE}")
+file(WRITE "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_python/endoscopy_tool_tracking.yaml" "${APP_YAML_FILE}")
+
+install(
+  DIRECTORY ${gxf_lstm_tensor_rt_inference_BINARY_DIR}/
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_python"
+  FILES_MATCHING PATTERN "*.so"
+  PATTERN "CMakeFiles" EXCLUDE)
```
tbirdso, line 82:
```suggestion
  PATTERN "CMakeFiles" EXCLUDE
  COMPONENT "holohub-apps"  
)
```

run
```diff
@@ -678,6 +678,9 @@ build() {
   run_command cmake --build ${build_path} -j
   ret=$?
   check_exit_code $ret "Error building application."
+  run_command cmake --install ${build_path}
+  ret=$?
+  check_exit_code $ret "Error installing application."
```
tbirdso, line 683: Could you please add an `--install` CLI argument to opt in to this block? Status quo is building without packaging, having installation as the default will likely lead to redundant library copies

applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
```diff
@@ -151,3 +151,28 @@ if(FLOW_BENCHMARKING)
...
+  FILES_MATCHING PATTERN "*.so"
+  PATTERN "CMakeFiles" EXCLUDE)
+
+install(
+  FILES
+    ${lstm_tensor_rt_inference_BINARY_DIR}/liblstm_tensor_rt_inference.so
+    ${tool_tracking_postprocessor_BINARY_DIR}/libtool_tracking_postprocessor.so
+  DESTINATION "${CMAKE_SOURCE_DIR}/install/endoscopy_tool_tracking_cpp"
```
tbirdso, line 176: Can we install CMake targets and let CMake determine the `.so` files to copy, rather than copying `.so` files ourselves? Something like:

```suggestion
install(
  TARGETS  lstm_tensor_rt_inference
  DESTINATION "endoscopy_tool_tracking_cpp"
```

tbirdso (commented):
Thanks @mocsharp for updating to use CMake installation. Reorienting towards the build system rather than manual library copies is looking better. Left a few CMake comments.

applications/object_detection_torch/package-app.sh
mocsharp: That's doable, and we can always attempt to automate calling and fallback to printing the commands. I created this PR to reduce the number of steps for QA to follow in the test procedures, and I did not want to automate every step as
1) I believe printing the commands and arguments gives the users an excellent example of using the CLI because there are just too many options
2) in case users install the PyPI package in their virtual environment, then automating this won't work

If having automated solutions is necessary, I can create a separate issue and work on it later.

CMakeLists.txt
mocsharp: Cool! Thanks!

applications/object_detection_torch/package-app.sh
tbirdso: Okay, understood, using this as a minimum configuration for SQA while still exposing users to available options makes sense. Would be nice to explore automating and consolidating scripts, but OK if it's a low priority. Thanks for clarifying, good to leave as is for this effort.

run
mocsharp: Thanks. I have added a `--install` argument to the `build` command.

applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
tbirdso: Discussed offline, this would need a closer look, the `install(lstm_tensor_rt_inference)` call needs to be after the target declaration but the application is declared first, and we don't want to install `lstm_tensor_rt_inference` to a specific app folder in the general case. Good to move forward as-is for the scope of this work. Thanks Victor for testing it out.

tbirdso (approved):
Thanks for addressing review comments @mocsharp , looks good to me!

PR #471, opened 2024-08-19
Utility: Pin linting dependencies and fix linting errors

tbirdso:
Utility: Pin linting tools versions
    
    Updates `run install_lint_deps` to retrieve pinned versions of linting
    packages.
    
    Linting tools can use different default rules across versions, leading
    to apparent new linting pipeline failures in files where no change
    occurred. This appears to have been the cause of the linting pipeline
    failure in 9755f256, which occurred concurrent with an update from
    `ruff==0.5.7` to `ruff==0.6.0` as the latest version. Pinning versions
    guards against similar unexpected pipeline failures in the future that
    can slow or block developer contributions.
    
    Changes:
    - Adds pinned versions for each PyPI and apt package installed with
      `run install_lint_deps`
    - Eagerly installs versions from PyPI and apt instead of checking for
      local tool availability first. Local tool versions may or may not
      match the expected version for linting, and could cause confusion if
      there is a version mismatch between local and CI installed versions.

Tutorial: Lint Self Supervised Training Notebook
    
    Apply linting fixes to `Contrastive_learning_Notebook.ipynb`
    following `ruff` and `black` linting tool updates.
    
    Reproduce with:
    ```
    sudo ./run install_lint_deps
    ./run lint
    ./run lint --fix
    ```
    
    Note that `__len__` is removed as a duplicate definition.

Pending local validation of the Self-Supervised Training Notebook with linting fixes applied.


tbirdso:
Looks like I don't currently have access to the `Cholec80` dataset for validation. @wendell-hom maybe you could help?

Other options:
- Move forward without explicit validation. I'm relatively confident the changes are non-breaking, as most of the changes to the notebook are from auto linting fixes and one manual change is removing a duplicate method.
- Ignore the notebook or revert to previous `ruff` version to skip the check.

wendell-hom (approved):

wendell-hom:
> Looks like I don't currently have access to the `Cholec80` dataset for validation. @wendell-hom maybe you could help?
> 
> Other options:
> 
> * Move forward without explicit validation. I'm relatively confident the changes are non-breaking, as most of the changes to the notebook are from auto linting fixes and one manual change is removing a duplicate method.
> * Ignore the notebook or revert to previous `ruff` version to skip the check.

@tbirdso I think we can move forward with this since the lint changes here are straightforward

PR #473, opened 2024-08-20
App Install - Prevent null variables during a clean build

mocsharp:
In a clean build, some variables may not be available during the cmake `install` stage. 

The fix is to use `CMAKE_BINARY_DIR` + relative path to the dependencies.


tbirdso (approved):
Thanks @mocsharp , the fix looks good. Ideally in the future we can replace with target installs, but as discussed that may take some more consideration for general HoloHub impact and is beyond the scope of what's needed here.

PR #475, opened 2024-08-22
Re-enable default holohub data path for Endoscopy Tool Tracking

mocsharp:
Fix an issue introduced in #466 to use the Holohub data path (`data/endoscopy`) if exists.


tbirdso (approved):
Thanks @mocsharp . Will wait for confirmation from @ibstewart before merge

tbirdso:
Confirmed 👍

PR #478, opened 2024-08-26
Support no-cache option in dev_container script

mocsharp:
While following the instructions in `operators/advanced_network/README.md`, I noticed that the `dev_container` script doesn't support the specified `no-cache` option.


tbirdso (approved):
LGTM. In the future we could add a `--docker_opts` flag to add arbitrary docker options to the build command, similar to what we do in `dev_container launch`.

PR #484, opened 2024-08-28
dev_container: build_and_install command

mocsharp:
- Adds a new `build_and_install` command to build and install the application for supported apps.
- Updates endoscopy_tool_tracking and object_detection_torch instructions to use the new command.

### DIFF
--- a/dev_container
+++ b/dev_container
@@ -836,121 +836,7 @@ build_and_install_desc() { c_echo 'Build the application and install the binarie
 }
 
 build_and_install() {
-    local app_name=""
-    local app_language=""
-    local container_build_args=""
-    local container_launch_args=""
-    local docker_opts="--entrypoint=bash"
-    local image_name=""
-    local docker_file=""
-    local extra_build_args=""
-    local extra_build_with=""
-
-    # Parse CLI arguments next
-    while [[ $# -gt 0 ]]; do
-        case "$1" in
-        --base_img)
-            if [[ -n "$2" && ! "$2" =~ ^-- ]]; then
-                container_build_args+=" --base_img $2"
-                shift 2
-            else
-                echo "Error: --base_img requires a value"
-                build_and_run_desc
-                exit 1
-            fi
-            ;;
-        --docker_file)
-            if [[ -n "$2" && ! "$2" =~ ^-- ]]; then
-                docker_file="$2"
-                shift 2
-            else
-                echo "Error: --docker_file requires a value"
-                build_and_run_desc
-                exit 1
-            fi
-            ;;
-        --container_args)
-            if [[ -n "$2" ]]; then
-                docker_opts+=" $2"
-                shift 2
-            else
-                echo "Error: --container_args requires a value"
-                build_and_run_desc
-                exit 1
-            fi
-            ;;
-        --build_args)
-            if [[ -n "$2" ]]; then
-                extra_build_args="--configure-args '$2'"
-                shift 2
-            else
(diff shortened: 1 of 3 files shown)


mocsharp:
@tbirdso, initially, I made the `./run build` command to install, but it was later modified to install only when the `--install` option is used. That means the QA test plan steps are broken, and therefore, I am adding a new command instead of another `--install` switch to the existing dev_container's `build_and_run` command.

dev_container
```diff
@@ -816,6 +816,143 @@ launch() {
         "$@"
 }
 
+build_and_install_desc() { c_echo 'Build the application and install the binaries into install/<app> if supported
```
tbirdso, line 819: It looks like this overlaps heavily with `build_and_run`. To reduce the maintenance burden and avoid redundancy, could we please instead add a `run_build_args` option or similar to `build_and_run` to enable passing the `--install` flag? Possibly along with a `--no_run` flag if you'd like to avoid the requirement to launch the app.

tbirdso (changes requested):

dev_container
tbirdso: In the meantime, can we please revert app instructions and QA isntructions to rely on `dev_container build` and `run build` directly for now?

mocsharp: Let me find a way to consolidate the two as I think UX is also important here:
`./dev_container build_and_install my_app`
vs
`./dev_container build_and_run my_app --install --no_run` 

:rofl:  I guess I'm too lazy to type those few extra characters :rofl:

tbirdso: My concern is that we do not want to start introducing `<action1>_and_<action2>` commands every time we want to chain together multiple behaviors. The `build_and_run` scope was originally intended as a convenience to cover the majority of app build cases, with `dev_container` and `run` as the appropriate fallback for more advanced app cases, such as the case of installation.

I'm not opposed to streamlining additional cases, but we should do it in a way that doesn't require introducing new CLI entrypoints to `dev_container` with each action.

Given that the motivation for this change is time sensitive, my request is that we fall back to `dev_container` and `run` for now, and continue discussion on advanced UX once QA can move forward.

mocsharp: Ok, I have updated `build_and_install` to reuse `build_and_run` with a couple of new options. Please have a look.

tbirdso: Awesome, having `build_and_install` as a streamlined interface to `build_and_run` looks great and is OK for moving forward. Thanks for refactoring.

dev_container
```diff
@@ -816,6 +816,29 @@ launch() {
...
+            - Application-provided "Dockerfile", if it exists;
+            - Otherwise the top-level HoloHub "Dockerfile"
+            If `--img` is not specified then a custom image tag will be defined.
+        --language : Specify the app language implementation to run.
+                     Some applications provide both `cpp` and `python` implementations.
+        --container_args: Additional Docker run args.
+        --build_args : Build the app with additional CMake configuration arguments
+        --build_with : List of optional operators that should be built separated by semicolons (;)
+        --verbose : Print extra output to console
+        --dryrun : View build and run commands without doing anything
```
tbirdso, line 834: One correction for the command name. Also, can we reference `build_and_run_desc` instead of listing all arguments here?

```suggestion
    Usage: ./dev_container build_and_install <application_name>
    View options with `./dev_container build_and_run_desc`
```

tbirdso (approved):
Updates look good to me, thanks for adding this!

PR #485, opened 2024-08-29
Update script messages to use build_and_install

mocsharp:
Update the error message in the `package-app.sh` scripts to use `build_and_install` instead of `build_and_run`.


MMelQin (approved):
LGTM

tbirdso (approved):

PR #487, opened 2024-08-29
VSCode - Compound Task for DETT - Set address and worker-address with port

mocsharp:
When launching the Distributed Endoscopy Tool Tracking compound task on IGX, worker fragments, including the worker thread launched along with the driver, may not always be able to connect to the driver. 

This fix explicitly configures the worker address ports and reconfigures build tasks to build first and then delay start.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -437,11 +437,15 @@
                 "--fragments",
                 "video_in",
                 "--address",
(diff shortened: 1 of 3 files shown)


gigony (approved):
Thanks @mocsharp for updating the vscode launch configuration!
It looks good to me!

tbirdso (approved):

PR #489, opened 2024-09-02
Vila replayer

maximilianofir:
- allow replaying videos in ViLa_live application. 
- allows for testing ViLa on a variety of videos from different domains

### DIFF
--- a/applications/vila_live/README.md
+++ b/applications/vila_live/README.md
@@ -61,6 +61,19 @@ Once the main LMM-based app is running, you will see a link for the app at `http
   sudo ln -s libv4l2.so.0.0.0.0  libv4l2.so.0.0.999999
   ```
 
+## 📷⚙️ Video Options
+There are three options to ingest video data.
+
+1. use a physical device or capture card, such as a v4l2 device as described in the [Setup Instructions](##-⚙️-Setup-Instructions
+). Make sure the [vila_live.yaml](./vila_live.yaml) contains the v4l2_source group and specifies the device correctly. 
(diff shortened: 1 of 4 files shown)


tbirdso (commented):
Code changes look good to me. Could we please add the following:
- Document the CLI in the app README;
- Add a v4l2 section in the app YAML with default settings for the user to update, similar to https://github.com/nvidia-holoscan/holohub/blob/3067096ad0ab2bd42a07c03b7a97f1d2d5bad8ee/applications/body_pose_estimation/body_pose_estimation.yaml#L20

NigelNelson:
Are these changes enabling the app to use GXF entities from the video replayer?
If this is the case, I find it easier to mount an MP4 using FFMPEG and V4L2 loopback than go through the GXF entity creation scripts.

Regardless, if replayer support is desired I think the README.md should link to where to find information on how to use the replayer.

maximilianofir:
Hi Nigel,
That makes more sense. I was not aware of the V4L2 loopback option until yesterday.
Will adjust the MR in favor of this workflow.

jjomier:
@NigelNelson @maximilianofir what is the status of this MR?

tbirdso (approved):
Code LGTM. Would like to wait for @NigelNelson 's review as well before merge

NigelNelson:
Looks good to me! Only a small nit: it would be nice to rename the new "setup instructions" to a "video options" or something similar, as otherwise there are two "setup instruction" settings.

NigelNelson (approved):

maximilianofir:
Thanks all, I renamed the section to Video Options.

PR #491, opened 2024-09-03
Parse --config for CLI run support

mocsharp:
The Holoscan CLI `run` command passes `--config` to applications by default. 

Since the Object Detection Torch app is not a distributed app and does not support `--config` out of the box, in order to avoid the error at launch, this PR updates the app to parse `--config` .

### DIFF
--- a/applications/object_detection_torch/main.cpp
+++ b/applications/object_detection_torch/main.cpp
@@ -23,16 +23,20 @@
 #include <holoscan/operators/holoviz/holoviz.hpp>
 #include <holoscan/operators/inference/inference.hpp>
 #include <holoscan/operators/inference_processor/inference_processor.hpp>
-#include <holoscan/operators/video_stream_recorder/video_stream_recorder.hpp>
 #include <holoscan/operators/video_stream_replayer/video_stream_replayer.hpp>
+#include <holoscan/operators/video_stream_recorder/video_stream_recorder.hpp>
 
 class App : public holoscan::Application {
  public:
-  void set_source(const std::string& source) {
-    if (source == "aja") { is_aja_source_ = true; }
+  void set_source(const std::string &source) {
+    if (source == "aja") {
+      is_aja_source_ = true;
+    }
   }
 
-  void set_datapath(const std::string& path) { datapath = path; }
+  void set_datapath(const std::string &path) {
+    datapath = path;
+  }
 
   // Specifies if the output of the visualizer should be recorded.
   enum class Record { NONE, VISUALIZER };
@@ -56,8 +60,8 @@ class App : public holoscan::Application {
     if (is_aja_source_) {
       source = make_operator<ops::AJASourceOp>("aja", from_config("aja"));
     } else {
-      source = make_operator<ops::VideoStreamReplayerOp>(
-          "replayer", from_config("replayer"), Arg("directory", datapath));
+      source = make_operator<ops::VideoStreamReplayerOp>("replayer", from_config("replayer"),
+                                                         Arg("directory", datapath));
     }
 
     auto in_dtype = is_aja_source_ ? std::string("rgba8888") : std::string("rgb888");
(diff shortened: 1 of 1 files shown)


applications/object_detection_torch/main.cpp
```diff
@@ -239,6 +240,9 @@ bool parse_arguments(int argc, char **argv, std::string &config_name, std::strin
       break;
 
     switch (c) {
+    case 'c':
+      config_name = optarg;
```
tbirdso, line 244: Note that under current behavior a config passed with `--config` will be overwritten if another config is also passed as a positional argument. Can you confirm whether that is intended? OK to move forward?

tbirdso (commented):
Is the `--config` parsing requirement documented in the [HAP specification](https://docs.nvidia.com/holoscan/sdk-user-guide/cli/hap.html#)? I see environment variables but not CLI requirements. Am I overlooking it, or could that be added in the future?

applications/object_detection_torch/main.cpp
mocsharp: Inside a Holohub container, I was able to launch the app using both of the following:
```
$ ./run launch object_detection_torch  --extra_args "--config /workspace/holohub/applications/object_detection_torch/object_detection_torch.yaml"
$ ./run launch object_detection_torch  --extra_args /workspace/holohub/applications/object_detection_torch/object_detection_torch.yaml
```

The positional argument is handled right below the switch block if the user passes in the YAML config file without a `--config` switch, as noted in the second command above.

tbirdso: Right, I wanted to note that the following command is also now a valid equivalent of those two examples above:
```sh
./run launch object_detection_torch  --extra_args "/workspace/holohub/applications/object_detection_torch/object_detection_torch.yaml --config path/to/my/bad/config.yaml
```

I think it is OK to favor backwards compatibility as you've noted and move forward.

mocsharp: In that case, the value passed in with `--config` will be overwritten. 
I will add a check to avoid that but unlikely people will use both.

mocsharp:
```
$ ./run launch object_detection_torch  --extra_args "--config /workspace/holohub/applications/object_detection_torch/object_detection_torch.yaml path_to_bad.yaml"

[command] export PYTHONPATH=${PYTHONPATH}:/opt/nvidia/holoscan/lib/cmake/holoscan/../../../python/lib:/workspace/holohub/build/object_detection_torch/python/lib:/workspace/holohub && export HOLOHUB_DATA_PATH=/workspace/holohub/data && export HOLOSCAN_INPUT_PATH=/opt/nvidia/holoscan/data
[command] cd /workspace/holohub/build/object_detection_torch/applications/object_detection_torch
[command] ./object_detection_torch --data /workspace/holohub/data/object_detection_torch --config /workspace/holohub/applications/object_detection_torch/object_detection_torch.yaml path_to_bad.yaml
[warning] [main.cpp:245] Configuration file already set using --config, ignoring 'path_to_bad.yaml'
```

tbirdso: Thanks, the warning rather than override looks good.

Could you please remove autoformatting style updates so that only the handful of functional changes are included?

mocsharp: Done, restore the unformatted code.

gigony (approved):
Thanks @mocsharp for updating the application! It looks good to me.

> Is the --config parsing requirement documented in the [HAP specification](https://docs.nvidia.com/holoscan/sdk-user-guide/cli/hap.html#)? I see environment variables but not CLI requirements. Am I overlooking it, or could that be added in the future?

As @tbirdso mentioned, I had the same question about whether we explicitly specified the requirement or not.

applications/object_detection_torch/main.cpp
mocsharp: @gigony @tbirdso 

> Is the --config parsing requirement documented in the [HAP specification](https://docs.nvidia.com/holoscan/sdk-user-guide/cli/hap.html#)? I see environment variables but not CLI requirements. Am I overlooking it, or could that be added in the future?

This is not documented as its internal behavior: the CLI always passes `--config <file-path>` to the application when launching it. @gigony and I discussed this while designing distributed applications and the CLI. 
I will update Holoscan [CLI packaging doc](https://docs.nvidia.com/holoscan/sdk-user-guide/holoscan_packager.html) to include that requirement.

tbirdso (approved):
Thanks @mocsharp ! Changes look good to me

PR #506, opened 2024-09-23
Improving Holohub installation

jjomier:
Installing extensions, operators and applications in a similar directory structure as Holoscan SDK


CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/deps)
 include(HoloHubConfigHelpers)
 
 # Set install directory
-if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+if(CMAKE_INSTALL_PREFIX_TO_SOURCE)
```
tbirdso, line 28: @jjomier Can you clarify why the change from [`CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT`](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT.html) is required?

jjomier: I was debating about that, but it seems that might makes things easier. Let me revert back.

jjomier:
@mocsharp OK to merge?

mocsharp:
> @mocsharp OK to merge?

This breaks the packaging scripts. Please open an issue, and I will find time to fix the scripts. Thanks.

applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
```diff
@@ -156,29 +156,10 @@ endif()
 # Install application and dependencies into the install/ directory for packaging
 install(
   TARGETS endoscopy_tool_tracking
-  DESTINATION endoscopy_tool_tracking_cpp
-  COMPONENT "holohub-apps"
+  DESTINATION endoscopy_tool_tracking/cpp
```
tbirdso, line 159: Following discussion with @jjomier @mocsharp , we are internally discussing realignment to produce a new install tree per build environment. I.e., suggest the following commands:
```
./dev_container build_and_install endoscopy_tool_tracking
./dev_container build_and_Install multiai_ultrasound
```
May generate one install tree per __app build environment__:
```
install/
├── endoscopy_tool_tracking
  ├── bin
  ├── lib
  ├── python
...
├── multiai_ultrasound
  ├── bin
  ├── lib
  ├── python
```
That reflection of the build environment and installation standard practices may reduce effort for aligning subsequent packaging interests.

tbirdso (changes requested):

applications/endoscopy_tool_tracking/cpp/CMakeLists.txt
jjomier: Seems that we are good with the current installation tree.

jjomier:
It won't work out of the box unless we set the proper paths. Let me see if I can modify the INSTALL_RPATH.

tbirdso (approved):
Updates look good, thank you! We have converged on the structure:

```
/ bin
 | - / myapp
 | --- / cpp
 | --- / python
/include
/ lib
 | - / gxf_extensions
/ python
/ share
```

PR #515, opened 2024-09-26
Added support for build_and_run for adv_networking_bench

jjomier:
Added Docker container to support build_and_run and also added install() command for CMake installation


applications/adv_networking_bench/python/CMakeLists.txt
```diff
@@ -17,4 +17,9 @@
 # to set dependencies on adequate operators
 # add_custom_target(advanced_networking_bench_python ALL
 #   DEPENDS advanced_network_python
-# )
\ No newline at end of file
+# )
+
+install(
+  FILES main.py
+  DESTINATION adv_networking_bench/python
```
tbirdso, line 24: Not sure yet about `DESTINATION` here, suggest we revisit in https://github.com/nvidia-holoscan/holohub/pull/506. Not necessarily a blocker.

applications/adv_networking_bench/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+ARG BASE_IMAGE
+ARG GPU_TYPE
+
+FROM ${BASE_IMAGE} as base
+
+ARG UBUNTU_VERSION=22.04
+ARG OFED_VERSION=24.01-0.3.3.1
```
tbirdso, line 23: Is this version standard? Seems that 24.07-0.6.1.0 will be best for stack alignment, unless there is a clear reason to pin to 24.01.

tbirdso (commented):
Generally looks good. Couple of comments.

applications/adv_networking_bench/python/CMakeLists.txt
jjomier: @tbirdso I think we are in agreement that this is good right?

applications/adv_networking_bench/Dockerfile
cliffburdick: There's no standard version of OFED, but the ANO and Holoscan both require it. we should pick one version to use for both.

applications/adv_networking_bench/python/CMakeLists.txt
tbirdso: No, outcome from our discussion was that apps should install under `bin`:
```
holohub/
  ├── install-<arch>/
    ├── bin
        ├── myapp.py
```

Having an app directory `install/adv_networking_bench/<language>` is nonstandard and would not generalize well.

jjomier: I don't think that would work as some application configuration might have the same names. Why is it bad to pub it in bin/adv_networking_bench/python as we have for HSDK examples?

tbirdso: >  some application configuration might have the same file names (config.yaml).

Good point.

> Why is it bad to put it in bin/adv_networking_bench/python as we have for HSDK examples?

Let's do that. It may complicate planned HAP support a bit more but I think it is the best compromise.

applications/adv_networking_bench/Dockerfile
jjomier: @tbirdso @cliffburdick OK to use 24.07-0.6.1.0?

cliffburdick: yes

jjomier:
@tbirdso @cliffburdick can we merge this?

tbirdso (approved):

PR #516, opened 2024-09-27
cast enums to int for fmt v10 compatibility

grlee77:
This follow up to #514 fixes a few more CUDA_TRY macros that were missed there along with also casting `DLDeviceType` and `NppStatus` enums when logging them.


agirault (approved):
looks good, ty!

tbirdso (approved):
Thank you!

PR #517, opened 2024-09-27
Update run script to pass file to codespell for ignored words instead of explicit word list on command line

MMelQin:
Description:

On resolving false positive lint errors in one of my branches, I saw that the run script calls codespell with ignored words on the command line. The call appears twice in the script, which causes duplication of the ignore word list.

Per codespell [README](https://github.com/codespell-project/codespell?tab=readme-ov-file#ignoring-words), using a file containing the ignored words (one per line) and passing the file as option to the command is an alternative, and, it is better and more consistent with the exclude file option that the script already uses.

Testing:

Tested the updated script with the newly added ignored word file, and codespell ignored the words as specified with no false positives.


run
```diff
@@ -459,7 +459,7 @@ lint() {
       run_command isort ${DIR_TO_RUN} || exit_code=1
       # Fix python black code formatting issues, run:
       run_command black ${DIR_TO_RUN} || exit_code=1
-      run_command codespell -w -i 3 ${DIR_TO_RUN} -L nd,ba,activ,Activ,dne,unexpect,hsi,donot,thirdparty \
+      run_command codespell -w -i 3 ${DIR_TO_RUN} --ignore-words codespell_ingore_words.txt \
```
tbirdso, line 462: Filename typo:

```suggestion
      run_command codespell -w -i 3 ${DIR_TO_RUN} --ignore-words codespell_ignore_words.txt \
```

tbirdso (changes requested):
Nice! Small typo correction and then looks good to merge

tbirdso (approved):
Great, thanks!

PR #523, opened 2024-10-02
Fix TRT depreciation warnings

jjomier:
kEXPLICIT_BATCH is Ignored because networks are always "explicit batch" in TensorRT 10.0.

### DIFF
--- a/gxf_extensions/lstm_tensor_rt_inference/tensor_rt_inference.cpp
+++ b/gxf_extensions/lstm_tensor_rt_inference/tensor_rt_inference.cpp
@@ -635,8 +635,13 @@ gxf::Expected<std::vector<char>> TensorRtInference::convertModelToEngine() {
   if (enable_fp16_.get()) { builderConfig->setFlag(nvinfer1::BuilderFlag::kFP16); }
 
   // Parses ONNX with explicit batch size for support of dynamic shapes/batch
-  NvInferHandle<nvinfer1::INetworkDefinition> network(builder->createNetworkV2(
-      1U << static_cast<uint32_t>(nvinfer1::NetworkDefinitionCreationFlag::kEXPLICIT_BATCH)));
+  #if NV_TENSORRT_MAJOR < 10
+    const auto explicitBatch =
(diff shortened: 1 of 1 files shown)


tbirdso (approved):
Thanks!

gxf_extensions/lstm_tensor_rt_inference/tensor_rt_inference.cpp
```diff
@@ -635,8 +635,13 @@ gxf::Expected<std::vector<char>> TensorRtInference::convertModelToEngine() {
   if (enable_fp16_.get()) { builderConfig->setFlag(nvinfer1::BuilderFlag::kFP16); }
 
   // Parses ONNX with explicit batch size for support of dynamic shapes/batch
-  NvInferHandle<nvinfer1::INetworkDefinition> network(builder->createNetworkV2(
-      1U << static_cast<uint32_t>(nvinfer1::NetworkDefinitionCreationFlag::kEXPLICIT_BATCH)));
+  #if NV_TENSORRT_MAJOR < 10
+    const auto explicitBatch =
+              1U << static_cast<uint32_t>(nvinfer1::NetworkDefinitionCreationFlag::kEXPLICIT_BATCH);
+  #else
+    const auto explicitBatch = 1U;
```
AndreasHeumann, line 642: Shouldn't this be 0U instead of 1U?

jjomier: good point!

tbirdso: Note that we currently use `1` in 2.6 development (TRT 10.3), would need to realign there too. May need more discussion: https://gitlab-master.nvidia.com/holoscan/holoscan-sdk/-/blob/main/public/modules/holoinfer/src/infer/trt/utils.cpp?blame=1#L94

shekhardw: Discussed this with Tom before that value must be 0U, but seems like the recommendation from TRT team is to use a value of 1U.

jjomier: What's the final consensus? use 1U?

PR #524, opened 2024-10-03
Fix setup script for runtime container

jjomier:
When using a base image based on Ubuntu 22.04 and installation of Holoscan SDK, some dependencies are missing.

### DIFF
--- a/run
+++ b/run
@@ -379,10 +379,13 @@ setup() {
      install_cuda_dependencies_package libcudnn8-dev cuda$short_cuda_version
   fi
 
-  # Make sure we do not install any cuda12 packages
-  install_cuda_dependencies_package libnvinfer-dev cuda$short_cuda_version
-  install_cuda_dependencies_package libnvinfer-plugin-dev cuda$short_cuda_version
-  install_cuda_dependencies_package libnvonnxparsers-dev cuda$short_cuda_version
+  # Check the current version of libnvinfer-bin
+  installed_libnvinferbin=$(dpkg --status libnvinfer-bin 2>/dev/null | grep -Po '^Version: \K[^-]*')
+  installed_libnvinferbin_version=${installed_libnvinferbin%.*}
+  install_cuda_dependencies_package libnvinfer-headers-dev $installed_libnvinferbin_version
+  install_cuda_dependencies_package libnvinfer-dev $installed_libnvinferbin_version
+  install_cuda_dependencies_package libnvinfer-plugin-dev $installed_libnvinferbin_version
+  install_cuda_dependencies_package libnvonnxparsers-dev $installed_libnvinferbin_version
 
   # Install the autocomplete
   echo "Installing autocomplete"
--- a/run
+++ b/run
@@ -334,8 +334,8 @@ setup() {
   fi
 
   # git
-  git_version=$(dpkg --status git | grep -Po '^Version: \K[^-]*')
-  if [[ $git_version == "" ]]; then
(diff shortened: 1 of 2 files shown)


run
```diff
@@ -333,6 +333,12 @@ setup() {
     apt-get install --no-install-recommends -y libv4l-dev
   fi
 
+  # git
+  git_version=$(dpkg --status git | grep -Po '^Version: \K[^-]*')
+  if [[ $git_version == "" ]]; then
```
tbirdso, line 338: Could we simplify to `which git`? Seems OK if git is already available in `PATH` and not necessarily installed via dpkg.

run
```diff
@@ -373,10 +379,13 @@ setup() {
      install_cuda_dependencies_package libcudnn8-dev cuda$short_cuda_version
   fi
 
-  # Make sure we do not install any cuda12 packages
-  install_cuda_dependencies_package libnvinfer-dev cuda$short_cuda_version
-  install_cuda_dependencies_package libnvinfer-plugin-dev cuda$short_cuda_version
-  install_cuda_dependencies_package libnvonnxparsers-dev cuda$short_cuda_version
+  # Check the current version of libnvinfer-bin
+  installed_libnvinferbin=$(dpkg --status libnvinfer-bin 2>/dev/null | grep -Po '^Version: \K[^-]*')
+  installed_libnvinferbin_version=${installed_libnvinferbin%.*}
```
tbirdso, line 384: Could you comment on the behavior you were seeing that prompted this change? Did we have different component version installed by default?

tbirdso: Isn't the CUDA version tied to the `nvinfer` package version? Not sure I understand why we want to strip the CUDA version from the package when it is a common field:
```
# apt list | grep nvinfer

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

libnvinfer-bin/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-dev/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-dispatch-dev/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-dispatch10/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-headers-dev/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-headers-plugin-dev/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-lean-dev/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-lean10/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-plugin-dev/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-plugin10/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-vc-plugin-dev/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer-vc-plugin10/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
libnvinfer10/now 10.3.0.26-1+cuda12.5 amd64 [installed,local]
```

jjomier: For context the previous implementation was done to make sure cuda 12 was not installed and only cuda 11. This is clearly not working for CUDA 12.2 vs 12.6. The new code will help.

tbirdso: Got it, thanks for clarifying.

tbirdso (approved):

PR #530, opened 2024-10-09
Add Holoviz UI example.

AndreasHeumann:


tbirdso (approved):
Code looks good. Please ping for merge after resolving the conflict.

AndreasHeumann:
@tbirdso Rebased and resolved conflicts. I also removed the callback from the SourceOp to the application. I planned to modify the source based on UI settings but did not do it in the end. With that I makes no sense to have the callback in the source  op.

PR #531, opened 2024-10-11
Update package-app.sh scripts to reflect recent cmake install changes

mocsharp:
Update the packaging scripts to copy necessary files to the application binary file for packaging.


jjomier (approved):

applications/endoscopy_tool_tracking/cpp/package-app.sh
```diff
@@ -32,7 +32,14 @@ if [ $(get_host_arch) == "aarch64" ]; then
     PLATFORM=igx-orin-devkit
 fi
 
+echo -e "Copying the required files to the application directory..."
+cp -R "$GIT_ROOT/install/lib/liblstm_tensor_rt_inference.so" "$APP_PATH"
+cp "$GIT_ROOT/install/lib/libtool_tracking_postprocessor.so" "$APP_PATH"
+cp "$GIT_ROOT/install/lib/gxf_extensions/libgxf_lstm_tensor_rt_inference_lib.so" "$APP_PATH"
+cp "$GIT_ROOT/install/lib/gxf_extensions/libgxf_lstm_tensor_rt_inference.so" "$APP_PATH"
```
tbirdso, line 39: Under updated CMake install support, we should assume that CMake is responsible for determining all of the libraries that need to be distributed in `install/lib`. Can we bulk copy those libraries to the application directory instead of naming them individually?

```suggestion
cp "$GIT_ROOT/install/lib/*" "$APP_PATH"
cp "$GIT_ROOT/install/lib/gxf_extensions/*" "$APP_PATH"
```

applications/endoscopy_tool_tracking/python/package-app.sh
```diff
@@ -32,7 +32,15 @@ if [ $(get_host_arch) == "aarch64" ]; then
     PLATFORM=igx-orin-devkit
 fi
 
+echo -e "Copying the required files to the application directory..."
+cp -R "$GIT_ROOT/install/lib/liblstm_tensor_rt_inference.so" "$APP_PATH"
+cp "$GIT_ROOT/install/lib/libtool_tracking_postprocessor.so" "$APP_PATH"
+cp "$GIT_ROOT/install/lib/gxf_extensions/libgxf_lstm_tensor_rt_inference_lib.so" "$APP_PATH"
+cp "$GIT_ROOT/install/lib/gxf_extensions/libgxf_lstm_tensor_rt_inference.so" "$APP_PATH"
+cp -R "$GIT_ROOT/install/python/lib/holohub" "$APP_PATH"
```
tbirdso, line 40: Same as above, with addition of Python module:

```suggestion
cp "$GIT_ROOT/install/lib/*" "$APP_PATH"
cp "$GIT_ROOT/install/lib/gxf_extensions/*" "$APP_PATH"
cp -R "$GIT_ROOT/install/python/lib/holohub" "$APP_PATH"
```

tbirdso (changes requested):

applications/endoscopy_tool_tracking/cpp/package-app.sh
mocsharp: The problem is that the `install/` dir may contain other application dependencies; for example, if a user builds and installs two different applications, the `install/lib/` dir may contain dependencies from those two applications.

applications/endoscopy_tool_tracking/python/package-app.sh
mocsharp: Same as above

applications/endoscopy_tool_tracking/cpp/package-app.sh
tbirdso: > if a user builds and installs two different applications

Your concern is valid. We currently explicitly disallow this case under the revised HoloHub installation design. We/I should update HoloHub documentation to better reflect that expectation.

In the event that a user does attempt to install multiple applications at once (not supported), the worst case for CLI is that extra library copies occur. IMO that issue presents low risk.

Holding to the suggestion here, let's allocate dependency management to the build system rather than explicitly inlining dependencies in scripts where avoidable.

tbirdso (approved):
Thank you @mocsharp !

PR #534, opened 2024-10-14
Fixing CMake warnings for volume_rendering operator

jjomier:
CMake fetchcontent_populate has been deprecated: https://cmake.org/cmake/help/latest/policy/CMP0169.html

### DIFF
--- a/operators/volume_renderer/CMakeLists.txt
+++ b/operators/volume_renderer/CMakeLists.txt
@@ -52,16 +52,11 @@ FetchContent_Declare(
 # enable CMP0077 to allow overwriting option() statements in FetchContent sub-projects
 cmake_policy(SET CMP0077 NEW)
 
-FetchContent_GetProperties(ClaraViz)
-if(NOT claraviz_POPULATED)
-    FetchContent_Populate(ClaraViz)
-    set(CLARA_VIZ_PUBLIC_CMAKE_TOOLS_DIR "${claraviz_SOURCE_DIR}/cmake")
-    add_subdirectory(${claraviz_SOURCE_DIR}/thirdparty ${claraviz_BINARY_DIR}/thirdparty)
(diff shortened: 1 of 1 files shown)


operators/volume_renderer/CMakeLists.txt
```diff
@@ -52,16 +52,11 @@ FetchContent_Declare(
...
-    FetchContent_Populate(ClaraViz)
-    set(CLARA_VIZ_PUBLIC_CMAKE_TOOLS_DIR "${claraviz_SOURCE_DIR}/cmake")
-    add_subdirectory(${claraviz_SOURCE_DIR}/thirdparty ${claraviz_BINARY_DIR}/thirdparty)
-    add_subdirectory(${claraviz_SOURCE_DIR}/src ${claraviz_BINARY_DIR}/src)
-endif()
-set(FETCHCONTENT_QUIET ON)
-
+FetchContent_MakeAvailable(ClaraViz)
 set(CLARA_VIZ_PUBLIC_CMAKE_TOOLS_DIR "${claraviz_SOURCE_DIR}/cmake")
+add_subdirectory(${claraviz_SOURCE_DIR}/src ${claraviz_BINARY_DIR}/src)
```
tbirdso, line 57: I don't think we should need `add_subdirectory` after `FetchContent_MakeAvailable`. Can you confirm whether this is still required to build?

jjomier: Doesn't work without it. @AndreasHeumann to confirm.

tbirdso: Interesting, thanks for confirming. OK to move forward as-is, though would be good to understand if we can remove this in the long term.

tbirdso (approved):

operators/volume_renderer/CMakeLists.txt
AndreasHeumann: I think this is caused by the way ClaraViz is setup. It's using `ExternalProject_Add` but I don't remember the details.

PR #541, opened 2024-10-17
Adding capability to add additional arguments to build container

jjomier:
Passing --secret to `build` and `build_and_run` functions

### DIFF
--- a/dev_container
+++ b/dev_container
@@ -486,6 +486,7 @@ Usage: ./dev_container build [options]
     --verbose : Print variables passed to docker build command
     --no-cache : Do not use cache when building the image
     --build-arg : provides extra arguments to docker build command
+    --secret : provides secret to docker build command
     '
 }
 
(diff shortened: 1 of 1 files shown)


dev_container
```diff
@@ -486,6 +486,7 @@ Usage: ./dev_container build [options]
     --verbose : Print variables passed to docker build command
     --no-cache : Do not use cache when building the image
     --build-arg : provides extra arguments to docker build command
+    --secret : provides secret to docker build command
```
tbirdso, line 489: I'm OK to move forward with this, but wouldn't it be better in the future to accept general `--docker_opts`? I'm concerned about continuously expanding `build` and `build_and_run` options to mirror the Docker CLI as opposed to simply forwarding general arguments.

tbirdso (approved):

dev_container
jjomier: I did try that but passing options with -- was causing issue with the parsing of the command line. Let me see if I can address that.

PR #544, opened 2024-10-21
Sample app: NV-CLIP NIM

mocsharp:
This PR adds a new sample application that demonstrates the use of NV-CLIP NIM: NV-CLIP is a multimodal embeddings model for image and text.


applications/nvidia_nim/nvidia_nim_nvclip/metadata.json
```diff
@@ -0,0 +1,37 @@
...
+		"version": "1.0",
+		"changelog": {
+			"1.0": "Initial Release"
+		},
+		"holoscan_sdk": {
+			"minimum_required_version": "1.0.3",
+			"tested_versions": [
+				"1.0.3",
+				"2.1.0"
+			]
```
tbirdso, line 20: 2.5.0?

run
```diff
@@ -469,7 +469,8 @@ lint() {
       # Fix python black code formatting issues, run:
       run_command black ${DIR_TO_RUN} || exit_code=1
       run_command codespell -w -i 3 ${DIR_TO_RUN} --ignore-words codespell_ignore_words.txt \
-                            --skip="*.onnx,*.min.js,*.min.js.map,Contrastive_learning_Notebook.ipynb,./data" || exit_code=1
+                            --skip="*.onnx,*.min.js,*.min.js.map,Contrastive_learning_Notebook.ipynb,./data" \
+                            --skip="./applications/nvidia_nim/nvidia_nim_nvclip/nvidia_nim.yaml" || exit_code=1
```
tbirdso, line 473: Could you elaborate on this change? Would be good to move images to `.png` or similar instead of `.yaml` fields

tbirdso (commented):
Looks like the YAML file has image bytestrings, could we move those resources to standard image files and read in at runtime?

run
mocsharp: Using base64 encoded images is because it avoids broken links if the images were removed. Plus, the API takes base64 encoded images.  The other option is to move them into a separate file.

applications/nvidia_nim/nvidia_nim_nvclip/metadata.json
mocsharp: Added 2.5.0 and tested both 1.0.3 and 2.1.0

run
mocsharp: Moving the sample encoded images to a separate YAML file also triggered codespell errors.

applications/nvidia_nim/nvidia_nim_nvclip/app.py
```diff
@@ -0,0 +1,196 @@
...
+from requests.models import PreparedRequest
+
+logging.getLogger("httpx").setLevel(logging.WARN)
+logging.getLogger("openai").setLevel(logging.WARN)
+
+logger = logging.getLogger("NVIDIA_NIM_CHAT")
+logging.basicConfig(level=logging.INFO)
+
+
+class OpenAIOperator(Operator):
```
binliunls, line 35: Hi @mocsharp , 
    May I ask if there is a plan to add all NIM API call related functions into the current holoscan [inference operator](https://docs.nvidia.com/holoscan/sdk-user-guide/inference.html#inference-operator)? If so, what's the plan?

Thanks,  
Bin

mocsharp: Hi @binliunls. It is likely not the case with this NVCLIP sample application.

Please feel free to file a new issue if you have any proposals and examples you would like to see. Thank you!

run
```diff
@@ -515,6 +516,7 @@ lint() {
 
     echo "Code spelling"
     run_command codespell $DIR_TO_RUN --skip="*.onnx,*.min.js,*.min.js.map,Contrastive_learning_Notebook.ipynb,./data" \
+                          --skip="./applications/nvidia_nim/nvidia_nim_nvclip/nvidia_nim.yaml" \
```
MMelQin, line 519: Please consider using [codelspell ](https://github.com/codespell-project/codespell?tab=readme-ov-file#using-a-config-file)`-config` option or another codespell option to provide the list of excluded files to avoid duplication. I have recently added the ignore-words file to consolidate the false positive words, and the files can also be done similarly.

run
MMelQin: Please see below on using a config file to consolidate the ignored files in one place.

run
mocsharp: That's what I initially did but it had way too many words to add because of base64 string. I'll think of something else.

MMelQin: The last sentence of my comment must have confused you. I meant to say, use a file containing the excluded files instead of listing all of the files (wild card and full path) on the command line.

run
mocsharp: Removed.

applications/nvidia_nim/nvidia_nim_nvclip/README.md
```diff
@@ -0,0 +1,73 @@
...
+  api_key:
+
+```
+
+`base_url`: The URL of your NIM instance. Defaults to NVIDIA hosted NIMs.
+`api_key`: Your API key to access NVIDIA hosted NIMs.
+
+## Run the sample application
+
+There are a couple of options to run the sample application:
```
MMelQin, line 28: Only one option is described below, since the section `Start the Application` works in the container. For using `./run launch <app>` in host dev env, you'd have to `./run build` and pip install reqs for Python app.

tbirdso (approved):
Looks good to me! Thanks for resolving image concerns.

PR #545, opened 2024-10-22
multiai_ultrasound: Fix allocation error when using AJA source

AndreasHeumann:
Use increased block size for FormatConverterOP when input is RGBA instead of RGB because the converter needs to add an additional temporary buffer in that case.


jjomier (approved):

tbirdso:
🚀

PR #548, opened 2024-10-23
Add .viminfo and .lesshst to .gitignore

JohnMoon-VTS:
These files are created during many developer flows and, with the dev container mounting this Git repo as the home directory, they pollute the Git working tree.


jjomier (approved):
Thank you!

PR #550, opened 2024-10-25
Optimize endoscopy tool tracking app

AndreasHeumann:
Optimize endoscopy tool tracking app

Modify ToolTrackingPostProcessorOp to generate coordinates using
CUDA. This avoids the roundtrip from GPU to CPU memory and back.

Use RMMAllocator for video replayer.

Fix type  of `dla_core` parameter of TensorRtInference.

Remove unused `clock` parameter.

Use single `out` port instead of `out_coords` and `out_mask` since
both outputs are now in device memory.

Latency measured with flow benchmarking on x86 with RTX 6000 Ada
and replayer `realtime`set to `false`:
before optimization 4.8 ms, after optimization 3.09 ms.

### DIFF
--- a/applications/endoscopy_tool_tracking/cpp/main.cpp
+++ b/applications/endoscopy_tool_tracking/cpp/main.cpp
@@ -17,6 +17,7 @@
 
 #include <getopt.h>
 
+#include <holoscan/holoscan.hpp>
 #include <holoscan/operators/aja_source/aja_source.hpp>
 #include <holoscan/operators/format_converter/format_converter.hpp>
 #include <holoscan/operators/holoviz/holoviz.hpp>
@@ -24,7 +25,6 @@
 #include <holoscan/operators/video_stream_replayer/video_stream_replayer.hpp>
 #include <lstm_tensor_rt_inference.hpp>
 #include <tool_tracking_postprocessor.hpp>
-#include "holoscan/holoscan.hpp"
 #ifdef VTK_RENDERER
 #include <vtk_renderer.hpp>
 #endif
@@ -37,6 +37,11 @@
 #ifdef YUAN_QCAP
 #include <qcap_source.hpp>
 #endif
+
+#include <holoscan/version_config.hpp>
+
+#define HOLOSCAN_VERSION \
+  (HOLOSCAN_VERSION_MAJOR * 10000 + HOLOSCAN_VERSION_MINOR * 100 + HOLOSCAN_VERSION_PATCH)
 
 class App : public holoscan::Application {
  public:
@@ -111,6 +116,10 @@ class App : public holoscan::Application {
       height = 480;
       source = make_operator<ops::VideoStreamReplayerOp>(
           "replayer", from_config("replayer"), Arg("directory", datapath));
+#if HOLOSCAN_VERSION >= 20600
+      // the RMMAllocator supported since v2.6 is much faster than the default UnboundAllocator
+      source->add_arg(Arg("allocator", make_resource<RMMAllocator>("video_replayer_allocator")));
+#endif
       source_block_size = width * height * 3 * 4;
       source_num_blocks = 2;
     }
@@ -151,16 +160,19 @@ class App : public holoscan::Application {
             "pool", 1, lstm_inferer_block_size, lstm_inferer_num_blocks),
         Arg("cuda_stream_pool") = cuda_stream_pool);
 
-    const uint64_t tool_tracking_postprocessor_block_size = 107 * 60 * 7 * 4;
-    const uint64_t tool_tracking_postprocessor_num_blocks = 2;
+    // the tool tracking post process outputs
+    // - a RGBA float32 color mask
+    // - coordinates with x,y and size in float32
+    const uint64_t tool_tracking_postprocessor_block_size =
+        std::max(107 * 60 * 7 * 4 * sizeof(float), 7 * 3 * sizeof(float));
+    const uint64_t tool_tracking_postprocessor_num_blocks = 2 * 2;
     auto tool_tracking_postprocessor = make_operator<ops::ToolTrackingPostprocessorOp>(
         "tool_tracking_postprocessor",
         Arg("device_allocator") =
             make_resource<BlockMemoryPool>("device_allocator",
                                            1,
                                            tool_tracking_postprocessor_block_size,
-                                           tool_tracking_postprocessor_num_blocks),
-        Arg("host_allocator") = make_resource<UnboundedAllocator>("host_allocator"));
+                                           tool_tracking_postprocessor_num_blocks));
 
     if (this->visualizer_name == "holoviz") {
       std::shared_ptr<BlockMemoryPool> visualizer_allocator;
@@ -188,17 +200,7 @@ class App : public holoscan::Application {
 
     // Flow definition
     add_flow(lstm_inferer, tool_tracking_postprocessor, {{"tensor", "in"}});
-
-    if (this->visualizer_name == "holoviz") {
-      add_flow(tool_tracking_postprocessor,
-               visualizer_operator,
-               {{"out_coords", input_annotations_signal}, {"out_mask", input_annotations_signal}});
-    } else {
-      // device tensor on the out_mask port is not used by VtkRendererOp
-      add_flow(tool_tracking_postprocessor,
-               visualizer_operator,
-               {{"out_coords", input_annotations_signal}});
-    }
+    add_flow(tool_tracking_postprocessor, visualizer_operator, {{"out", input_annotations_signal}});
 
     std::string output_signal = "output";  // replayer output signal name
     if (source_ == "deltacast") {
--- a/operators/lstm_tensor_rt_inference/lstm_tensor_rt_inference.cpp
+++ b/operators/lstm_tensor_rt_inference/lstm_tensor_rt_inference.cpp
@@ -94,7 +94,8 @@ void LSTMTensorRTInferenceOp::setup(OperatorSpec& spec) {
              "dla_core",
              "DLA Core",
              "DLA Core to use. Fallback to GPU is always enabled. "
-             "Default to use GPU only.");
+             "Default to use GPU only.",
+             ParameterFlag::kOptional);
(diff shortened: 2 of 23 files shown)


operators/lstm_tensor_rt_inference/lstm_tensor_rt_inference.cpp
```diff
@@ -94,7 +94,8 @@ void LSTMTensorRTInferenceOp::setup(OperatorSpec& spec) {
              "dla_core",
              "DLA Core",
              "DLA Core to use. Fallback to GPU is always enabled. "
-             "Default to use GPU only.");
+             "Default to use GPU only.",
+             ParameterFlag::kOptional);
```
grlee77, line 98: thanks for adding this to suppress warnings in the logs when it is not provided

operators/lstm_tensor_rt_inference/python/lstm_tensor_rt_inference.cpp
```diff
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights
+ * reserved. SPDX-License-Identifier: Apache-2.0
```
grlee77, line 3:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 * SPDX-License-Identifier: Apache-2.0
```

operators/tool_tracking_postprocessor/CMakeLists.txt
```diff
@@ -25,10 +26,21 @@ add_library(tool_tracking_postprocessor SHARED
   tool_tracking_postprocessor.cuh
   )
 
-set_target_properties(tool_tracking_postprocessor PROPERTIES CUDA_ARCHITECTURES "70;80")
+set_target_properties(tool_tracking_postprocessor
```
grlee77, line 29: I see CUDA_ARCHITECTURES was removed. Are these already being set at the top-level of the project?

operators/tool_tracking_postprocessor/tool_tracking_postprocessor.cpp
```diff
@@ -1,6 +1,6 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights
+ * reserved. SPDX-License-Identifier: Apache-2.0
```
grlee77, line 3:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 * SPDX-License-Identifier: Apache-2.0
```

operators/tool_tracking_postprocessor/python/tool_tracking_postprocessor.cpp
```diff
@@ -73,12 +73,11 @@ class PyToolTrackingPostprocessorOp : public ToolTrackingPostprocessorOp {
   // Define a constructor that fully initializes the object.
   PyToolTrackingPostprocessorOp(
       Fragment* fragment, const py::args& args, std::shared_ptr<Allocator> device_allocator,
-      std::shared_ptr<Allocator> host_allocator, float min_prob = 0.5f,
+      float min_prob = 0.5f,
       std::vector<std::vector<float>> overlay_img_colors = VIZ_TOOL_DEFAULT_COLORS,
       std::shared_ptr<holoscan::CudaStreamPool> cuda_stream_pool = nullptr,
       const std::string& name = "tool_tracking_postprocessor")
       : ToolTrackingPostprocessorOp(ArgList{Arg{"device_allocator", device_allocator},
-                                            Arg{"host_allocator", host_allocator},
```
grlee77, line 81: If this is removed, I think you likely also need to update the other variants of the endoscopy tool tracking app that are using this operator:

- `applications/endoscopy_tool_tracking_distributed`
- `applications/h264/h264_endoscopy_tool_tracking`

operators/tool_tracking_postprocessor/tool_tracking_postprocessor.cpp
```diff
@@ -109,118 +90,105 @@ void ToolTrackingPostprocessorOp::compute(InputContext& op_input, OutputContext&
...
   if (stream_handler_result != GXF_SUCCESS) {
     throw std::runtime_error("Failed to add the CUDA stream to the outgoing messages");
   }
 
-  auto result_host = gxf::Entity(std::move(out_message_host.value()));
-  auto result_device = gxf::Entity(std::move(out_message_device.value()));
+  auto result_host = gxf::Entity(std::move(out_coords_message.value()));
+  auto result_device = gxf::Entity(std::move(out_mask_message.value()));
   op_output.emit(result_host, "out_coords");
   op_output.emit(result_device, "out_mask");
```
grlee77, line 193: Originally these tensors were both on a single "out" port, but I [had to change to separate ports for the distributed version of the app](https://github.com/nvidia-holoscan/holohub/pull/262) because UCX cannot mix host and device tensors in the same entity transmitted by `UcxTransmitter` (the same UCX transport layer, e.g. cuda_copy, etc. has to be used for all tensors in a given entity)
```suggestion
  op_output.emit(result_host, "out_coords");
  op_output.emit(result_device, "out_mask");
```

Now that it looks like `out_coords` and `out_mask` are both in device memory, then it would be good to go back to having only a single "out" port again, which may be a bit more efficient, particularly in the distributed case as it reduces the number of separate messages that have to be sent between the fragments.

applications/endoscopy_tool_tracking/cpp/main.cpp
```diff
@@ -151,16 +160,19 @@ class App : public holoscan::Application {
             "pool", 1, lstm_inferer_block_size, lstm_inferer_num_blocks),
         Arg("cuda_stream_pool") = cuda_stream_pool);
 
-    const uint64_t tool_tracking_postprocessor_block_size = 107 * 60 * 7 * 4;
-    const uint64_t tool_tracking_postprocessor_num_blocks = 2;
+    // the tool tracking post process outputs
+    // - a RGBA float32 color mask
+    // - coordinates with x,y and size in float32
+    const uint64_t tool_tracking_postprocessor_block_size =
```
grlee77, line 166: Is leaving it as `BlockMemoryPool` instead of `RMMAllocator` here is mainly so we don't have to assign a different allocator depending on the version of Holoscan? (or was there some performance benefit to keeping the block memory pool instead?)

grlee77 (changes requested):
Thanks, Andreas it looks great. I had two main questions left to consider:

1.) I think a couple other apps using tool tracking will need corresponding changes in the operator arguments (no `host_allocator` in order to continue to work.
2.) Do you want to merge "out_mask" and "out_coords" back to just a single "out" port like they were before #262 was merged

operators/tool_tracking_postprocessor/CMakeLists.txt
AndreasHeumann: No, I don't think so. If it is not specified then it's using the default chosen by ncc. Using "70;80" seems to be somewhat artificially used. Using "all" like in HSDK does not work since the x86 CUDA driver does not include the kernels for mobile GPUs (e.g. SM87 (Orin)). Since the compiled code usually is used on the installed GPU using that option seems to make most sense.

operators/tool_tracking_postprocessor/python/tool_tracking_postprocessor.cpp
AndreasHeumann: Thanks, I forgot these.

operators/tool_tracking_postprocessor/tool_tracking_postprocessor.cpp
AndreasHeumann: Yes, makes sense. I'll change that back to a single output port.

applications/endoscopy_tool_tracking/cpp/main.cpp
AndreasHeumann: This is just to make the code less cluttered and easier to read. For the video replayer we only have RMM or Unbound allocator, so adding the extra code to use RMM allocator makes sense. But for the others I think using BlockMemory and RMM allocator makes no difference.

grlee77: thanks for confirming. I am fine with this as-is then.

operators/tool_tracking_postprocessor/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2034 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tbirdso, line 1:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2023-2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

tbirdso (approved):
Looks good! One copyright fixup. Will wait for approval from @grlee77 before merge.

grlee77 (approved):
Thanks for making the suggested changes, approved!

operators/tool_tracking_postprocessor/CMakeLists.txt
AndreasHeumann: 2034 :-)

AndreasHeumann:
Thanks, fixed the copyright year. Ready to be merged.

PR #553, opened 2024-10-28
Fix package-app.sh scripts for Endoscopy Tool Tracking

mocsharp:
Fix package-app.sh scripts in Endoscopy Tool Tracking app for HSDK 2.6


tbirdso (approved):
Looks good! Verified on my end that this fixes the observed copy error:
```
cp: cannot stat './install/lib/*': No such file or directory
```

PR #563, opened 2024-10-31
Update dev_container script to support Wayland.

AndreasHeumann:
Tested both `./dev_container launch` and `./dev_container vscode` with X11 and Wayland. VSCode already had the options.


tbirdso (approved):
Great, looks good! Can merge after a short time in case others want to review.

PR #571, opened 2024-11-05
Upgrade ANO to DOCA 2.8

e-ago:


operators/advanced_network/Dockerfile
```diff
@@ -24,13 +24,20 @@ ARG BUILDARCH
...
 WORKDIR /opt
 
-RUN    eval `dpkg-architecture` \
-        && wget https://www.mellanox.com/downloads/DOCA/DOCA_v2.7.0/host/doca-host_2.7.0-204000-24.04-ubuntu2204_${BUILDARCH}.deb -O doca-host.deb \
-        && ls -lh && apt install ./doca-host.deb \
-        && apt update \
-        && apt install -y doca-all \
-        && apt install -y doca-gpu doca-gpu-dev
+ARG DEBIAN_FRONTEND=noninteractive
+ARG DOCA_REPO_LINK=https://linux.mellanox.com/public/repo/doca/2.8.0/ubuntu22.04/x86_64
```
cliffburdick, line 32: This looks like it removes BUILDARCH

e-ago: because BUILDARCH says amd64 but the link is at x86_64

e-ago:
@cliffburdick @jjomier if we can merge it today so rmax folks can submit their MR

operators/advanced_network/Dockerfile
cliffburdick: Should we parameterize on that so we don't need two dockerfiles?

cliffburdick (approved):

cliffburdick:
@e-ago I don't have permission to merge

PR #572, opened 2024-11-05
Add psd_pipeline_sim application

JohnMoon-VTS:
When trying to use Holoscan as an SDR framework (like GNU Radio or REDHAWK), there are a few missing pieces such as a generic IQ file reader application and simple blocks like FFT and PSD. It also lacks some support for VITA 49.

We'd like to add an application which takes in file data, does some signal processing, and outputs VITA 49 spectral data packets.

In the future, we plan on adding a VITA 49 ingest operator which leverages ANO, but for now, the file-based simulation is helpful.

One challenge we encountered in developing this pipeline is how to handle metadata flow between operators. In this case, we've dropped a shared data structure in `operators/psd`, but this doesn't seem like the right way to do it. Coming from GNU Radio, there's the concept of [stream tags](https://wiki.gnuradio.org/index.php/Stream_Tags) which could be helpful here. Does Holoscan have any equivalent? This would make it much easier to make generic, share-able blocks.


JohnMoon-VTS:
Thanks to @dylan-eustice for some advice offline about using [Holoscan's metadata flow](https://docs.nvidia.com/holoscan/sdk-user-guide/holoscan_create_app.html#understanding-metadata-flow)! I've updated to use that instead which obviated the need for the ugly shared data structure.

JohnMoon-VTS:
Closing this PR in favor of #632. We may re-introduce in the future if there's interest in the file-based simulation of the pipeline.

PR #587, opened 2024-11-13
modified patch application script to be more explicit in CPP files

sohamm17:
Previously, calls like `holoscan::Application()` would also be modified (e.g., [link](https://github.com/nvidia-holoscan/holohub/blob/3edbcf6eda2bd856faacbb31cdc6ac9934e30033/benchmarks/model_benchmarking/cpp/model_benchmarking.cpp#L60))

Made a little change so that it only considers modification when `holoscan::Application` is being inherited in cpp files.


tbirdso (approved):
💯

PR #604, opened 2024-11-27
Address RMMAllocator error for UCX Endoscopy Tool Tracking application

mocsharp:
This PR addresses a `RMMAllocator` crash when running the application with `realtime` set to `false` (play the video as fast as possible).

```bash
[error] [rmm_allocator.cpp:190] Unexpected error while allocating memory [00007]('video_replayer_allocator') : std::bad_alloc: out_of_memory: RMM failure at:bazel-out/k8-opt/bin/external/rmm/_virtual_includes/rmm/rmm/mr/device/pool_memory_resource.hpp:424: Maximum pool size exceeded
[error] [memory_buffer.hpp:79] video_replayer_allocator Failed to allocate 1229760 size of memory of type 1. Error code: GXF_FAILURE
```

As @grlee77 suggested, the fix is to explicitly set the RMM Allocator's initial and maximum sizes.

Adds the ability to enable/disable data flow benchmarking through configuration.

### DIFF
--- a/applications/distributed/ucx/ucx_endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
+++ b/applications/distributed/ucx/ucx_endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
@@ -21,10 +21,17 @@ extensions:
   - libgxf_serialization.so
   - lib/gxf_extensions/libgxf_lstm_tensor_rt_inference.so
 
+application:
+  title: Endoscopy Tool Tracking - UCX
+  version: 1.0
+  inputFormats: []
+  outputFormats: ["screen"]
+  benchmarking: true # default: false, true to enable Data Flow Benchmarking, false otherwise
+
 replayer:
   basename: "surgical_video"
   frame_rate: 0   # as specified in timestamps
-  repeat: true    # default: false
+  repeat: false    # default: false
   realtime: true  # default: true
   count: 0        # default: 0 (no frame count restriction)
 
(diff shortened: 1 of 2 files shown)


tbirdso:
@mocsharp could you please provide more details on the error and root cause? Are the RMMAllocator crash and benchmarking option related?

mocsharp:
> @mocsharp could you please provide more details on the error and root cause? Are the RMMAllocator crash and benchmarking option related?

Updated description.

applications/distributed/ucx/ucx_endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
```diff
@@ -21,10 +21,17 @@ extensions:
...
+  version: 1.0
+  inputFormats: []
+  outputFormats: ["screen"]
+  benchmarking: true # default: false, true to enable Data Flow Benchmarking, false otherwise
+
 replayer:
   basename: "surgical_video"
   frame_rate: 0   # as specified in timestamps
-  repeat: true    # default: false
+  repeat: false    # default: false
```
tbirdso, line 34: Should HoloHub user defaults be `benchmarking: false` and `repeat: true`?

tbirdso (commented):
One question on defaults, code looks good otherwise. Will wait for approval from @grlee77 to merge.

tbirdso:
ping @grlee77 for review

jjomier:
@grlee77 can you please review?

applications/distributed/ucx/ucx_endoscopy_tool_tracking/cpp/endoscopy_tool_tracking.yaml
grlee77: For benchmarking, it seems okay either way, but we should make the comment consistent with the default

mocsharp: `benchmarking` now defaults to `false` and `repeat=true`

tbirdso:
Thanks all, looks good!

PR #607, opened 2024-12-02
Ext: Link DLPack dependency in GXF Emergent Source extension

tbirdso:
Link the `holoscan::core` CMake target in the Emergent Source extension to ensure that DLPack is found.

`holoscan::core` implicitly provides third party include paths for dependencies packaged with the Holoscan SDK distribution.

Addresses GitHub issue #606.

Verified build error #606 is resolved. Note that this is not fully tested as I do not have access to the Emergent SDK.


gigony (approved):
Thanks @tbirdso for fixing the issue!
It looks good to me!

PR #608, opened 2024-12-02
Add script to do basic system checks

cliffburdick:


operators/advanced_network/managers/adv_network_mgr.cpp
```diff
@@ -142,10 +142,13 @@ AdvNetStatus ANOMgr::allocate_memory_regions() {
...
           const auto alloc_res = cuMemAlloc(&cuptr, align);
+
           if (alloc_res != CUDA_SUCCESS) {
-            HOLOSCAN_LOG_CRITICAL("Could not allocate {:.2f}MB of GPU memory: {}",
+            const char* err_str = nullptr;
+            cuGetErrorString(result, &err_str);
+            HOLOSCAN_LOG_CRITICAL("Could not allocate {:.2f}MB of GPU memory. Error: {}",
                                   align / 1e6,
-                                  static_cast<int>(alloc_res));
+                                  err_str);
```
agirault, line 151: Sounds good, though seems out of scope of this MR

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,632 @@
...
+    """
+    Parses command-line arguments.
+    """
+    parser = argparse.ArgumentParser(
+        description="Check system tuning for ANO performance",
+        epilog=(
+            "Examples:\n"
+            "  python check_system.py -cpu-freq       # Check CPU frequency governor\n"
+            "  python check_system.py -mrrs     # Check MRRS settings for NVIDIA NICs\n\n"
+            "  python check_system.py -mps      # Check max payload size settings for NVIDIA NICs\n\n"
```
agirault, line 32: Outdated.

Though it raises a good point to discuss: if we add "setters", the script should be named something else than `check_system` IMO. The setters and name change could wait for another MR to not block this.

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,632 @@
...
+        except PermissionError:
+            logging.error(
+                f"CPU {cpu}: Permission denied while accessing scaling governor file. Run as root."
+            )
+
+
+def check_mrrs():
+    """
+    Checks if the Maximum Read Request Size (MRRS) of NVIDIA Ethernet controllers
+    is set to 4096.
```
agirault, line 164: Can you give a little bit of background? Mine is `4194304` and I was able to do linerate with the benchmarking app.

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,632 @@
...
+        ),
+    )
+
+    parser.add_argument("--set", choices=["mrrs"], help=("  mrrs      - Update MRRS of NICs\n"))
+    return parser.parse_args()
+
+
+def get_nic_info():
+    """
+    Parses the output of `ibdev2netdev -v` to extract and return a list of tuples,
```
agirault, line 72: some systems don't have this (isn't that technically for RDMA?).

Could we instead do the same as [here](https://github.com/nvidia-holoscan/holohub/blob/1b8c0569f21b216685be18140e9b166cdbfa6c66/applications/adv_networking_bench/cpp/CMakeLists.txt#L130-L140) which doesn't depend on any tool installation?

```bash
bus_address_list=$(lspci | grep -E \"Ethernet.*ConnectX\" | awk '{print $1}')
...
interface_name=$(find /sys/class/net -exec bash -c 'echo {} $(readlink -f {}/device)' \; | grep ${bus_address} | awk '{print $1}')
```

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,632 @@
...
+
+def check_mrrs():
+    """
+    Checks if the Maximum Read Request Size (MRRS) of NVIDIA Ethernet controllers
+    is set to 4096.
+    """
+    try:
+        _, addrs = get_nic_info()
+
+        for pci_address in addrs:
```
agirault, line 169: please loop over interface name as well so we can print it in the log

Example:
```log
2024-12-09 10:39:20 - WARNING - 0005:03:00.0: MRRS is set to 4194304, not 4096.
```

Would be

```log
2024-12-09 10:39:20 - WARNING - NIC enP5p3s0f0np0 [0005:03:00.0]: MRRS is set to 4194304, not 4096.
```

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,632 @@
...
+
+def check_max_payload_size():
+    """
+    Checks the Maximum Payload Size (MPS) of NVIDIA Ethernet controllers
+    from the DevCtl section and ensures it is set to 256 bytes.
+    """
+    try:
+        _, addrs = get_nic_info()
+
+        for pci_address in addrs:
```
agirault, line 200: same comment as https://github.com/nvidia-holoscan/holohub/pull/608/files#r1876238270

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,632 @@
...
+            # Query detailed device information using lspci -vv
+            mps_result = subprocess.run(
+                ["lspci", "-vv", "-s", pci_address], capture_output=True, text=True, check=True
+            )
+
+            # Parse MaxPayload information from the DevCtl section
+            lines = mps_result.stdout.splitlines()
+            devctl_found = False
+            for i, line in enumerate(lines):
+                if "DevCtl:" in line:
```
agirault, line 231: I don't have this for mine:

```
lspci -vv -s 0005:03:00
0005:03:00.0 Ethernet controller: Mellanox Technologies MT2910 Family [ConnectX-7]
        Subsystem: Mellanox Technologies MT2910 Family [ConnectX-7]
        Control: I/O- Mem+ BusMaster+ SpecCycle- MemWINV- VGASnoop- ParErr- Stepping- SERR- FastB2B- DisINTx+
        Status: Cap+ 66MHz- UDF- FastB2B- ParErr- DEVSEL=fast >TAbort- <TAbort- <MAbort- >SERR- <PERR- INTx-
        Latency: 0
        Interrupt: pin A routed to IRQ 195
        IOMMU group: 6
        Region 0: Memory at 2b00000000 (64-bit, prefetchable) [size=32M]
        Expansion ROM at 2e31800000 [virtual] [disabled] [size=1M]
        Capabilities: <access denied>
        Kernel driver in use: mlx5_core
        Kernel modules: mlx5_core

0005:03:00.1 Ethernet controller: Mellanox Technologies MT2910 Family [ConnectX-7]
        Subsystem: Mellanox Technologies MT2910 Family [ConnectX-7]
        Control: I/O- Mem+ BusMaster+ SpecCycle- MemWINV- VGASnoop- ParErr- Stepping- SERR- FastB2B- DisINTx+
        Status: Cap+ 66MHz- UDF- FastB2B- ParErr- DEVSEL=fast >TAbort- <TAbort- <MAbort- >SERR- <PERR- INTx-
        Latency: 0
        Interrupt: pin B routed to IRQ 195
        IOMMU group: 6
        Region 0: Memory at 2b02000000 (64-bit, prefetchable) [size=32M]
        Expansion ROM at 2e31900000 [virtual] [disabled] [size=1M]
        Capabilities: <access denied>
        Kernel driver in use: mlx5_core
        Kernel modules: mlx5_core
```

And obviously I've had no issues running the ANO app. Not sure how we should deal with such errors when we see that it's not actually necessary.

operators/advanced_network/python/check_system.py
cliffburdick: Maybe tune_system.py? I think the setters should stay together since there's a lot of commonality

operators/advanced_network/python/check_system.py
cliffburdick: That value is wrong then. It should be a 4-digit hex number. You will get line rate in most circumstances, but in extreme cases having this larger allows PCIe to be more efficient. Can you paste what it prints?

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,653 @@
...
+
+            # Once all data for a GPU is collected, compare clocks
+            if (
+                current_gpu is not None
+                and sm_current is not None
+                and sm_max is not None
+                and mem_current is not None
+                and mem_max is not None
+            ):
+                logging.info(f"GPU {current_gpu}: Checking clocks...")
```
agirault, line 371:
```suggestion
                logging.debug(f"GPU {current_gpu}: Checking clocks...")
```

operators/advanced_network/python/check_system.py
cliffburdick: I think at this point we've resorted to ANO being for NVIDIA NICs only and it requires the ANO container. That container always contains this script. The script above works on today's card's (maybe?), but it may break on future generations. For example, does it work on BF?

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,653 @@
...
+                sm_current = None
+                sm_max = None
+                mem_current = None
+                mem_max = None
+
+            # Parse current clocks under "Clocks" section
+            elif sm_current is None and line.startswith("SM") and "MHz" in line:
+                sm_current = int(line.split(":")[1].strip().split()[0])
+            elif mem_current is None and line.startswith("Memory") and "MHz" in line:
+                mem_current = int(line.split(":")[1].strip().split()[0])
```
agirault, line 353: I think instead of parsing all this, another cleaner method could be as below:

1. a `get_gpu_info` method using:
   ```bash
   nvidia-smi --query-gpu=index,pci.bus_id --format=csv,noheader
   ```
2. the check clock can then use `nvidia-smi --query-gpu=<field> --format=csv,noheader,no
units` on `clocks.sm`,`clocks.max.sm`,`clocks.mem`, and `clocks.max.mem`

operators/advanced_network/python/check_system.py
cliffburdick: You probably need sudo

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,653 @@
...
+                    logging.info(f"GPU {current_gpu}: SM Clock is correctly set to {sm_max} MHz.")
+
+                if mem_current != mem_max:
+                    logging.warning(
+                        f"GPU {current_gpu}: Memory Clock is set to {mem_current} MHz, but should be {mem_max} MHz."
+                    )
+                else:
+                    logging.info(
+                        f"GPU {current_gpu}: Memory Clock is correctly set to {mem_max} MHz."
+                    )
```
agirault, line 387: Would like to see GPU ID and not just address in logging. would be addressed by doing https://github.com/nvidia-holoscan/holohub/pull/608/files#r1876256759

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,653 @@
...
+
+def check_bar1_size():
+    """
+    Checks the BAR1 size of all NVIDIA GPUs using nvidia-smi.
+    Logs the BAR1 size for each GPU and ensures it is non-zero.
+    """
+    try:
+        # Run nvidia-smi to get BAR1 memory information
+        result = subprocess.run(
+            ["nvidia-smi", "-q", "-d", "MEMORY"], capture_output=True, text=True, check=True
```
agirault, line 374: similarly as clocks above, could we use `--query` for a clearer and better to parse output for bar1 ?

cliffburdick: BAR1 size is not listed as a parameter you can query

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,653 @@
...
+                if "MiB" in bar1_total_str:
+                    bar1_total = int(bar1_total_str.split()[0])
+
+            # Once BAR1 size is found, log it
+            if current_gpu is not None and bar1_total is not None:
+                if bar1_total > 1024:
+                    logging.info(f"GPU {current_gpu}: BAR1 size is {bar1_total} MiB.")
+                else:
+                    logging.warning(
+                        f"GPU {current_gpu}: BAR1 size is {bar1_total} MiB. This may indicate an issue."
```
agirault, line 405: Would be nice to update logs to indicate the threshold, like is done for clocks, MRRS (quantitative) or governor and PIX/PXB (qualitative).

operators/advanced_network/python/check_system.py
agirault: I agree they should stay together. we can revisit on the next round

operators/advanced_network/python/check_system.py
agirault: > we've resorted to ANO being for NVIDIA NICs only and it requires the ANO container. That container always contains this script

That's fair, though one would probably run the script outside the container just locally (which I did for testing), but then not an issue when running the app.

>  The script above works on today's card's (maybe?), but it may break on future generations. For example, does it work on BF?

The command for `interface_name` is going to work always. For the first one though, there probably can be a better heuristic to list the NICs. If we can figure that out that would be nice. `lspci | grep "Ethernet controller"` is safe also, but will grab any (example: the rj45 on the igx too).

That's fine not to block this MR since you pointed out `ibdev2netdev` is in the container. Would be great to think about a more generic way if we can, so lmk if you have an answer to the thoughts above.

operators/advanced_network/python/check_system.py
```diff
@@ -0,0 +1,653 @@
...
+                    mtu_value = int(f.read().strip())
+                    if mtu_value <= 1518:
+                        logging.warning(
+                            f"Interface {iface} has an MTU of {mtu_value} bytes. "
+                            "If possible use larger frame sizes for better performance"
+                        )
+                    else:
+                        logging.info(
+                            f"Interface {iface} has an acceptable MTU of {mtu_value} bytes."
+                        )
```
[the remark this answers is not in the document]
cliffburdick: For MTU size it's really application-dependent. It doesn't help them to increase the size of this if they're sending 1000B frames. It's really intended to say if you can send larger frames, try to.

operators/advanced_network/python/check_system.py
cliffburdick: There's no way to know what the larger BAR1 size is since it varies by GPU

operators/advanced_network/python/check_system.py
agirault: Another root permission issue looks like:

```bash
$ setpci -s 0005:03:00 68.w
ffff
ffff

$ sudo setpci -s 0005:03:00 68.w
2910
2910
```

operators/advanced_network/python/check_system.py
cliffburdick: >The command for interface_name is going to work always.

The command is based on a grep of lspci where the name of the device doesn't change. It would be more correct to do something like look at the PCI vendor ID. We don't want to include other NIC types (like on-board NICs captured by "Ethernet Controller" since those don't have any of the capabilities we're checking for.

operators/advanced_network/python/check_system.py
cliffburdick: I will add a check to make sure they run it with root. There's no way you can get -vv information from lspci that we need without root afaik.

operators/advanced_network/python/check_system.py
```diff
@@ -198,9 +198,9 @@ def check_mrrs():
             mrrs_value = (int(mrrs_result.stdout.strip(), 16) & 0xF000) >> 12
 
             if mrrs_value == 5:
-                logging.info(f"{pci_address}: MRRS is correctly set to 4096.")
+                logging.info(f"{name}/{pci_address}: MRRS is correctly set to 4096.")
```
agirault: gotta loop on the result of `get_nic_info`. `name` is a list too. current:

```
2024-12-09 11:42:46 - WARNING - ['cx7_0', 'cx7_1']/0005:03:00.0: MRRS is set to 4194304, not 4096.
```

operators/advanced_network/python/tune_system.py
```diff
@@ -0,0 +1,625 @@
...
+                                )
+                            break
+                    else:
+                        logging.error(
+                            f"{pci_address}: Unable to find MaxPayload information under DevCtl."
+                        )
+                    break
+
+            if not devctl_found:
+                logging.error(f"{pci_address}: DevCtl section not found.")
```
agirault, line 264: missing `name`

agirault, line 264:
```suggestion
                            f"{name}/{pci_address}: Unable to find MaxPayload information under DevCtl."
                        )
                    break

            if not devctl_found:
                logging.error(f"{name}/{pci_address}: DevCtl section not found.")
```

operators/advanced_network/python/check_system.py
cliffburdick: Thanks. Fixed

operators/advanced_network/python/tune_system.py
```diff
@@ -0,0 +1,629 @@
...
+    return parser.parse_args()
+
+
+def get_nic_info():
+    """
+    Parses the output of `ibdev2netdev -v` to extract and return a list of tuples,
+    where each tuple contains the interface name and its PCIe address.
+
+    Returns:
+        List[Tuple[str, str]]: A list of tuples with (interface_name, pcie_address).
```
agirault, line 100: So that's not correct, looks like it returns `Tuple[List[str])`, specifically two lists. Would have been nicer to do what your doc says though (and I guess what you had meant to do initially), to avoid looping over the `idx`.

operators/advanced_network/python/check_system.py
agirault: no I don't, that works. enforcing root for now works

operators/advanced_network/python/check_system.py
agirault: Nice. Seems like the first command would then be
```bash
lspci -nn | grep -iE 'Ethernet controller.*\[15b3:'
```
Can keep that for another round

operators/advanced_network/python/tune_system.py
```diff
@@ -0,0 +1,629 @@
...
+        sys.exit(1)
+    except subprocess.CalledProcessError as e:
+        logging.error(f"Error while checking Max Payload Size: {e}")
+        sys.exit(1)
+
+
+def check_hugepages():
+    """
+    Checks if hugepages are allocated and ensures that the total allocated hugepage memory
+    is at least 500 MB.
```
agirault, line 281: would we be happy if it's only 64kb size? Thought the size mattered too not just total

operators/advanced_network/python/check_system.py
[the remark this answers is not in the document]
agirault: I hear you, but then the test itself should always be application-dependent, in which case the `1518` is not right. If we're going to put that value as a check, I think we should list it in the log.

A lot of the other checks are also application dependent, for example I only have a fraction of them set but the benchmark app is hitting line rate, so I think it's still relevant to explain the heuristic chosen for warning vs info.

Improvements I see for the future:
1. we provide more in-depth information in those logs (why/when that matters)
2. we make checks dynamic based on application config to only display what is relevant

operators/advanced_network/python/check_system.py
agirault: 😭

agirault:
Needs a little `python3 -m black .` and some signing of the commits.

operators/advanced_network/python/tune_system.py
cliffburdick: That's true. Like the other things, the answer is that it depends. Specifically some systems may only want to use certain sizes (maybe they already use it for another application), and they don't want the bigger ones. I'm ok with just saying "hugepages are enabled/used" as the metric rather than trying to validate a certain size.

agirault: Here is my thought:
1. when running the script by itself outside the context of applications, we should either
  a. have arbitrary rules, or
  b. don't do it at all (here just checking hugepages are created is fine, for other stuff like max payload size, mrrs, mtu etc just skip)
2. when running at startup in the app, we should check what specific config is done, and that might help augment these checks with more specific heuristics. And if users still don't like them, it's just warnings not errors, and they can ignore them, or we can have a flag to skip them.

I just would like to be consistent. IMO I'd rather ask "too much" in the vanilla run of the script to recommend optimal settings, since you could argue that every single one of these checks also dependson the need (you might be fine even without a lot of these settings), but maybe that's not reasonable?

operators/advanced_network/python/tune_system.py
agirault: want to switch to a `List[Tuple[str, str]]` and simplify the for loops, or just update that description for now?

operators/advanced_network/python/tune_system.py
cliffburdick: >for other stuff like max payload size, mrrs, mtu etc just skip

I didn't follow this part. Why would we not want to warn if their system is not set up optimally? Hugepages can be considered some level of optimality if any size is enabled and not just the largest.

> I just would like to be consistent. IMO I'd rather ask "too much" in the vanilla run of the script to recommend optimal settings, since you could argue that every single one of these checks also dependson the need (you might be fine even without a lot of these settings), but maybe that's not reasonable?

This is what Rivermax does. When you start it you typically get a whole bunch of warnings about things you should update. I don't see any issues warning them even if they never plan to do it since it seems like being more aggressive is better.

operators/advanced_network/python/tune_system.py
cliffburdick: Fixed

operators/advanced_network/python/check_system.py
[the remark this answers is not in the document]
cliffburdick: I personally think a warning that you ignore, but reminds you that using jumboframes is more efficient than not is better than not saying anything and having the user think the small sizes are fine. You might be hitting line rate with 1518B packets on your system, but other systems might require more cores/queues to hit the same thing.

operators/advanced_network/python/tune_system.py
agirault: no that sounds great to me, we're on the same page there, I was just thinking that we could do the same with huge pages, and say the optimal one is probably large huge pages than 64k (especially since you have larger sizes available like 1gb here) and if they don't agree they can ignore it.

operators/advanced_network/python/check_system.py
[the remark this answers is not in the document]
agirault: let me rephrase: the test right now might tell you "you're good" or "you're not good" and it's not clear why to users. I'd like to see the threshold we use logged, like for the other checks, even if it's not a perfect heuristic and is application dependent.

And if in the future we can make that threshold value more adequate based on specific app configs, great.

operators/advanced_network/python/tune_system.py
cliffburdick: We could do that. Detecting hugepage sizes when they're not enabled is a real pain from what I recall. You have to check CPU flags and it's not the same from ARM to x86. When they're enabled you can just check the sys fs and walk through all the sizes of the directories

operators/advanced_network/python/check_system.py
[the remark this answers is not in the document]
cliffburdick: I can update the logs to specify the size we're looking for

operators/advanced_network/python/tune_system.py
cliffburdick: Can we add that as a later step to get this merged?

agirault (approved):
great start, good to merge as-is, I have suggestions for improvements and actionability I can do later

operators/advanced_network/python/tune_system.py
agirault: sure. and I meant when enabled already, so just making a check on the page size enabled (and not just the sum of them)

PR #615, opened 2024-12-05
Update PyTorch Install for VILA_Live and Florence 2

NigelNelson:
With the upgrade to HSDK 2.6 using CUDA 12.6, building the VILA_Live and Florence-2 Dockerfiles resulted in errors during the building of Flash-attn 2. This is because Flash-attn 2 uses PyTorch during the build process, and the installed PyTorch used an older version of CUDA than the local 12.6. This MR simplifies the PyTorch install to use the public indexurl. It has been tested for x86 w/ dGPU & IGX w/ dGPU, but iGPU still needs to be verified.

### DIFF
--- a/applications/florence-2-vision/Dockerfile
+++ b/applications/florence-2-vision/Dockerfile
@@ -19,11 +19,6 @@ ARG BASE_IMAGE
 ARG GPU_TYPE
 ARG COMPUTE_CAPACITY
 ############################################################
-# Torch image (Used to extract the ARM64 dGPU pytorch wheel)
-############################################################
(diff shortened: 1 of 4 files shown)


tbirdso (approved):
Great! Nice improvement over pulling the PyTorch container

tbirdso:
>iGPU still needs to be verified.

@NigelNelson  is this a blocker for merge?

NigelNelson:
> > iGPU still needs to be verified.
> 
> @NigelNelson is this a blocker for merge?

No it is not

PR #616, opened 2024-12-05
Laser Detection Latency Application

jjomier:
Add laser detection app to demonstrate the latency between two different camera sources.


tbirdso (approved):
Looks good! Thanks @jjomier @anishag14

PR #618, opened 2024-12-06
Missing images for laser latency application

jjomier:


tbirdso (approved):
Do we also need to update `.gitignore`?

PR #620, opened 2024-12-06
Loopback mode

cliffburdick:


applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -184,24 +184,14 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
...
-    // NOTE: we can't actually ignore all standard linux packets on a real network (with a switch),
-    //       at least ARP packets should be processed, or delegate to linux for standard traffic.
-    if (adv_net_get_q_id(burst) == 0) {
-      adv_net_free_all_pkts_and_burst(burst);
-      return;
-    }
-
-    // Count packets received
-    ttl_pkts_recv_ += burst_size;
-
```
agirault, line 198: I don't think we should remove this without removing the queue 0 from other config files. Maybe out of scope of this MR? Would like to at least separate it in different commits so that anyone can come back to history and understand why that was done

agirault:
I know this is draft, but curious already: how does this differ from the current loopback we're able to do? It does loopback on a single interface? Why are we looking for that?

cliffburdick:
> I know this is draft, but curious already: how does this differ from the current loopback we're able to do? It does loopback on a single interface? Why are we looking for that?

What we have now is not considered loopback. Loopback is a "local only" path for a packet, whereas right now the packet actually leaves the NIC and is sent on the wire before coming back. This new mode is useful if you don't have a NIC.

applications/adv_networking_bench/cpp/default_bench_op_rx.h
cliffburdick: This should be removed and we should add the flow_isolation parameter into each of them so it's not confusing to users to have that code.

applications/adv_networking_bench/adv_networking_bench_default_tx_rx_loopback.yaml
```diff
@@ -0,0 +1,96 @@
...
+  stop_on_deadlock_timeout: 500
+
+advanced_network:
+  cfg:
+    version: 1
+    manager: "dpdk"
+    master_core: 3
+    debug: false
+    log_level: "info"
+    loopback: true
```
agirault, line 35: `sw_loopback`?

applications/adv_networking_bench/cpp/default_bench_op_rx.h
agirault: agreed. or default `flow_isolation` to true. I believe we have jira tickets for all this, are you ok doing this in a separate MR?

operators/advanced_network/adv_network_types.h
```diff
@@ -369,6 +369,7 @@ struct CommonConfig {
   int master_core_;
   AdvNetDirection dir;
   AnoMgrType manager_type;
+  bool loopback_;
```
agirault, line 372: `sw_loopback_` ?

operators/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
```diff
@@ -49,9 +50,12 @@ struct RxWorkerParams {
...
   uint32_t batch_size;
   struct rte_ring* ring;
+  struct rte_ring* lb_ring;
   struct rte_mempool* flowid_pool;
-  struct rte_mempool* burst_pool;
-  struct rte_mempool* meta_pool;
+  struct rte_mempool* rx_burst_pool;
+  struct rte_mempool* tx_burst_pool;
+  struct rte_mempool* rx_meta_pool;  
+  struct rte_mempool* tx_meta_pool;
```
agirault, line 58: sorry if this is too early as you're clearly still in draft, but if you can eventually add an inline comment for all of them for a short description of their role that'd be great, especially noting that so many of them are exclusively for SW loopback which isn't relevant once running properly.

applications/adv_networking_bench/adv_networking_bench_default_tx_rx_loopback.yaml
cliffburdick: That would make the feature more complicated because we'd have to differentiate between which interfaces are/aren't used in loopback, and allocate pools differently. I'm not sure we wanted that complexity yet...

operators/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
```diff
@@ -417,17 +436,23 @@ void DpdkMgr::initialize() {
   int max_rx_batch_size = 0;
```
dylan-eustice, line 436: Don't map memory regions if running in SW loopback mode, `rte_eth_dev` calls will fail.
```suggestion
  if (!loopback_) {
    if (map_mrs() != AdvNetStatus::SUCCESS) {
      HOLOSCAN_LOG_CRITICAL("Failed to map MRs");
      return;
    }
  }
```

cliffburdick: Weird I didn't see a problem here. Was it showing up for you?

applications/adv_networking_bench/adv_networking_bench_default_tx_rx_loopback.yaml
agirault: sounds good to me, top level it is. can we still nme it `sw_loopback` ?

Also, if we set this to `sw_loopback`, do we need the address on the interfaces to indicate loopback as well?

operators/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
```diff
@@ -49,9 +50,12 @@ struct RxWorkerParams {
...
   uint32_t batch_size;
   struct rte_ring* ring;
+  struct rte_ring* lb_ring;          // Ring used between TX and RX when in SW loopback mode
   struct rte_mempool* flowid_pool;
-  struct rte_mempool* burst_pool;
-  struct rte_mempool* meta_pool;
+  struct rte_mempool* rx_burst_pool; // Pool used to pull out bursts from RX pool
+  struct rte_mempool* tx_burst_pool; // Pool used for loopback mode to return transmitted bursts
+  struct rte_mempool* rx_meta_pool;  // Pool used for RX metadata structures
+  struct rte_mempool* tx_meta_pool;  // Pool used in loopback for returning transmitted metadata
```
agirault, line 58: suggestive nitpick for consideration, fine otherwise
```suggestion
  struct rte_mempool* burst_pool;        // Pool used to pull out bursts from RX pool
  struct rte_mempool* meta_pool;         // Pool used for RX metadata structures
  struct rte_mempool* sw_lb_burst_pool_; // Pool used in sw loopback mode to return transmitted bursts
  struct rte_mempool* sw_lp_meta_pool_;  // Pool used in sw loopback mode to return transmitted metadata
```

operators/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
```diff
@@ -49,9 +50,12 @@ struct RxWorkerParams {
   int num_segs;
   uint32_t batch_size;
   struct rte_ring* ring;
+  struct rte_ring* lb_ring;          // Ring used between TX and RX when in SW loopback mode
```
agirault, line 53:
```suggestion
  struct rte_ring* sw_lb_ring;          // Ring used between TX and RX when in SW loopback mode
```

operators/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
dylan-eustice: Yes, on my setup with no NIC, I get the following error:
```
Invalid port_id=0
[2024-12-10 20:32:46.176] [623] [holoscan] [critical] [adv_network_dpdk_mgr.cpp:164] Failed to get device info for port 0
```

operators/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
dylan-eustice, line 1: For some reason, I'm still not seeing `rte_pktmbuf_free(reinterpret_cast<rte_mbuf**>(burst->pkts[0])[p]);` free all segments of a packet when running in loopback, even though I can see the chaining code being executed in the loopback worker.

On my end, I fixed this by adding. 
```
    rte_pktmbuf_free(reinterpret_cast<rte_mbuf**>(burst->pkts[0])[p]);
 +  if (loopback_) {
 +     for (int seg = 1; seg < burst->hdr.hdr.num_segs; seg++) {
 +       rte_pktmbuf_free(reinterpret_cast<rte_mbuf**>(burst->pkts[seg])[p]);
 +     }
 +   }
```

Not sure what the root cause could be.

cliffburdick: That is strange. That function definitely should walk the whole list. Can you print:

`burst->pkts[0])[p]->next`

cliffburdick: > It seems like each burst received from the ANO's RX output port is the same pointer

In the current code it's a shared_ptr that's created every time a burst is made, but I'm changing that in PR #630 . I don't know why you were seeing a duplicate pointer...

dylan-eustice: I was confused too, it seemed like the pointer should change. But the underlying pointer was the same value each time it was logged in `AdvNetworkOpTx::compute`.

operators/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
```diff
@@ -330,8 +330,8 @@ void DpdkMgr::initialize() {
           },
   };
   
-  loopback_ = cfg_.common_.loopback_;
-  if (loopback_) {
+  loopback_ = cfg_.common_.loopback_ ? LoopbackType::LOOPBACK_TYPE_SW : LoopbackType::DISABLED;
```
dylan-eustice, line 333: Was this enum defined in a header file that didn't get pushed?

operators/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
cliffburdick: This may be used for other types of loopback, so leaving the name for now.

operators/advanced_network/adv_network_types.h
cliffburdick: Changed to enum type

applications/adv_networking_bench/adv_networking_bench_default_tx_rx_loopback.yaml
cliffburdick: Changed it to a string that can be "sw" or "disabled" (default)

PR #621, opened 2024-12-06
Enables nsys in iGPU container

sohamm17:


AndreasHeumann:
@sohamm17 Instead of installing NSight for iGPU could we add an option which allows the user to map any local NSight installation into the dev container? Something like '--nsys_locattion=/opt/nvidia/nsight-systems/2024.2.2'. Also print a message indicating to use that option when using `./run ... --nsys_profile` and nsys is not found?
This would be more flexible (can use other NSight version for dGPU as well) and would also avoid hardcoding the `2024.2.2` version. The `nsys` executable would also need to be mapped to `/usr/local/bin` to avoid having the path in the `run` script (or use the `update-alternatives` command).

sohamm17:
> @sohamm17 Instead of installing NSight for iGPU could we add an option which allows the user to map any local NSight installation into the dev container? Something like '--nsys_locattion=/opt/nvidia/nsight-systems/2024.2.2'. Also print a message indicating to use that option when using `./run ... --nsys_profile` and nsys is not found? This would be more flexible (can use other NSight version for dGPU as well) and would also avoid hardcoding the `2024.2.2` version. The `nsys` executable would also need to be mapped to `/usr/local/bin` to avoid having the path in the `run` script (or use the `update-alternatives` command).

Great suggestion @AndreasHeumann . Let me make the changes.

sohamm17:
> > @sohamm17 Instead of installing NSight for iGPU could we add an option which allows the user to map any local NSight installation into the dev container? Something like '--nsys_locattion=/opt/nvidia/nsight-systems/2024.2.2'. Also print a message indicating to use that option when using `./run ... --nsys_profile` and nsys is not found? This would be more flexible (can use other NSight version for dGPU as well) and would also avoid hardcoding the `2024.2.2` version. The `nsys` executable would also need to be mapped to `/usr/local/bin` to avoid having the path in the `run` script (or use the `update-alternatives` command).
> 
> Great suggestion @AndreasHeumann . Let me make the changes.

@AndreasHeumann Kindly review.

AndreasHeumann (approved):
Looks good to me, thanks!

tbirdso (approved):
LGTM

PR #629, opened 2024-12-10
Remove unnecessary find_project

wendell-hom:
The `find_project` in the laser_detection_latency app is picked up when building any application since we are using add subdirectory to include the laser detection latency app with: `add_subdirectory(laser_detection_latency)`
 
Other applications should not be constrained by the Holoscan SDK version required by this app.


tbirdso:
Thanks @wendell-hom , good spot.

tbirdso (approved):

PR #632, opened 2024-12-11
Add psd_pipeline application

JohnMoon-VTS:
Add a new ANO-driven signal processing pipeline that ingests high-rate VITA 49 IQ data and computes/packetizes averaged, lower-rate, VITA 49.2 spectral PSDs. A few operators that may be reused are added under the common "operators" directory as well.

Note: build will require #627 to merge.


dylan-eustice:
So this has implementations of both sending and receiving VITA 49 flows with the ANO? Very cool.

JohnMoon-VTS:
The VITA 49 RX is done with ANO, yes. But the VITA 49 PSDs being sent out the other side of the pipeline are being packetized and sent by the [`vita49_psd_packetizer` operator](https://github.com/JohnMoon-VTS/holohub/tree/ano-psd-pipeline/operators/vita49_psd_packetizer) (in the CPU). These PSD packets are low rate and we're using a third-party library for the packetization that didn't play nice with `nvcc`, so this seemed like the path of least resistance.

If you think it'd make more sense to send them out via ANO as well, we're happy to consider it!

operators/vita49_psd_packetizer/context.cpp
```diff
@@ -0,0 +1,1133 @@
+/**
+ * SPDX-FileCopyrightText: 2024 Geon Technologies, LLC
+ * SPDX-License-Identifier: LGPL-3.0-only
+ * Originally generated by vrtgen: https://github.com/Geontech/vrtgen
+ */
```
tbirdso, line 5: I'm concerned about providing and compiling LGPL source code as part of this project. LGPL is typically not compatible with the default Apache license used in HoloHub projects unless dynamically linked. In this case `context.cpp` is directly compiled as part of the PSD application, meaning any derived PSD application components likely would need to be LGPL licensed as well.

Some possible options:
- Does `vrtgen` maintain a package that could be dynamically linked?
- Could `vrtgen` generate files under a permissive license? You might reach out to the `vrtgen` maintainers to understand options here.
- We might accept a HoloHub contribution if all components are properly licensed for compatibility with LGPL components, but will require further discussion. cc @jjomier for a second opinion.

operators/vita49_psd_packetizer/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+  GIT_REPOSITORY https://github.com/NVIDIA/MatX.git
+  GIT_TAG v0.9.0
+)
+FetchContent_MakeAvailable(MatX)
+
+add_library(packetlib
+    packet_sender.cpp
+    signal_data.cpp
+    context.cpp
+)
```
tbirdso, line 32: Note that the LGPL `context.cpp` component is typically incompatible with other Apache-2.0 sources.

tbirdso (changes requested):
Thanks @JohnMoon-VTS for your project submission! I've added a few high level comments, with one potential blocker regarding licensing.

@dylan-eustice would you mind taking a review pass as you're more familiar with the domain?

operators/vita49_psd_packetizer/context.cpp
tbirdso: After offline discussion with @jjomier we'd like to avoid committing LGPL code to HoloHub to minimize risk of exposure to other projects.

_If it is not possible to use permissive licensing by updating `vrtgen` or using a different library_, then we'll request the following app changes:
- Instead of committing LGPL code, fetch `vrtgen` dynamically and generate context sources at config/build time. This might be accomplished with CMake's `FetchContent`.
- Instead of compiling LGPL code with Apache-2.0 code, isolate `vrtgen` generated code in a dedicated CMake shared library target ([doc](https://cmake.org/cmake/help/latest/command/add_library.html#id2)). Then, have other app components dynamically link against that target.
- Explicitly document in the operator and app READMEs that LGPL code is included via `vrtgen`.
- Sign-off on the commit (`git commit -s`) to attest that licenses are properly respected. (We always require this for HoloHub contributions)

operators/data_writer/data_writer.cu
```diff
@@ -0,0 +1,51 @@
...
+    holoscan::Operator::initialize();
+    make_tensor(data_host, {num_bursts.get(), burst_size.get()}, MATX_HOST_MEMORY);
+}
+
+void DataWriter::compute(InputContext& op_input, OutputContext& op_output, ExecutionContext&) {
+    auto data = op_input.receive<in_t>("in").value();
+    copy(data_host, std::get<0>(data));
+
+    // TODO: the MatX write_csv function segfaults for some reason
+    // io::write_csv(psd_host, "psdOut.csv", ",");
```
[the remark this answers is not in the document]
JohnMoon-VTS: I don't believe so - we've talked with @cliffburdick offline about this and he was able to reproduce, but I don't see a GitHub issue for it.

operators/data_writer/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Overview
+
+Writes binary data from its input to an output file. This operator is intened
+for use as a debugging aid.
+
+## Description
+
+The data writer operator takes in a `std::tuple<tensor_t<complex, 2>, cuda_stream_t>`,
+copies the data to a host tensor, then writes the data out to a binary file.
```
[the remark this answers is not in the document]
JohnMoon-VTS: Okay, will do.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -0,0 +1,383 @@
...
+      auto channel = channel_list.at(channel_num);
+      channel->current_context.context_changed = context_changed_h(ctxt);
+      channel->current_context.bandwidth_hz = get_bandwidth_hz_h(ctxt);
+      channel->current_context.rf_ref_freq_hz = get_rf_ref_freq_hz_h(ctxt);
+      channel->current_context.reference_level_dbm = get_ref_level_dbm_h(ctxt);
+      channel->current_context.gain_stage_1_db = get_gain_1_db_h(ctxt);
+      channel->current_context.gain_stage_2_db = get_gain_2_db_h(ctxt);
+      channel->current_context.sample_rate_sps = get_sample_rate_sps_h(ctxt);
+      channel->context_received = true;
+      // TODO: when context changes, we should flush data
```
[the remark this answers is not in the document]
JohnMoon-VTS: Yes, we intentionally left that as a TODO for now. There's some non-trivial logic that needs to be added to handle this case and it's fairly minor for us at this stage.

applications/psd_pipeline/README.md
```diff
@@ -0,0 +1,185 @@
+<!--
+SPDX-FileCopyrightText: 2024 Valley Tech Systems, Inc.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+# PSD Pipeline
+
+## Overview
```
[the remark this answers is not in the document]
JohnMoon-VTS: Yes, will do!

applications/psd_pipeline/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+# SPDX-FileCopyrightText: 2024 Valley Tech Systems, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+cmake_minimum_required(VERSION 3.20)
+project(psd_pipeline CXX CUDA)
+
+find_package(holoscan 2.5.0 REQUIRED CONFIG
+             PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")
+
+set(CMAKE_CUDA_ARCHITECTURES "70;80;90")
```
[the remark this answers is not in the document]
JohnMoon-VTS: We do require portability for this application. We're moving the built container to an embedded system.

operators/vita49_psd_packetizer/context.cpp
JohnMoon-VTS: I understand why you're hesitant to let LGPL code into HoloHub. We'll work a solution on our end (maybe just doing what you suggest above).

operators/data_writer/data_writer.cu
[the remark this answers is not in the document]
cliffburdick: It's something about mixing the python version that's used in matx and holohub. I didn't have time to debug more, but this function works fine in a standalone environment. When it's in holohub there seems to be some paths and symbols that are mixed up.

JohnMoon-VTS:
@tbirdso, I think this is ready for re-review. We moved out the LGPL code to a repo on our end that gets pulled in at build-time.

tbirdso (approved):
Thanks @JohnMoon-VTS for the updates and for moving the LGPL sources to a dedicated repo. You may want to consider making the `vita49_psd` [external target](https://github.com/vts-i/vita49-psd/blob/main/CMakeLists.txt#L15) a shared rather than static library if you are distributing binaries at any time in reflection of LGPL compliance. (_not legal advice_)

Updates look good to me, PR needs approval from an ANO stakeholder. Let me follow up.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.h
```diff
@@ -0,0 +1,321 @@
...
+    uint32_t v = bswap_32_h(*((uint32_t*)&val));
+    return *((float*)&v);
+}
+
+inline __device__ float bswap_32(float val) {
+    uint32_t v = bswap_32(*((uint32_t*)&val));
+    return *((float*)&v);
+}
+
+inline uint64_t bswap_64_h(uint64_t val) {
```
cliffburdick, line 118: For future reference, this should be used on the device: https://docs.nvidia.com/cuda/cuda-math-api/cuda_math_api/group__CUDA__MATH__INTRINSIC__INT.html?highlight=__byte_perm#_CPPv411__byte_permjjj

It's likely not faster for the 16-bit endian swap, but it will be for 32/64.

operators/fft/fft.cu
```diff
@@ -0,0 +1,128 @@
...
+    make_tensor(outputs,
+                {num_channels.get(), num_bursts.get(), burst_size.get()},
+                MATX_DEVICE_MEMORY);
+}
+
+void FFT::compute(InputContext& op_input, OutputContext& op_output, ExecutionContext& context) {
+    auto input = op_input.receive<in_t>("in").value();
+    auto meta = metadata();
+    auto channel_num = meta->get<uint16_t>("channel_number", 0);
+    auto out = outputs.Slice<2>({static_cast<index_t>(channel_num), 0, 0},
```
cliffburdick, line 91: same comment on using `slice()` instead of the member function

operators/high_rate_psd/high_rate_psd.cu
```diff
@@ -0,0 +1,51 @@
...
+                {num_channels.get(), num_bursts.get(), burst_size.get()},
+                MATX_DEVICE_MEMORY);
+    scale_factor = 1.0 / pow(burst_size.get(), 2);
+}
+
+void HighRatePSD::compute(InputContext& op_input, OutputContext& op_output, ExecutionContext&) {
+    auto input = op_input.receive<in_t>("in").value();
+    auto meta = metadata();
+    auto channel_num = meta->get<uint16_t>("channel_number", 0);
+    auto out = outputs.Slice<2>({static_cast<index_t>(channel_num), 0, 0},
```
cliffburdick, line 40: `slice()`

operators/low_rate_psd/low_rate_psd.cu
```diff
@@ -0,0 +1,54 @@
...
+        "Number of channels",
+        "Number of channels to allocate memory for");
+}
+
+void LowRatePSD::initialize() {
+    holoscan::Operator::initialize();
+    make_tensor(outputs, {num_channels.get(), burst_size.get()}, MATX_DEVICE_MEMORY);
+    make_tensor(maxima, {burst_size.get()}, MATX_DEVICE_MEMORY);
+    make_tensor(minima, {burst_size.get()}, MATX_DEVICE_MEMORY);
+    (maxima = ones() * 127.0).run();
```
cliffburdick, line 33: Isn't this the same as `(maximum = 127.0).run();`?

operators/low_rate_psd/low_rate_psd.cu
```diff
@@ -0,0 +1,54 @@
...
+    (minima = ones() * -128.0).run();
+}
+
+void LowRatePSD::compute(InputContext& op_input, OutputContext& op_output, ExecutionContext&) {
+    auto input = op_input.receive<in_t>("in").value();
+    auto meta = metadata();
+    auto channel_num = meta->get<uint16_t>("channel_number", 0);
+    auto out = outputs.Slice<1>({static_cast<index_t>(channel_num), 0},
```
cliffburdick, line 41: `slice()`

cliffburdick (commented):
This looks great! I think it showcases many different tools together very well. How easy is it for a user other than you to run?

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -0,0 +1,383 @@
...
+void Vita49ConnectorOpRx::free_bufs_and_emit_arrays(
+        OutputContext& op_output,
+        std::shared_ptr<struct Channel> channel) {
+  std::vector<Vita49ConnectorOpRx::RxMsg> completed_msgs = free_bufs(channel);
+  if (completed_msgs.empty()) {
+      HOLOSCAN_LOG_INFO("no completed messages yet");
+      return;
+  }
+
+  auto meta = metadata();
```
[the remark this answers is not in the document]
JohnMoon-VTS: Yeah, that was us. By this stage, it appears to be low enough overhead that it's not causing an issue for us. But yeah, in the future, any more perf we could get out of the metadata passing would be welcome!

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -0,0 +1,383 @@
...
+  meta->set("change_indicator", channel->current_context.context_changed);
+  meta->set("bandwidth_hz", channel->current_context.bandwidth_hz);
+  meta->set("rf_ref_freq_hz", channel->current_context.rf_ref_freq_hz);
+  meta->set("reference_level_dbm", channel->current_context.reference_level_dbm);
+  meta->set("gain_stage_1_db", channel->current_context.gain_stage_1_db);
+  meta->set("gain_stage_2_db", channel->current_context.gain_stage_2_db);
+  meta->set("sample_rate_hz", channel->current_context.sample_rate_sps);
+
+  cudaStream_t stream = completed_msgs[0].stream;
+  auto data = channel->rf_data.Slice<2>({static_cast<index_t>(channel->cur_idx), 0, 0},
```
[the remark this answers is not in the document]
JohnMoon-VTS: Understood - will update!

applications/psd_pipeline/advanced_network_connectors/vita49_rx.h
JohnMoon-VTS: Okay, I'll try out using the native CUDA functions for the swap. We ended up just using these since we knew the logic worked correctly from REDHAWK experience.

operators/low_rate_psd/low_rate_psd.cu
JohnMoon-VTS: Probably, yeah! We knew this initialization looked silly, but I couldn't find anything in the MatX docs for initializing a tensor to all one value like this. Will update.

JohnMoon-VTS:
Thanks for the detailed review, @cliffburdick! I'll incorporate your feedback as soon as I can.

> How easy is it for a user other than you to run?

It should be pretty easy for anyone who has a VITA 49 data source of some kind. Some parameters would need to be tweaked for the particular format (e.g. how many samples come in a packet) and particular hardware (e.g. memory sizes and PCIe addresses), but those should be called out in the README.

JohnMoon-VTS:
@cliffburdick, I've made the changes you've requested and rebased on top of 2.9.0. I think this is ready for re-review. Thanks!

tbirdso:
Approved for merge in offline discussion with @cliffburdick . Thank you @JohnMoon-VTS for making the requested changes.

PR #634, opened 2024-12-11
Add HOLOHUB_WHEEL_TESTING option

wendell-hom:
Update holohub applications that are tested by nightly CI for python wheels flow.

When this option is enabled:
* no need to the holoscan package with `find_package`
* don't build cpp application and it's operators

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,7 @@ include(CTest)
 # Options
 option(HOLOHUB_DOWNLOAD_DATASETS "Download datasets" ON)
 option(HOLOHUB_BUILD_PYTHON "Build Holoscan SDK Python Bindings" ON)
+option(HOLOHUB_WHEEL_TESTING "Build Applications for Python Wheel Testing" OFF)
 
 # Enable flow benchmarking
 option(FLOW_BENCHMARKING "Enable Flow Benchmarking" OFF)
--- a/applications/endoscopy_tool_tracking/CMakeLists.txt
+++ b/applications/endoscopy_tool_tracking/CMakeLists.txt
@@ -36,7 +36,7 @@ if(HOLOHUB_DOWNLOAD_DATASETS)
   )
 endif()
 
-if (NOT HOLOHUB_WHEEL_TESTING)
+if(NOT HOLOHUB_WHEEL_TESTING)
     add_subdirectory(cpp)
 endif()
 add_subdirectory(python)
--- a/applications/endoscopy_tool_tracking/CMakeLists.txt
+++ b/applications/endoscopy_tool_tracking/CMakeLists.txt
@@ -36,5 +36,7 @@ if(HOLOHUB_DOWNLOAD_DATASETS)
   )
 endif()
 
-add_subdirectory(cpp)
+if (NOT HOLOHUB_WHEEL_TESTING)
+    add_subdirectory(cpp)
+endif()
 add_subdirectory(python)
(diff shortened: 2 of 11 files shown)


CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ include(CTest)
 # Options
 option(HOLOHUB_DOWNLOAD_DATASETS "Download datasets" ON)
 option(HOLOHUB_BUILD_PYTHON "Build Holoscan SDK Python Bindings" ON)
+option(HOLOHUB_WHEEL_TESTING "Build Applications for Python Wheel Testing" OFF)
```
tbirdso, line 38: Recommend that we consider support for a `HOLOHUB_BUILD_CPP_APPS` option. May need to consider how this impacts Python projects relying on C++ operator/extension builds.

Note that the `HOLOHUB_BUILD_PYTHON` option exists but might not be uniformly applied across projects right now.

Alternatively, maybe `HOLOHUB_BUILD_PYTHON_ONLY` would be clearer in explicitly disallowing C++ build paths? What do you think?

```suggestion
option(HOLOHUB_BUILD_PYTHON "Build Holoscan SDK Python Bindings" ON)
option(HOLOHUB_BUILD_CPP_APPS "Build HoloHub C++ applications" OFF)
```

applications/multiai_ultrasound/CMakeLists.txt
```diff
@@ -36,8 +36,10 @@ if(HOLOHUB_DOWNLOAD_DATASETS)
   )
 endif()
 
-add_subdirectory(operators)
-add_subdirectory(cpp)
+if(NOT HOLOHUB_WHEEL_TESTING)
+    add_subdirectory(operators)
+    add_subdirectory(cpp)
+endif()
```
tbirdso, line 42: Seems like this would break `multiai_ultrasound` in Python testing. `multiai_ultrasound` depends on Python-wrapped C++ operators. Based on that it doesn't seem like `multiai_ultrasound` is a valid candidate for Python wheel test coverage.

applications/endoscopy_tool_tracking/CMakeLists.txt
```diff
@@ -36,5 +36,7 @@ if(HOLOHUB_DOWNLOAD_DATASETS)
   )
 endif()
 
-add_subdirectory(cpp)
+if(NOT HOLOHUB_WHEEL_TESTING)
+    add_subdirectory(cpp)
+endif()
```
tbirdso, line 41: Control flow for core subdirectories should not depend on a TESTING variable. Need to consider rename and potential conditions that could invalidate the condition, such as in the case of `multiai_ultrasound` depending on C++ operators.

tbirdso (changes requested):
Hi @wendell-hom , thanks for looking into this.

From my current reading of this PR, I think we should take another look at the approach of how Holohub application language subdirectories are enabled and what we are trying to test. As I understand it, this PR is trying to solve the following problem:
1. We want to test HoloHub Python applications with only a Holoscan wheel. No C++ or CMake sources are available via deb or container environment.
2. Many HoloHub projects eagerly search for a Holoscan CMake config with `find_package(holoscan)`
3. Many HoloHub projects eagerly include C++ applications with `add_subdirectory(cpp)`

As you've demonstrated in this PR, the issue at hand is that HoloHub projects do _not_ uniformly support Python-only builds. We should consider addressing that root problem before we try to add test coverage support.

To move forward I'd like to better understand:
1. What are the requirements for HoloHub wheel support that we're attempting to cover with testing? Is there some reason these applications in particular require wheel-only support?
2. Does the current proposed approach work in testing? How are C++ operators (LSTM Inference, Visualizer iCardio) obtained when they are disabled for wheel-only builds?
3. Are there best practices that we could establish and/or enforce for HoloHub projects to ensure wheel compatibility, before adding coverage?

cc @jjomier for comments on background and HoloHub infrastructure design

PR #635, opened 2024-12-12
Remove shared pointer interface

cliffburdick:


ronyrad:
Looks good to me

agirault:
History seems a bit wacky. Can you rebase and clean this up to unblock merge? Ty

agirault:
Also @cliffburdick , could you provide the results of `ctest . --verbose` in the build directory with no other change to confirm this does not break the test? Thank you!

cliffburdick:
> ctest . --verbose

I'm getting 47Gbps and it's failing, but I was only getting 47Gbps on my setup before that. Will look into it.

tbirdso (approved):
Approved per offline discussion

PR #636, opened 2024-12-12
Fixed wrong URLs

jjomier:
Fixing URLs not pointing to the right target


tbirdso:
FYI @anishag14 @mocsharp

tbirdso (approved):

PR #637, opened 2024-12-12
First iteration of packaging for holoscan-networking

agirault:
## Changes

- add `install` rule for networking-related cmake target (+ `COMPONENTS`, though not necessary if not passing a specific `COMPONENT` to the deb configure method - see docs below)
- create cmake method to configure debian packages `holohub_configure_deb` and apply to `holoscan-networking`
- create cmake method to configure build tree at a package level with `-D PKG_<name>=1`
- add docs in `CONTRIBUTING` guide

> _**Note**: will upgrade `run` and `dev_container` infra for supporting packaging in a future MR_

## Results

- docs: https://github.com/nvidia-holoscan/holohub/blob/ag/ano-cpack/CONTRIBUTING.md#adding-a-packager

- build:
```bash
cmake -S . -B build/holoscan-networking -D PKG_holoscan-networking
cmake --build build/holoscan-networking -j
cpack --config build/holoscan-networking/pkg/CPackConfig-holoscan-networking.cmake
```
```
CPack: Create package using DEB
CPack: Install projects
CPack: - Run preinstall target for: Holohub
CPack: - Install project: Holohub []
CPack: Create package
CPack: - package: /workspace/holohub/holoscan-networking_0.1.0_arm64.deb generated.
``` 

- inspect:

```bash
dpkg -I build/adv_networking_bench/holoscan-networking_0.1.0_arm64.deb
```
```
new Debian package, version 2.0.
 size 2324324 bytes: control archive=914 bytes.
     288 bytes,    10 lines      control              
    1696 bytes,    18 lines      md5sums              
 Architecture: arm64
 Depends: holoscan (>= 2.6), mlnx-dpdk, doca-sdk-gpunetio, doca-sdk-flow, doca-sdk-eth
 Description: Holoscan Networking
 Maintainer: Alexis Girault <agirault@nvidia.com>
 Package: holoscan-networking
 Priority: optional
 Section: devel
 Version: 0.1.0
 Installed-Size: 5299
```

```sh
dpkg --contents build/holoscan-networking/holoscan-networking_0.1.0_arm64.deb | awk '!/\/$/ {print $6}'
```
```
./usr/bin/adv_networking_bench/cpp/adv_networking_bench
./usr/bin/adv_networking_bench/cpp/adv_networking_bench_default_rx.yaml
./usr/bin/adv_networking_bench/cpp/adv_networking_bench_default_tx.yaml
./usr/bin/adv_networking_bench/cpp/adv_networking_bench_default_tx_rx.yaml
./usr/bin/adv_networking_bench/cpp/adv_networking_bench_doca_tx_rx.yaml
./usr/bin/adv_networking_bench/cpp/adv_networking_bench_rmax_rx.yaml
./usr/bin/adv_networking_bench/python/main.py
./usr/lib/libadvanced_network_common.so
./usr/lib/libadvanced_network_rx.so
./usr/lib/libadvanced_network_tx.so
./usr/lib/libano_manager_doca.so
./usr/lib/libano_manager_dpdk.so
./usr/lib/libbasic_network.so
./usr/python/lib/holohub/advanced_network_common/__init__.py
./usr/python/lib/holohub/advanced_network_common/_advanced_network_common.cpython-310-aarch64-linux-gnu.so
./usr/python/lib/holohub/advanced_network_rx/__init__.py
./usr/python/lib/holohub/advanced_network_rx/_advanced_network_rx.cpython-310-aarch64-linux-gnu.so
./usr/python/lib/holohub/advanced_network_tx/__init__.py
./usr/python/lib/holohub/advanced_network_tx/_advanced_network_tx.cpython-310-aarch64-linux-gnu.so
./usr/python/lib/holohub/basic_network/__init__.py
./usr/python/lib/holohub/basic_network/basic_network_common.py
./usr/python/lib/holohub/basic_network/basic_network_rx.py
./usr/python/lib/holohub/basic_network/basic_network_tx.py
```

> _**Note**: if you do not pass `COMPONENTS` to `holohub_configure_deb`, the package will include every configured target with installation rules, so extra `-D PKG_`, `-D APP_` and `-D OP_` would currently pollute the content of your package._


cliffburdick (approved):
looks good! I'm a little confused about the location though. I thought this was going to go into the holoscan-sdk or holoscan-networking repo rather than holohub. Otherwise it's not clear to me what the difference is between a package and an operator if I were a user.

CONTRIBUTING.md
```diff
@@ -229,6 +236,49 @@ add_holohub_application(my_application DEPENDS
...
 and update for your new application.
 
+You can then build that application by configuring your build with `-D APP_my_application:BOOL=1` :
+
+```bash
+cmake -S . -B ./build -D APP_my_application:BOOL=1
+cmake --build ./build -j
+```
+
+### Adding a Packager
```
tbirdso, line 246: Nitpick: "packager" implies some defined entity that is doing packaging, seems like CMake/CPack are the "packager"

```suggestion
### Adding a Package
```

CONTRIBUTING.md
```diff
@@ -229,6 +236,49 @@ add_holohub_application(my_application DEPENDS
...
+
+1. Ensure the applications/operator cmake targets and/or files you want to package have an `install` rule define in your application/operator CMakeLists.txt files (see [CMake docs](https://cmake.org/cmake/help/latest/command/install.html)). Optionally, pass a `COMPONENT` argument to the `install` rule to control precisely which of your targets/files will get packaged. This can be useful if - for example - you want to create dev vs run packages, not include python bindings, or split your libraries in separate packages for different backends.
```
tbirdso, line 248:
```suggestion
1. Ensure the applications/operator CMake targets and/or files you want to package have an `install` rule define in your application/operator CMakeLists.txt files (see [CMake docs](https://cmake.org/cmake/help/latest/command/install.html)). Optionally, pass a `COMPONENT` argument to the `install` rule to control precisely which of your targets/files will get packaged. This can be useful if - for example - you want to create dev vs run packages, not include python bindings, or split your libraries in separate packages for different backends.
```

cmake/HoloHubConfigHelpers.cmake
```diff
@@ -13,6 +13,36 @@
...
+  message(DEBUG "${pkgname} apps = ${DEPS_APPLICATIONS}")
+  foreach(dep IN LISTS DEPS_EXTENSIONS)
+    set("EXT_${dep}" ON CACHE BOOL "Build the ${dep} GXF extension" FORCE)
+  endforeach()
+  foreach(dep IN LISTS DEPS_OPERATORS)
+    set("OP_${dep}" ON CACHE BOOL "Build the ${dep} holoscan operator" FORCE)
+  endforeach()
+  foreach(dep IN LISTS DEPS_APPLICATIONS)
+    set("APP_${dep}" ON CACHE BOOL "Build the ${dep} application" FORCE)
+  endforeach()
```
tbirdso, line 42: Seems like this block should be conditional on the package being enabled, otherwise every HoloHub build will always enable these options. See https://github.com/nvidia-holoscan/holohub/blob/main/cmake/HoloHubConfigHelpers.cmake#L28

```suggestion
if(PKG_${NAME})
  foreach(dep IN LISTS DEPS_EXTENSIONS)
    set("EXT_${dep}" ON CACHE BOOL "Build the ${dep} GXF extension" FORCE)
  endforeach()
  foreach(dep IN LISTS DEPS_OPERATORS)
    set("OP_${dep}" ON CACHE BOOL "Build the ${dep} holoscan operator" FORCE)
  endforeach()
  foreach(dep IN LISTS DEPS_APPLICATIONS)
    set("APP_${dep}" ON CACHE BOOL "Build the ${dep} application" FORCE)
  endforeach()
endif()
```

cmake/modules/holohub_configure_deb.cmake
```diff
@@ -0,0 +1,82 @@
...
+  else()
+    # package all installed targets
+    set(CPACK_DEB_COMPONENT_INSTALL 0)
+  endif()
+
+  # standard configurations
+  set(CPACK_STRIP_FILES TRUE)
+  set(CPACK_GENERATOR DEB)
+  set(CPACK_DEBIAN_FILE_NAME DEB-DEFAULT)
+  # set(CPACK_DEBIAN_PACKAGE_SHLIBDEPS ON)
```
tbirdso, line 65: Keep or remove?

tbirdso (commented):
@agirault this is looking great! Added comments.

tbirdso:
> looks good! I'm a little confused about the location though. I thought this was going to go into the holoscan-sdk or holoscan-networking repo rather than holohub. Otherwise it's not clear to me what the difference is between a package and an operator if I were a user.

@cliffburdick what is the `holoscan-networking` repo? Could you please link it here if it is publicly available?

@agirault maybe it would help to add a note to the [README](https://github.com/nvidia-holoscan/holohub?tab=readme-ov-file#overview) to distinguish packages from operators?

cmake/HoloHubConfigHelpers.cmake
agirault: It's done here: https://github.com/nvidia-holoscan/holohub/pull/637/files/d884c75abdc479a55121ab36a7855e07505805e8#diff-6771461c01c7c17ac1f47ce83a973b8923fce6ba006db2cbc0061b04abd1d858R24-R26

I just find the "return early" approach more elegant, and it also lowers the level of nesting below.

CONTRIBUTING.md
agirault: Yeah that's just semantics, but that's actually why I went for packager over package: contributors don't add a package (unlike how they actually add an app or operator source code), they add configurations to create a package, so I felt like packager was more appropriate? 

Or just `Adding a Package Configuration`?

cmake/modules/holohub_configure_deb.cmake
agirault: I would have liked to investigate it, but would need more tooling in the dev container to run it and more time to analyse results: https://cmake.org/cmake/help/v3.3/module/CPackDeb.html#variable:CPACK_DEBIAN_PACKAGE_SHLIBDEPS

I can leave this here commented, or remove

CONTRIBUTING.md
tbirdso: >Adding a Package Configuration

I like this one 👍

cmake/HoloHubConfigHelpers.cmake
tbirdso: Ah I missed that, thanks!

cmake/modules/holohub_configure_deb.cmake
tbirdso: Let's remove

cliffburdick:
> > looks good! I'm a little confused about the location though. I thought this was going to go into the holoscan-sdk or holoscan-networking repo rather than holohub. Otherwise it's not clear to me what the difference is between a package and an operator if I were a user.
> 
> @cliffburdick what is the `holoscan-networking` repo? Could you please link it here if it is publicly available?
> 
> @agirault maybe it would help to add a note to the [README](https://github.com/nvidia-holoscan/holohub?tab=readme-ov-file#overview) to distinguish packages from operators?

It doesn't exist, but I think some of the original conversation was whether to put it in holohub, holoscan-sdk or a new repo here: https://github.com/nvidia-holoscan

I'm in favor of the last option because it's more modular and it's a very large component at this point that is less and less of a fit for holohub.

tbirdso (approved):
LGTM. Per offline discussion ANO will remain in HoloHub for the time being.

@agirault merge when ready!

PR #638, opened 2024-12-12
Added support for multiple queues per core

cliffburdick:


agirault:
could you add a test?

cliffburdick:
> could you add a test?

Doing this might be a little tricky since we don't have an example of sending packets to two different ports. We may have to set up two different operators running independently to do this. Alternatively we could expand the main sample to allow that.

agirault:
> _I hate that I can't create a thread on github MRs!_

Right looks like we need two things for simpler testing of our apis:
1. sw loopback to run 
2. an example that leverage that capability

Any unit tests we could start setting up instead in the meantime? or is this so tied to the system that we really need integration tests? in which case, I think we really need (2) (that's one use case, but we'll need it for more test scenarios like HDS, etc...) and we can leverage the current test infra to run on a known hw like we have in our nightly build/test. (1) will be nice to run on more machines and scale more tests, but I think (2) can be sufficient for now.

I recall you were talking about setting an example that would do tx and rx and both ports to test 200gbps, I suppose if we did that we could use it to test BW but also to test this MR?

cliffburdick:
> > _I hate that I can't create a thread on github MRs!_
> 
> Right looks like we need two things for simpler testing of our apis:
> 
> 1. sw loopback to run
> 2. an example that leverage that capability
> 
> Any unit tests we could start setting up instead in the meantime? or is this so tied to the system that we really need integration tests? in which case, I think we really need (2) (that's one use case, but we'll need it for more test scenarios like HDS, etc...) and we can leverage the current test infra to run on a known hw like we have in our nightly build/test. (1) will be nice to run on more machines and scale more tests, but I think (2) can be sufficient for now.
> 
> I recall you were talking about setting an example that would do tx and rx and both ports to test 200gbps, I suppose if we did that we could use it to test BW but also to test this MR?

The loopback MR is not merged yet. I'll need to learn how to add your tests and run them using your script, but it would be good to do it for all of them. The 200G test should also be easy to add in.

The challenge with this particular PR is the transmitter needs to send to two different ports so we have at least two queues being used. This would require changes to the code that I wasn't sure if we wanted to introduce the complexity at this point. The other option is we can use scapy to just send packets from python at a very low rate and just make sure it gets them in the proper queues.

agirault:
> This would require changes to the code that I wasn't sure if we wanted to introduce the complexity at this point

I think we need something to test it (I assume you wrote an app or edited one already to test locally), and if we're concerned about updating the current benchmarking app and make it too complex, I agree, let's just create a separate one then (or do what you suggest with scapy), but my recommendation would be to stick to proper holoscan/ano test app.

cliffburdick:
> > This would require changes to the code that I wasn't sure if we wanted to introduce the complexity at this point
> 
> I think we need something to test it (I assume you wrote an app or edited one already to test locally), and if we're concerned about updating the current benchmarking app and make it too complex, I agree, let's just create a separate one then (or do what you suggest with scapy), but my recommendation would be to stick to proper holoscan/ano test app.

Yes, I used scapy to test it, but ultimately I wanted the customer that submitted the big PR to holohub to test it also since they have a transmitter that will stress it. I can set up a scapy test.

agirault:
Let's just setup something so we can run automated tests, the project is growing, we can't expect a customer/partner local test to not be affected by all the incoming changes.

cliffburdick:
> could you add a test?

Test has been added

cliffburdick:
@agirault is this ready to merge or are we stuck?

agirault (approved):
I'm going to approve since I'll be looking at refactoring the test harness later anyway.

PR #640, opened 2024-12-13
Added per-queue timeout value

cliffburdick:
Right now a batch is sent to the application when the number of packets matches the batch size. This might be fine for the nominal case of steady traffic, but when the traffic is less predictable and bursty we may want a timeout that packets will be set if any amount have been received.


tbirdso:
Thanks @dylan-eustice for confirming review/testing offline.

tbirdso (approved):
Approving following offline discussion

PR #639, opened 2024-12-13
Improvements to run list and run build

agirault:
## Context

While preparing support for generating packages from the `run` and `dev_container` scripts, having a good `list` mechanism seemed like it would impact existing infrastructure. Instead of a large MR that mixes infra improvements and packaging support, this MR is scoped for infra improvements that the packaging support will then be able to leverage.

### List improvements

- list applications, operators, and packages
- uses simpler but more robust heuristic based on required cmake functions needed to configure these targets (while the metadata.json should be present but might not be)

### Build improvements

- support building operators in addition to apps
- verify input validity using the improve listing from the last commit
- allow passing a path to the app/operator dir, for improve UX by leveraging
terminal autocompletion

### List improvements

- enhanced human-readable list of targets to build or package
```bash
$ ./run list

== APPLICATIONS =================

adv_networking_bench
basic_networking_ping
body_pose_estimation
colonoscopy_segmentation
cvcuda_basic
dds_video
deltacast_transmitter
endoscopy_depth_estimation
endoscopy_out_of_body_detection
endoscopy_tool_tracking
evt_cam_calibration
grpc_endoscopy_tool_tracking
grpc_h264_endoscopy_tool_tracking
h264_endoscopy_tool_tracking
h264_video_decode
high_speed_endoscopy
holoviz_hdr
holoviz_srgb
holoviz_ui
holoviz_vsync
holoviz_yuv
hyperspectral_segmentation
imaging_ai_segmentator
laser_detection
model_benchmarking
multiai_endoscopy
multiai_ultrasound
network_radar_pipeline
nvidia_nim_chat
nvidia_nim_imaging
nvidia_nim_nvclip
object_detection_torch
openigtlink_3dslicer
orsi_in_out_body
orsi_multi_ai_ar
orsi_segmentation_ar
prohawk_video_replayer
pva_video_filter
qt_video_replayer
realsense_visualizer
simple_pdw_pipeline
simple_radar_pipeline
tao_peoplenet
ucx_endoscopy_tool_tracking
ucx_h264_endoscopy_tool_tracking
ultrasound_segmentation
usb_cam_calibration
velodyne_lidar_app
video_deidentification
volume_rendering
volume_rendering_xr
webrtc_video_server
xr_hello_holoscan

== OPERATORS ====================

advanced_network
apriltag_detector
basic_network
cvcuda_holoscan_interop
dds_shapes_subscriber
dds_video_publisher
dds_video_subscriber
deltacast_videomaster
emergent_source
grpc_operators
[...]

PR #646, opened 2024-12-17
Fix run list for non-cmake targets

agirault:
Regression introduced in https://github.com/nvidia-holoscan/holohub/pull/639

was only looking for cmake targets, did not account for pure python targets

Before: 53 apps.
After: ~~97~~ 96 apps.

### DIFF
--- a/run
+++ b/run
@@ -690,13 +690,13 @@ build() {
   opt_prefix=""
   is_app=false
   is_pkg=false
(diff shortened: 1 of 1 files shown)


tbirdso (approved):
LGTM, tested and verified locally with `./run list` and `./run list_apps`. @agirault feel free to merge.

PR #648, opened 2024-12-19
Build: Set default `CUDA_ARCHITECTURES` for builds

tbirdso:
Set a default `CMAKE_CUDA_ARCHITECTURES` list for builds to inherit that covers most modern architectures.

The list here is at parity with Holoscan SDK build defaults. Subprojects in HoloHub may choose to override this value with their own default settings or set `CUDA_ARCHITECTURES` directly on targets.

Resolves an observed issue on IGX 1.1 where not setting this property resulted in a default value of "52" in the Multi-AI Ultrasound application, leading to a "PTX" failure message. With this fix the Multi-AI Ultrasound
    application inherits the updated list of compute capabilities by default.
    
Note that existing builds may need to clear the `CMakeCache.txt` or the
    specific architectures cache entry to apply the fix.

Reference: https://cmake.org/cmake/help/v3.24/variable/CMAKE_CUDA_ARCHITECTURES.html


CMakeLists.txt
```diff
@@ -57,6 +57,11 @@ set(HOLOHUB_DATA_DIR "${CMAKE_BINARY_DIR}/data" CACHE PATH "Data Download direct
 set(CMAKE_INSTALL_RPATH
     "\$ORIGIN:\$ORIGIN/../../../lib:\$ORIGIN/../../lib:/opt/nvidia/holoscan/lib/")
 
+# Set the default CUDA target platforms
```
AndreasHeumann, line 60: Should add the Holoscan SDK https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/cmake/modules/SetupCUDA.cmake to Holohub and use that?
This would make sure that Holohub and Holoscan SDK use the same mechanism.

I also noticed that there are may instances of CUDA_ARCHITECTURES and CMAKE_CUDA_ARCHITECTURES scattered throughout all the CMakeLists.txt files. Should we remove them and consolidate in this one central location?

Except for the tool_tracking_postprocess which needs `native` because it's using dynamic parallelism.

AndreasHeumann (changes requested):

CMakeLists.txt
tbirdso: Great suggestion. I'll update this PR to align to `SetupCUDA.cmake`.

The existing values for `CUDA_ARCHITECTURES` / `CMAKE_CUDA_ARCHITECTURES` throughout HoloHub generally look ok. We should allow projects to override the HoloHub default. I'd lean towards leaving those as-is, but probably no major impact either way (except for the tool tracking case you noted).

tbirdso: I'm reconsidering whether we should use `SetupCUDA.cmake` to set HoloHub project defaults due to the inlined  [enable_language call](https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/cmake/modules/SetupCUDA.cmake#L33). `SetupCUDA.cmake` assumes the Holoscan SDK build environment is in use, whereas HoloHub projects may not require CUDA.

Here are two relevant HoloHub details:
1. The CUDA language should be optional on a per-project basis. Many projects do not depend directly on the CUDA Toolkit nor require CUDA language support, such as pure Python projects. Using `SetupCUDA.cmake` to set HoloHub project defaults would add an unnecessary dependency in those cases.
2. Most HoloHub projects do not consider portability at all, but instead demonstrating building and running an application on only the user's machine. `SetupCUDA.cmake` tailors a portable list of architectures for CUDA compilation and is mostly extraneous in that case.

Based on that, I suggest that `"native"` might be the appropriate `CMAKE_CUDA_ARCHITECTURES` _default_ value in HoloHub. Reflecting the details above:
1. Setting the CMake `CMAKE_CUDA_ARCHITECTURES` variable at the top level _without_ enabling the CUDA language makes the default available to projects using CUDA without introducing a new dependency to non-CUDA projects.
2. Setting the default as "native" matches the standard HoloHub default target platform which is the user's system. Projects requiring portability can override that default such as done in `SetupCUDA.cmake`.

@AndreasHeumann what do you think?

CMakeLists.txt
```diff
@@ -57,6 +57,11 @@ set(HOLOHUB_DATA_DIR "${CMAKE_BINARY_DIR}/data" CACHE PATH "Data Download direct
...
+if(NOT CMAKE_CUDA_ARCHITECTURES)
+  set(CMAKE_CUDA_ARCHITECTURES "60;70;72;75;80;86;87;89;90;90a")
```
tbirdso, line 62: Suggested change based on above discussion:

```suggestion
  set(CMAKE_CUDA_ARCHITECTURES "native")
```

jjomier (approved):

tbirdso:
Moved forward with the "native" stopgap in https://github.com/nvidia-holoscan/holohub/pull/651. Let's continue discussion on best practice for a long term fix in this PR.

CMakeLists.txt
AndreasHeumann: Good points, I agree with you.

tbirdso:
Closing as duplicate of https://github.com/nvidia-holoscan/holohub/pull/651

PR #649, opened 2024-12-22
Remove workaround after dependencies have been fixed

MMelQin:
This PR remove workarounds that had to be used in the earlier merged versions.
- Two requirements files were needed, with the first one install holoscan since it pins its dependencies, and the second with all other dependencies to avoid version conflicts of a couple module (without functionality issue). Holoscan has since removed/relaxed pinning, so a single requirements file works
- Single Dockerfile supporting both x86_64 and aarch64 builds. Python Wheel for Torch on aarch64 and CUDA 12.6 was not available, and around 12/18/2024 the nightly build Wheel started appearing. As of now, RC test version is also availabe, so the pull request had been updated to make use of this RC build while waiting for the released version in early Jan 2025.

The[ release date of Torch 2.6 CUDA 12.6](https://dev-discuss.pytorch.org/t/pytorch-release-2-6-0-final-rc-is-available/2748) is now set at 01/29/2025. This app will be updated later to use that version, and for now, OK to be reviewed and merged.

Screenshot of running the app on IGX DevKit 1.0GA is attached.
HoloHubApp-Screenshot_2025-01-22


jjomier:
@MMelQin is this ready to merge after review?

MMelQin:
> @MMelQin is this ready to merge after review?

I am adding more changes and testing on IGX dev kit. Plan to have it reviewed later this week. Was giving the other SDK a bit of TLC last week.

tbirdso (approved):
LGTM

PR #652, opened 2025-01-02
Infra: Bump for Holoscan SDK v2.8.0 container release

tbirdso:
Bump default HoloHub base container to Holoscan SDK v2.8.0

### DIFF
--- a/dev_container
+++ b/dev_container
@@ -460,11 +460,11 @@ get_host_gpu() {
 }
 
 get_default_base_img() {
-    echo -n "nvcr.io/nvidia/clara-holoscan/holoscan:v2.7.0-"$(get_host_gpu)
+    echo -n "nvcr.io/nvidia/clara-holoscan/holoscan:v2.8.0-"$(get_host_gpu)
 }
 
 get_default_img() {
-    echo -n "holohub:ngc-v2.7.0-"$(get_host_gpu)
+    echo -n "holohub:ngc-v2.8.0-"$(get_host_gpu)
 }
 
 # This function returns the compute capacity of the system's GPU (8.6, 8.9, etc.)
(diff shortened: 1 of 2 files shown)


wendell-hom:
@tbirdso do we need to bump the BASE_IMAGE in by `operators/advanced_network/Dockerfile`?

tbirdso:
> @tbirdso do we need to bump the BASE_IMAGE in by `operators/advanced_network/Dockerfile`?

Added to move forward in this PR. In the future ANO should probably rely on the default base image provided by HoloHub via the `dev_container` script rather than providing an inline default, as it doesn't scale well for us to update individual projects when we bump Holoscan SDK versions. cc @agirault @cliffburdick 

See the VILA Dockerfile example, which avoids providing a default `BASE_IMAGE` and instead relies on HoloHub to provide the latest base container: https://github.com/nvidia-holoscan/holohub/blob/main/applications/vila_live/Dockerfile#L18

wendell-hom:
> > @tbirdso do we need to bump the BASE_IMAGE in by `operators/advanced_network/Dockerfile`?
> 
> Added to move forward in this PR. In the future ANO should probably rely on the default base image provided by HoloHub via the `dev_container` script rather than providing an inline default, as it doesn't scale well for us to update individual projects when we bump Holoscan SDK versions. cc @agirault @cliffburdick
> 
> See the VILA Dockerfile example, which avoids providing a default `BASE_IMAGE` and instead relies on HoloHub to provide the latest base container: https://github.com/nvidia-holoscan/holohub/blob/main/applications/vila_live/Dockerfile#L18

Yes, sounds good.  Would be good to reduce the duplication to update BASE_IMAGE

wendell-hom (approved):

PR #655, opened 2025-01-10
Update holoscan version

bhashemian:
Update the holoscan Docker image version to 2.8.0 in the README, which is the version that will be installed in the Docker images.


tbirdso (approved):
Hi @drbeh , looks good, thanks! Do you have permissions to merge?

PR #656, opened 2025-01-14
Remove call to install CLI dependencies

mocsharp:
Since CLI is no longer bundled with HSDK (2.9 and onward), we can remove the call to parse and install CLI dependencies.


tbirdso (approved):
Thanks @mocsharp . Should this be merged now or when 2.9 releases?

mocsharp:
> Thanks @mocsharp . Should this be merged now or when 2.9 releases?

For 2.9 please. Thank you.

PR #666, opened 2025-01-28
Adding AJA operator - LWS2 branch

jjomier:
AJA operator is removed from HSDK for LWS2. Adding it in Holohub.


tbirdso (approved):
Changes LGTM. I do not have hardware to test.

Has AJA been removed from upstream sources yet?

PR #670, opened 2025-01-30
Use AJA code for HSDK 1.0 for LWS2 branch

jjomier:
The LWS2 branch is based on HSDK 1.0 so using AJA code from 1.0 instead


ichergui (approved):
LGTM. Thanks @jjomier

PR #683, opened 2025-02-14
Fix `aja_video_capture` build and test coverage

tbirdso:
Changes:
- Fix `aja_video_capture` build and run errors resulting from missing
    Holoscan package reference and malformed import statement
- Update patch files for testing following HoloHub migration
    
Resolves observed errors where AJA source was not covered in testing.
    
    Note: Not tested on AJA hardware. Verified that apps now build and run with
    `BUILD_TESTING=ON` and that C++/Python tests run and attempt to open AJA device.

Internal tracking: NVBug 5111567


jjomier (approved):
Thanks for taking care of this!

PR #684, opened 2025-02-14
App: Use HoloHub AJA migration in `endoscopy_out_of_body_detection` app

tbirdso:
Update `endoscopy_out_of_body_detection` Python app to import AJA operator from HoloHub rather than Holoscan SDK.

AJASourceOp has been migrated preemptively to HoloHub and will be removed from Holoscan SDK in v3.0. See 0af6365

To resolve internal dashboard failure with v3.0 testing.


bhashemian (approved):
Thanks for this update, @tbirdso! It looks good to me.

PR #702, opened 2025-02-24
Counting the number of unique applications, tutorials, ...

jjomier:


tbirdso (approved):
Looks good. Note that project counts will be less than featured in the HoloHub GitHub badges at https://github.com/nvidia-holoscan/holohub, which are counting impl directories (cpp/python) separately.

jjomier:
@tbirdso maybe we should remove the badge on the main page. what do you think?

tbirdso (approved):

tbirdso:
Looks great, thanks! Have not tested locally.

PR #705, opened 2025-02-26
Improve holoscan-networking package

agirault:
## Context

`holoscan-networking` Debian package is including some non-operational libs (bindings), missing some utilities, and installing on generic system paths.

## Changes

- Remove python bindings not operational
- Add missing installs for basic networking app and configs
- Add tune system script + make it executable
- Update install prefix to join with the hsdk installation under /opt/nvidia/holoscan
- install ANO examples in examples and not bin to match hsdk

### Before:
```bash
dpkg -L holoscan-networking | while read -r f; do [ -f "$f" ] && echo "$f"; done
/usr/bin/adv_networking_bench/cpp/adv_networking_bench_default_rx.yaml
/usr/bin/adv_networking_bench/cpp/adv_networking_bench_default_rx_multi_q.yaml
/usr/bin/adv_networking_bench/cpp/adv_networking_bench_default_tx.yaml
/usr/bin/adv_networking_bench/cpp/adv_networking_bench_default_tx_rx.yaml
/usr/bin/adv_networking_bench/cpp/adv_networking_bench_doca_tx_rx.yaml
/usr/bin/adv_networking_bench/cpp/adv_networking_bench_rmax_rx.yaml
/usr/bin/adv_networking_bench/python/main.py
/usr/lib/libadvanced_network_common.so
/usr/lib/libadvanced_network_rx.so
/usr/lib/libadvanced_network_tx.so
/usr/lib/libano_manager_doca.so
/usr/lib/libano_manager_dpdk.so
/usr/lib/libbasic_network.so
/usr/python/lib/holohub/advanced_network_common/__init__.py
/usr/python/lib/holohub/advanced_network_common/_advanced_network_common.cpython-310-aarch64-linux-gnu.so
/usr/python/lib/holohub/advanced_network_rx/__init__.py
/usr/python/lib/holohub/advanced_network_rx/_advanced_network_rx.cpython-310-aarch64-linux-gnu.so
/usr/python/lib/holohub/advanced_network_tx/__init__.py
/usr/python/lib/holohub/advanced_network_tx/_advanced_network_tx.cpython-310-aarch64-linux-gnu.so
/usr/python/lib/holohub/basic_network/__init__.py
/usr/python/lib/holohub/basic_network/basic_network_common.py
/usr/python/lib/holohub/basic_network/basic_network_rx.py
/usr/python/lib/holohub/basic_network/basic_network_tx.py
```

### After:
```bash
dpkg -L holoscan-networking | while read -r f; do [ -f "$f" ] && echo "$f"; done
/opt/nvidia/holoscan/bin/tune_system.py
/opt/nvidia/holoscan/examples/adv_networking_bench/adv_networking_bench
/opt/nvidia/holoscan/examples/adv_networking_bench/adv_networking_bench_default_rx.yaml
/opt/nvidia/holoscan/examples/adv_networking_bench/adv_networking_bench_default_rx_multi_q.yaml
/opt/nvidia/holoscan/examples/adv_networking_bench/adv_networking_bench_default_tx.yaml
/opt/nvidia/holoscan/examples/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
/opt/nvidia/holoscan/examples/adv_networking_bench/adv_networking_bench_doca_tx_rx.yaml
/opt/nvidia/holoscan/examples/adv_networking_bench/adv_networking_bench_rmax_rx.yaml
/opt/nvidia/holoscan/examples/basic_networking_ping/basic_networking_ping
/opt/nvidia/holoscan/examples/basic_networking_ping/basic_networking_ping_rx.yaml
/opt/nvidia/holoscan/examples/basic_networking_ping/basic_networking_ping_tx.yaml
/opt/nvidia/holoscan/lib/libadvanced_network_common.so
/opt/nvidia/holoscan/lib/libadvanced_network_rx.so
/opt/nvidia/holoscan/lib/libadvanced_network_tx.so
/opt/nvidia/holoscan/lib/libano_manager_doca.so
/opt/nvidia/holoscan/lib/libano_manager_dpdk.so
/opt/nvidia/holoscan/lib/libbasic_network.so
```


agirault:
@jjomier @tbirdso @mocsharp the holohub default app installs seem to be `<prefix>/bin/<app_name>/<lang>/<app_name>`

When we set `<prefix>` to `/opt/nvidia/holoscan` with CPack (for the debian package generation), holohub operators, python libs, and util scripts are installed in a consistent way compared to the HSDK (making it easier for linking, importing, and running. nice). However, apps install path are not consistent with the HSDK sample apps, which is under `examples` and not `bin` (`bin` is currently more scripts/utilities for the hsdk)

Any thoughts? I'm fine with either but figured I'd ask if there are opinions. Maybe having them under `examples` (like hsdk) or `holohub` or `holohub_apps` makes more sense than under `bin`?

cliffburdick (approved):

cmake/modules/holohub_configure_deb.cmake
```diff
@@ -59,6 +59,7 @@ function(holohub_configure_deb)
   endif()
 
   # standard configurations
+  set(CPACK_PACKAGING_INSTALL_PREFIX "/opt/nvidia/holoscan")
```
tbirdso, line 62: Should default installation be to `/opt/nvidia/holoscan` or distinct `/opt/nvidia/holohub` or similar?

agirault: I think at this stage probably fully subjective, and affects the question here also: https://github.com/nvidia-holoscan/holohub/pull/705#issuecomment-2685421095

If we think about holohub packaged operators/apps as "modules" to expand holoscan, installing it in `/opt/nvidia/holoscan` makes sense and makes ld and python import discovery easier by piggy backing on the HSDK paths.

tbirdso: Fair point, if folder pollution doesn't seem like a concern then not a blocker. Agreed it's nice to not have to update ldconfig and PYTHONPATH for libraries. PATH update may be necessary regardless of resolution in https://github.com/nvidia-holoscan/holohub/pull/705#:~:text=question%20here%20also%3A-,%23705%20(comment),-If%20we%20think.

agirault:
_\<vent>Hate that we can't have threads on github comments!\</vent>_

> Seems reasonable to stick with bin as a common standard for executable binaries (see [CMAKE_INSTALL_BINDIR](https://cmake.org/cmake/help/v3.24/command/install.html#installing-targets)). Noted above, IMO HoloHub packages should define and install to distinct system folders rather than polluting /opt/nvidia/holoscan. Downstream packages do not need to follow Holoscan SDK install heuristics

That's one option, seems to be subjective, maybe we can garner more feedback. I'll be fine with it either way, but I don't think we should just dismiss installing under `/opt/nvidia/holoscan` without more alignment. If we do follow the hsdk install heuristics (which I'd call extending it, not polluting it), not using `bin` for the examples would make sense.

> For instance, the Holoscan SDK examples folder is reasonably distinct because it contains binaries, Python files, and buildable source code. We do not expect holohub-networking to provide more than binaries and libraries.

We do:
- as soon as python apps are functional they'd be in `<prefix>/<app_name>/<lang>/<app_name>`
- only discussed in 1:1 with @jjomier so far, but he asked yesterday that we work towards distributing build-able source code for the `holoscan-networking` basic sample app.

tbirdso (approved):
🎉

PR #707, opened 2025-02-26
Rename ANO doca manager string to gpunetio

agirault:


operators/advanced_network/adv_network_types.h
```diff
@@ -173,7 +173,7 @@ enum class AnoMgrType {
 };
 
 static constexpr const char* ANO_MGR_STR__DPDK = "dpdk";
-static constexpr const char* ANO_MGR_STR__DOCA = "doca";
+static constexpr const char* ANO_MGR_STR__DOCA = "gpunetio";
```
cliffburdick, line 176: Should we change that string variable name?

operators/advanced_network/adv_network_types.h
```diff
@@ -188,7 +188,7 @@ inline AnoMgrType manager_type_from_string(const std::string& str) {
   if (str == ANO_MGR_STR__DOCA) return AnoMgrType::DOCA;
```
cliffburdick, line 188: Change the enum too?

operators/advanced_network/adv_network_types.h
agirault: Yeah thought the same thing when reviewing again. Not necessary but trivial enough, did it here: https://github.com/nvidia-holoscan/holohub/compare/a3c2dee23181b2a342e5b014b409583cf98eed04..5a352e3a58763a5e9d92a6bbf75f716b2fba879f

operators/advanced_network/adv_network_types.h
agirault: either, I feel like that one can wait until we rename the actual manager `DocaMgr`, and mapping from gpunetio as a string to doca as a manager is consistent. but fine either way, that one is low impact also

cliffburdick (approved):

cliffburdick (approved):

PR #709, opened 2025-02-27
Support Ninja Build in holohub infra

agirault:
# Context

Ninja if a fairly standard build system that generally shows significant speedups over Makefile. It is used by the Holoscan SDK among so many other things, but isn't leveraged by Holohub infra.

# Changes

- Add ninja-build to the `setup` script 
- Using the `Ninja` generator with CMake if `ninja` is present
- Update `run build` for readability

# Results

Example: `./run clear_cache && ./dev_container build_and_package holoscan-networking`

- before: 4m 58s 
- after: 3m 32s -> **29% speedup for this scenario**

### DIFF
--- a/run
+++ b/run
@@ -360,6 +360,13 @@ setup() {
     apt install --no-install-recommends -y cmake cmake-curses-gui
   fi
 
+  # Install Ninja
+  ninja_version=$(dpkg --status ninja-build 2>/dev/null | grep -Po '^Version: \K[^-]*')
+  if [[ $ninja_version == "" ]]; then
+    echo "Installing ninja-build"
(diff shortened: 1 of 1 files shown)


tbirdso:
Changes look good, slight concern about breaking backwards compatibility with existing build trees. How would you feel about waiting and merging this alongside the planned Holoscan SDK v3.0 release?

agirault:
Indeed, need a clean build dir.

No problem, pull this when you want to.

tbirdso (approved):
Verified for 3.0.0. Developers will need to clean their build directories to use Ninja.

PR #715, opened 2025-03-07
Fix port IDs when different from order in the config file

cliffburdick:
Currently the TX example expected the port ID to be 0, and it was hard-coded that way. If a user tried to change it to 1 it would not work due to some problems in the initialization order in DPDK vs the ANO order. This PR fixes these to be consistent.

Fix was tested with original configuration and the new one with the ports swapped.

### DIFF
--- a/applications/adv_networking_bench/cpp/default_bench_op_tx.h
+++ b/applications/adv_networking_bench/cpp/default_bench_op_tx.h
@@ -79,8 +79,7 @@ class AdvNetworkingBenchDefaultTxOp : public Operator {
     HOLOSCAN_LOG_INFO("AdvNetworkingBenchDefaultTxOp::initialize()");
     holoscan::Operator::initialize();
 
-    // port_id_ = adv_net_address_to_port(address_.get());
-    port_id_ = 0;
+    port_id_ = adv_net_address_to_port(address_.get());
 
     size_t buf_size = batch_size_.get() * payload_size_.get();
     if (!gpu_direct_.get()) {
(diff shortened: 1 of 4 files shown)


applications/adv_networking_bench/cpp/default_bench_op_tx.h
```diff
@@ -122,6 +117,9 @@ class AdvNetworkingBenchDefaultTxOp : public Operator {
       cudaMemset(gds_header_, 0, header_size_.get());
 
       populate_dummy_headers(pkt);
+      // ANO expects host order when setting
+      ip_src_ = ntohl(ip_src_);
+      ip_dst_ = ntohl(ip_dst_);
```
agirault, line 122: can you explain why that's only done in gpu-only, and not cpu or HDS ?

agirault:
Can you explain the issue and the change in the MR description?

applications/adv_networking_bench/cpp/default_bench_op_tx.h
cliffburdick: In GPU-only mode we need to construct the header one time on the host and put it in the correct order in memory before sending it out. In other modes the header is on the CPU and the ANO can do the swap for us. In reality a GPU kernel would do this if you were using GPU-only mode.

cliffburdick:
> Can you explain the issue and the change in the MR description? And how you tested the fix?

Added

agirault (approved):

PR #716, opened 2025-03-07
Contours instead of masks in colonoscopy segmentation app

brudfors:
I have modified the colonoscopy segmentation app to, instead of showing the binary segmentation masks, showing contours of the same masks:

image

The contouring is done on the GPU using CuPy. The implementation is in an operator `ContourOp` and supports categorical segmentation masks (i.e, 0,1,2,...,N): https://github.com/brudfors/holohub/blob/mbrudfors/contour/applications/colonoscopy_segmentation/colonoscopy_segmentation.py#L34

Having an example of a contour visualization of segmentation masks is valuable as it is a popular way of showing segmentations that does not "clutter" the visualization as much if the masks are large.


jjomier (approved):
LGTM!

jjomier:
@brudfors some linting errors and Gemini AI recommendations

jjomier:
@brudfors I think this is going to break our automated tests (unless we change the baseline images). Should we make the contours optional?

brudfors:
Thanks @jjomier! Recommendations added and lint fixes done.

jjomier (approved):

PR #718, opened 2025-03-07
Holoscan Networking Tutorial

agirault:
Description coming...

### DIFF
--- a/dev_container
+++ b/dev_container
@@ -878,8 +878,7 @@ build_and_package() {
         c_echo "Package ${pkg} not found in list of packages: (see ./run list_packages)"
         exit 1
     fi
-    dockerfile=$(./run get_pkg_dockerfile ${pkg})
-    build_and_run $pkg --no_run --docker_file ${dockerfile} "$@"
+    build_and_run $pkg --no_run "$@"
 }
 
 build_and_install_desc() { c_echo 'Build and install the binaries into install/<app> if supported
@@ -1084,7 +1083,12 @@ build_and_run() {
     fi
 
     if [[ -z "${docker_file}" ]]; then
-        docker_file=$(./run get_app_dockerfile ${app_name} ${app_language:-cpp})
+        packages=$(${HOLOHUB_ROOT}/run list_packages | awk '{print $1}')
+        if [[ " ${packages[*]} " =~ " ${app_name} " ]]; then
+            docker_file=$(./run get_pkg_dockerfile ${app_name})
+        else
+            docker_file=$(./run get_app_dockerfile ${app_name} ${app_language:-cpp})
+        fi
     fi
     container_build_args+=(--docker_file ${docker_file})
 
--- a/tutorials/high_performance_networking/README.md
+++ b/tutorials/high_performance_networking/README.md
@@ -1,0 +1,2185 @@
+## High Performance Networking with Holoscan
+
+This tutorial demonstrates how to use the advanced networking Holoscan operator (often referred to as ANO or `advanced_network` in HoloHub) for low latency and high throughput communication through NVIDIA SmartNICs. With a properly tuned system, the advanced network operator can achieve hundreds of Gbps with latencies in the low microseconds.
+
+!!! note
+
+    This solution is designed for users who want to create a Holoscan application that will interface with an external system or sensor over Ethernet.
+
+    - For high performance communication with systems also running Holoscan, refer to the [Holoscan distributed application documentation](https://docs.nvidia.com/holoscan/sdk-user-guide/holoscan_create_distributed_app.html) instead.
+    - For JESD-compliant sensor without Ethernet support, consider the [Holoscan Sensor Bridge](https://docs.nvidia.com/holoscan/sensor-bridge/latest/introduction.html) for an FPGA-based interface to Holoscan.
+
+## Prerequisites
+
+Achieving High Performance Networking with Holoscan requires a system with an [**NVIDIA SmartNIC**](https://www.nvidia.com/en-us/networking/ethernet-adapters/) and a [**discrete GPU**](https://www.nvidia.com/en-us/design-visualization/desktop-graphics/). That is the case of [NVIDIA Data Center](https://www.nvidia.com/en-us/data-center/) systems, or edge systems like the [NVIDIA IGX](https://www.nvidia.com/en-us/edge-computing/products/igx/) platform and the [NVIDIA Project DIGITS](https://www.nvidia.com/en-us/project-digits/). `x86_64` systems equipped with these components are also supported, though the performance will vary greatly depending on the PCIe topology of the system (more on this [below](#31-ensure-ideal-pcie-topology)).
+
+In this tutorial, we will be developing on an **NVIDIA IGX Orin platform** with [IGX SW 1.1](https://docs.nvidia.com/igx-orin/user-guide/latest/base-os.html) and an [NVIDIA RTX 6000 ADA GPU](https://www.nvidia.com/en-us/design-visualization/rtx-6000/), but the concepts should be applicable to other systems based on Ubuntu 22.04 as well. It should also work on other Linux distributions with a glibc version of 2.35 or higher by containerizing the dependencies and applications on top of an Ubuntu 22.04 image, but this is not actively tested at this time.
+
+## Background
+
+Achieving high performance networking is a complex problem that involves many system components and configurations which we will cover in this tutorial. Two of the core concepts to achieve this are named Kernel Bypass, and GPUDirect.
+
+### Kernel Bypass
+
+In this context, Kernel Bypass refers to bypassing the operating system's kernel to directly communicate with the network interface (NIC), greatly reducing the latency and overhead of the Linux network stack. There are multiple technologies that achieve this in different fashions. They're all Ethernet-based, but differ in their implementation and features. The goal of the `advanced_network` operator in Holoscan Networking is to provide a common higher-level interface to all these backends:
+
+- **RDMA**: Remote Direct Memory Access, using the open-source [`rdma-core`](https://github.com/linux-rdma/rdma-core) library. It differs from the other Ethernet-based backends with its server/client model and RoCE (RDMA over Ethernet) protocol. Given the extra cost and complexity to setup on both ends, it offers a simpler user interface, orders packets on arrival, and is the only one to offer a high reliability mode.
+- **DPDK**: the Data Plane Development Kit is an open-source project part of the Linux Foundation with a strong and long-lasting community support. Its RTE Flow capability is generally considered the most flexible solution to split packets ingress and egress data.
+- **DOCA GPUNetIO**: This NVIDIA proprietary technology differs from the other backends by transmitting and receiving packets from the NIC using a GPU kernel instead of CPU code, which is highly beneficial for CPU-bound applications.
+- **NVIDIA Rivermax**: NVIDIA's other proprietary kernel bypass technology. For a license fee, it should offer the lowest latency and lowest resource utilization for video streaming (RTP packets).
+
+??? example "Work In Progress"
+
+    The Holoscan Advanced Networking Operator integration testing infrastructure is under active development. As such:
+
+    - The **DPDK** backend is supported and distributed with the `holoscan-networking` package, and is the only backend actively tested at this time.
+    - The **DOCA GPUNetIO** backend is supported and distributed with the `holoscan-networking` package, with testing infrastructure under development.
+    - The **NVIDIA Rivermax** backend is supported for Rx only when building from source, but not yet distributed nor actively tested. Tx support is under development.
+    - The **RDMA** backend is under active development and should be available soon.
+
+Which backend is best for your use case will depend on multiple factors, such as packet size, batch size, data type, and more. The goal of the Advanced Networking Operator is to abstract the interface to these backends, allowing developers to focus on the application logic and experiment with different configurations to identify the best technology for their use case.
+
+### GPUDirect
+
+`GPUDirect` allows the NIC to read and write data from/to a GPU without requiring to copy the data the system memory, decreasing CPU overheads and significantly reducing latency. An implementation of `GPUDirect` is supported by all the kernel bypass backends listed above.
+
+!!! Warning
+
+    `GPUDirect` is only supported on Workstation/Quadro/RTX GPUs and Data Center GPUs. It is not supported on GeForce cards.
+
+??? info "How does that relate to peermem or dma-buf?"
+
+    There are two interfaces to enable `GPUDirect`:
+
+    - The [`nvidia-peermem`](https://docs.nvidia.com/cuda/gpudirect-rdma/) kernel module, distributed with the NVIDIA DKMS GPU drivers.
+        - Supported on Ubuntu kernels 5.4+, deprecated starting with kernel 6.8.
+        - Supported on NVIDIA optimized Linux kernels, including IGX OS and DGX OS.
+        - Supported by all MOFED drivers (requires rebuilding nvidia-dkms drivers afterwards).
+    - [`DMA Buf`](https://docs.kernel.org/driver-api/dma-buf.html), supported on Linux kernels 5.12+ with NVIDIA open-source drivers 515+ and CUDA toolkit 11.7+.
+
+
+## 1. Installing Holoscan Networking
+
+We'll start with installing the `holoscan-networking` package, as it provides some utilities to help tune the system, and requires some dependencies which will help us with the system setup.
+
+=== "IGX OS 1.1"
+
+    !!! tip "COMING SOON - GTC 2025"
+
+    ```bash
+    # DOCA repository: https://developer.nvidia.com/doca-downloads?deployment_platform=Host-Server&deployment_package=DOCA-Host&target_os=Linux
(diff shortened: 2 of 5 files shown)


tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2179 @@
...
+    ```bash
+    cat << EOF | sudo tee /etc/systemd/system/cpu-performance.service
+    [Unit]
+    Description=Set CPU governor to performance
+    After=multi-user.target
+
+    [Service]
+    Type=oneshot
+    ExecStart=/usr/bin/cpupower -c all frequency-set -g performance
+
```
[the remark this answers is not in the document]
agirault: That's wrong, `-c all` works too, and lets users understand they could pass a specific set of core ids instead.

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2179 @@
...
+    [Install]
+    WantedBy=multi-user.target
+    EOF
+    sudo systemctl enable cpu-performance.service
```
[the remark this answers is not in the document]
agirault: this is just wrong. the command here starts the service

dev_container
```diff
@@ -878,8 +878,7 @@ build_and_package() {
         c_echo "Package ${pkg} not found in list of packages: (see ./run list_packages)"
         exit 1
     fi
-    dockerfile=$(./run get_pkg_dockerfile ${pkg})
-    build_and_run $pkg --no_run --docker_file ${dockerfile} "$@"
+    build_and_run $pkg --no_run "$@"
 }
```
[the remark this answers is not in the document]
agirault: was removed in favor of doing it within `build_and_run` directly, which means `build_and_install $pkg` won't fail.

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2179 @@
...
+12. A descriptive name for that interface, currently only used for logging.
+13. The PCIe/bus address of that interface, as identified in previous sections.
+14. Each interface can have a `tx` (transmitting) or `rx` (receiving) section, or both if you'd like to configure both Tx and Rx on the same interface.
+15. The `queues` section lists the queues for that interface. Queues are a core concept of NICs: they handle the actual receiving or transmitting of network packets. Rx queues buffer incoming packets until they can be processed by the application, while Tx queues hold outgoing packets waiting to be sent on the network. The simplest setup uses only one receive and one transmit queue. Using more queues allows multiple streams of network traffic to be processed in parallel, as each queue can be assigned to a specific CPU core, and are assigned their own memory regions that are not shared.
+16. A descriptive name for that queue, currently only used for logging.
+17. The ID of that queue, which can be referred to later in the `flows` section.
+18. The number of packets per batch. The `advanced_network` Rx and Tx operators will "tick" for each batch.
```
cliffburdick, line 1991: The TX technically doesn't care about this number when transmitting. The user just has to make sure they're not transmitting more than this

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2179 @@
...
+21. Same as for `data1`. Each interface in this list should have a unique mac address. This one will do `rx` per config below.
+22. Whether to isolate the Rx flow. If true, any incoming packets that does not match the MAC address of this interface - or isn't directed to a queue when the `flows` section below is used - will be delegated back to Linux for processing (no kernel bypass). This is useful to let this interface handle ARP, ICMP, etc. Otherwise, any packets sent to this interface (ex: ping) will need to be processed (or dropped) by your application.
+23. `rx` queues have an `output_port` parameter so you can attach a downstream operator to receive data from this specific queue, as can be seen in the `#!cpp Application::compose()` function of the sample application. Multiple `rx` queues can share the same `output_port`. In contrast, `tx` queues have a single non-configurable port (name: `burst_in`) to which upstream operators will send all packets, which are then routed to the correct queue based on #TODO#.
```
cliffburdick, line 1996: based on the port/queue in the burst header

tutorials/high_performance_networking/README.md
agirault: I was wondering.

Why do we even have a limit for Tx? it can't just shoot to the NIC whatever our Tx operator sends it?

tutorials/high_performance_networking/README.md
agirault: the part I still can't grasp, is how is the port/queue defined in the burst header?

I assume port is inferred from the app passing the source mac address (defined below in our config file). Is that correct? and how is queue inferred?

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+24. This scenario is called HDS (Header-Data Split): the packet will first be written to a buffer in the `Data_RX_CPU` memory region, filling its `buf_size` of 64 bytes - which is consistent with the size of our header - then the rest of the packet will be written to the `Data_RX_GPU` memory region. Its `buf_size` of 1000 bytes is just what we need to write the payload size for our application, no byte wasted!
+25. The list of flows. Flows are responsible for routing packets to the correct queue based on various properties. If this field is missing, all packets will be routed to the first queue.
+26. The flow name, currently only used for logging.
+27. The flow `id` is used for #TODO#
```
agirault, line 1992: @cliffburdick help please

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+
+## Background
+
+Achieving high performance networking is a complex problem that involves many system components and configurations which we will cover in this tutorial. Two of the core concepts to achieve this are named Kernel Bypass, and GPUDirect.
+
+### Kernel Bypass
+
+In this context, Kernel Bypass refers to bypassing the operating system's kernel to directly communicate with the network interface (NIC), greatly reducing the latency and overhead of the Linux network stack. There are multiple technologies that achieve this in different fashions. They're all Ethernet-based, and leverage the `ibverbs` API to interact with the NIC, but differ in their implementation and features:
```
cliffburdick, line 26: GPUNetIO doesn't really leverage ibverbs, but instead does the same register poking that ibverbs on the CPU does, but via a GPU API.

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
+## High Performance Networking with Holoscan
+
+This tutorial demonstrates how to use the advanced networking Holoscan operator (often referred to as ANO or `advanced_network` in HoloHub) for low latency and high throughput communication through NVIDIA SmartNICs. With a
```
cliffburdick, line 3: Should we have a leading sentence describing why the ANO exists? The intro mentions what it's for, but it doesn't mention why not just use one of the other libraries directly. It's kind of mentioned below in the managers section, but I wasn't sure if we want to lead with that.

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+
+1. On IGX Orin developer kits, you can switch that setting through the BIOS: [see IGX Orin documentation](https://docs.nvidia.com/igx-orin/user-guide/latest/switch-network-link.html).
+2. On any system with a NVIDIA NIC (including the IGX Orin developer kits), you can run the commands below from a terminal:
+
+    1. Identify the PCI address of your NVIDIA NIC
+
+        === "ibdev2netdev"
+
+            ```bash
+            nic_pci=$(sudo ibdev2netdev -v | awk '{print $1}' | head -n1)
```
cliffburdick, line 208: If we wanted to we can simply leave the device name off mlxconfig and it will set all mellanox interfaces to ethernet.

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+
+        Apply now:
+
+        ```bash
+        sudo systemctl restart systemd-networkd
+        ```
+
+!!! note
+
+    If you are connecting the NIC to another NIC with a [LinkX interconnect](https://www.nvidia.com/en-us/networking/interconnect/), do the same on the other system with an IP address on the same network segment.
```
cliffburdick, line 346: LinkX is a very specific term for something that's just "a cable". Do we want to leave a product-specific comment here when really any compatible cable/xcvr would work?

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+    See <https://enterprise-support.nvidia.com/s/article/understanding-pcie-configuration-for-maximum-performance>
+
+Check current MRRS:
+
+=== "tune_system.py"
+
+    === "Debian installation"
+
+        ```bash
+        sudo /opt/nvidia/holoscan/bin/tune_system.py --check mrrs
```
cliffburdick, line 607: Should we mention here that this value is reset on reboot and needs to be set every time the system boots?

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+        If you work with containers, remember to mount this directory in your container as well with `-v /mnt/huge:/mnt/huge`.
+
+
+Rerunning the initial commands should now list 3 hugepages of 1GB each. 1GB will be the default huge page size if updated in the kernel bootline only.
+
+### 3.5 Isolate CPU cores
+
+!!! note
+
+    This optimization is less impactful when using the `gpunetio` backend since the GPU polls the NIC.
```
cliffburdick, line 822: Do we need to define what isolation means here?

tutorials/high_performance_networking/README.md
cliffburdick: Because the user still has to allocate a burst of data, and that burst is always bound by the maximum number of packets. We could extend the API such that instead they only allocate a smaller number, but I'm not sure if that would be useful.

RX burst are sent either when the number of packets matches the burst size, or when the timer expires.

tutorials/high_performance_networking/README.md
cliffburdick: The flow ID is used to tag the packets with what flow it arrived on. This is useful when sending multiple flows to a single queue, and the user application can differentiate which flow matched the packet based on this ID.

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+      tx: # (14)!
+        queues: # (15)!
+        - name: "ADC Samples" # (16)!
+          id: 0 # (17)!
+          batch_size: 10240 # (18)!
+          cpu_core: 11 # (19)!
+          memory_regions: # (20)!
+            - "Data_TX_GPU"
+          offloads:
+            - "tx_eth_src"
```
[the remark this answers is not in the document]
cliffburdick: 1) IP, UDP, and Ethernet Checksums or CRC are always done by the NIC currently.
2) tx_eth_src has the NIC insert the ethernet source mac address in hardware rather than the software populating it.

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+    Persistence Mode: Enabled
+    ```
+
+We then lock the GPU's clocks to the max values:
+
+
+=== "One-time"
+
+    ```bash
+    sudo nvidia-smi --lock-gpu-clocks=$(nvidia-smi --query-gpu=clocks.max.sm --format=csv,noheader,nounits)
```
cliffburdick, line 1045: Could use the shorthand `-lmc` and `lgc` to make this a little smaller

tutorials/high_performance_networking/README.md
agirault: great

tutorials/high_performance_networking/README.md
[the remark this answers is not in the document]
agirault: no other offload capability than `tx_eth_src` correct?

tutorials/high_performance_networking/README.md
agirault: agreed. I didn't know what term was best. Never heard of XCVR for example. I also heard of QSFP before then saw there was also OSFP. Should I use `SFP` ?

tutorials/high_performance_networking/README.md
agirault: oh :) yes. I guess in the future we could add a flag to the setters for persistence through systemd.

tutorials/high_performance_networking/README.md
agirault: Sounds good. Can you suggest some text?

tutorials/high_performance_networking/README.md
agirault: You know I've not settles on whether to use short or long form in the tutorial. technically you just copy paste it so long is nice and descriptive, but if we want people to remember it to run more often then short is nice (like `nvidia-smi -q` or `nvidia-smi topo -mp`).

Does that make you lean one way vs the other?

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2171 @@
...
+1. Press `Join Now`.
+2. Once approved, download the `Display Mode Selector` archive.
+3. Unzip the archive.
+4. Access your system without a X-server running, either through SSH or a Virtual Console (`Alt+F1`).
+5. Go down the right OS and architecture folder for your system (`linux/aarch64` or `linux/x64`).
+6. Run the `displaymodeselector` command like so:
+
+```bash
+chmod +x displaymodeselector
+sudo ./displaymodeselector --gpumode physical_display_enabled_8GB_bar1
```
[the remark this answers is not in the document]
cliffburdick: I'm not worried about those GPUs listed because they always ship with large BAR1 as far as I know. I just wasn't sure if some of the RTX cards have more options than 8GB.

`I there a use case when you don't want the max on your GPU (apart from the board & BIOS support)?`
Just the BIOS support

tutorials/high_performance_networking/README.md
cliffburdick: "Isolating a CPU in Linux prevents unwanted user or kernel threads from running on the selected CPU. This decreases context switching and avoids noisy neighbor problems that can increase latency in your application."

tutorials/high_performance_networking/README.md
cliffburdick: xcvr is just shorthand for transceiver. I think saying "If you are connecting the NIC to another NIC directly with a cable instead of going through a switch, ..."

tutorials/high_performance_networking/README.md
cliffburdick: oh weird, I know that works on other commands. Thanks

tutorials/high_performance_networking/README.md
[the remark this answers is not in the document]
cliffburdick: Correct

tutorials/high_performance_networking/README.md
cliffburdick: Sounds good

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2172 @@
...
+        )
+        target_include_directories(my_app
+            PRIVATE
+                my_include_dirs/
+                ...
+        )
+        target_link_libraries(my_app
+            PRIVATE
+                holoscan::core
+                holoscan::ops::advanced_network_rx
```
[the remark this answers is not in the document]
agirault: sure but not now

tutorials/high_performance_networking/README.md
```diff
@@ -0,0 +1,2172 @@
...
+
+        ```bash
+        ./applications/adv_networking_bench/cpp/default_bench_op_tx.h
+        ```
+
+!!! note
+
+    Designs investigations are expected soon for a generic way to prepare packets to send to the NIC.
+
+### 5.4 Build with CMake
```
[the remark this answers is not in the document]
agirault: installing `holoscan-networking` is step (1) in the tutorial. It's also required for step (4). Even then, at the top of 5.1, it says how you can find the right files depending on your installation, so I think that's sufficient? I'll add a sentence at the top of section (5)

tutorials/high_performance_networking/README.md
agirault: chose this:

```md
The CPU interacting with the NIC to route packets means that any perturbation can have a large impact on latency, especially with smaller packet/batch sizes requiring more frequent work. Isolating a CPU in Linux prevents unwanted user or kernel threads from running on the selected CPU. This decreases context switching and avoids noisy neighbor problems that can increase latency in your application.

We recommend isolating the CPU cores you will select to interact with the NIC (defined in the `advanced_network` configuration [described later](#51-understand-the-configuration-parameters) in this tutorial). This is done by setting additional flags on the kernel bootline.

Comment (cliffburdick) [reply tutorials/high_performance_networking/README.md:822]: I was thinking 

```
The CPU interacting with the NIC to route packets means that any perturbation can have a large impact on latency, especially with smaller packet/batch sizes requiring more frequent work.

To avoid dropping packets, we recommend isolating the CPU cores that are interacting with the NIC (defined in the advanced_network configuration [described later](https://github.com/nvidia-holoscan/holohub/pull/718#51-understand-the-configuration-parameters) in this tutorial). Isolating a CPU in Linux prevents unwanted user or kernel threads from running on the selected CPU. This decreases context switching and avoids noisy neighbor problems that can increase latency in your application. This is done by setting additional flags on the kernel bootline.
```

cliffburdick (approved):

PR #721, opened 2025-03-10
Fix Docker Build Failure

bhashemian:
A benchmarking dependency installation (`xdot`) caused the `./dev_container build` to fail.
It was due to a recent release of pygobject, which requires `libgirepository-2.0-dev`. This dependency cannot be satisfied in holoscan container, which is based on ubuntu 22.04.

This PR fixes the pygobject version to `3.50.0` to resolve the issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,7 +66,7 @@ RUN apt update \
     graphviz\
     ninja-build
 
-RUN pip install meson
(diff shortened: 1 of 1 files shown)


jjomier:
https://github.com/nvidia-holoscan/holohub/pull/722

PR #722, opened 2025-03-10
Fixing benchmarking requirements for Ubuntu 22.04

jjomier:


jjomier:
> Looks good, could you please mention the reason for pinning in the commit message? Seems like `3.52.1` release was added yesterday: https://pypi.org/project/PyGObject/#history
> 
> I wonder if there is an issue we should report upstream?

 girepository-2.0 does not exist for Ubuntu 22.04 and requirement for xdot. So the only "trick" is to pin an older version of PyGObject that works with xdot on Ubuntu 22.04

tbirdso (approved):

bhashemian:
> > Looks good, could you please mention the reason for pinning in the commit message? Seems like `3.52.1` release was added yesterday: https://pypi.org/project/PyGObject/#history
> > I wonder if there is an issue we should report upstream?
> 
> girepository-2.0 does not exist for Ubuntu 22.04 and requirement for xdot. So the only "trick" is to pin an older version of PyGObject that works with xdot on Ubuntu 22.04

It is actually a bug for `xdot` that have not fixed its dependency verfsion. Maybe it is worth reporting to them if not already done since it can potentially affect all the users on older version of ubuntu.

PR #735, opened 2025-03-13
Fixed lockup with buffer size and added stats reporting

cliffburdick:
1. Added new stats reporting mechanism. Right now this only finds issues on RX, but it will report when the software is not running fast enough. Will improve later.
2. Made clear separation between tx/rx allocations and frees since they may use different buffers underneath
3. Fixed issue where application would hang if the number of TX buffers was too low. This addresses this issue: https://jirasw.nvidia.com/browse/CLARAHOLOS-1740
4. Removed need for extra copy/malloc on TX side. Rivermax still needs to remove this, then the TX operator will be doing almost nothing.


agirault (approved):
ty so much, so useful.

Went over it, updated some logging message, otherwise lgtm.

Would have been nice to have the two separate things in separate commits or MRs.

PR #742, opened 2025-03-14
Update Contribution Guide and Docs for Workflows

bhashemian:
This pull request primarily updates the `CONTRIBUTING.md` file to include guidelines for adding workflows and updates existing sections to reflect the inclusion of workflows. Additionally, a new `README.md` file is added for workflows.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 ## Table of Contents
 - [Introduction](#introduction)
 - [Types of Contributions](#types-of-contributions)
-- [Developer Workflow](#developer-workflow)
+- [Developer Process](#developer-process)
 - [Preparing your submission](#preparing-your-submission)
 - [Reporting issues](#reporting-issues)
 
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -243,6 +246,39 @@ cmake -S . -B ./build -D APP_my_application:BOOL=1
 cmake --build ./build -j
 ```
 
+
+### Adding a Workflow
+
+> NOTE: Workflows are _end-to-end_ reference applications that demonstrate complete pipelines from sensors to results. To determine if your contribution should be a workflow rather than an application, consider whether it represents a complete "from sensor to insight" pipeline. Workflows typically integrate multiple components (operators, models, etc.) to solve a specific use case in a production-ready manner, while applications may focus on demonstrating specific functionality or techniques.
```
tbirdso, line 252: > production-ready manner

We should be careful about "production-ready" claims. Has the workflow gone through QA?

tbirdso (approved):
Looks good, one note on "production readiness" terminology, not blocking

CONTRIBUTING.md
bhashemian: That’s a good point. Will remove it. Thanks

PR #743, opened 2025-03-14
Remove holoscan operators from ANO

cliffburdick:
The ANO RX/TX operators were originally created to make the interface with Holoscan SDK easier to use. They provided a native way for users to send and receive bursts of packets to/from the ANO. 

Over time we found that the TX/RX operators were quite limiting and didn't really need to be there. Specifically:

1. There is a performance hit since the operators must go through a holoscan port and make a copy of the burst structure. In some cases they also called `delete`.
2. They caused some weird startup magic because DPDK requires its init routine to be called once, but we didn't know if the TX or RX operator was starting first. We had to maintain an atomic variable to see which was created first so we didn't call it twice.
3. They created another potential source of packet loss that would have to be tracked (the holoscan port)
4. They could become a bottleneck and very high rates, and changing the design to require multiple TX or RX operators would be challenging.


applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -172,11 +171,12 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
...
 
-    // Get new input burst (ANO batch of packets)
-    auto burst_opt = op_input.receive<BurstParams*>("burst_in");
-    if (!burst_opt) { return; }
-
-    auto burst = burst_opt.value();
+    BurstParams *burst;
+    auto status = get_rx_burst(&burst);
+    if (status != Status::SUCCESS) {
+      HOLOSCAN_LOG_DEBUG("No RX burst available");
```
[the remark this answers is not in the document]
cliffburdick: No, this is a normal condition

applications/adv_networking_bench/cpp/main.cpp
```diff
@@ -32,7 +32,13 @@ class App : public holoscan::Application {
   void compose() override {
     using namespace holoscan;
 
-    HOLOSCAN_LOG_INFO("Initializing advanced network operator");
+    auto adv_net_config = from_config("advanced_network").as<NetworkConfig>();
+    if (advanced_network::adv_net_init(adv_net_config) != advanced_network::Status::SUCCESS) {
+      HOLOSCAN_LOG_ERROR("Failed to initialize advanced network");
+      exit(1);
```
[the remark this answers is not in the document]
cliffburdick: This is an example application and we want it to exit.

operators/advanced_network/CMakeLists.txt
```diff
@@ -18,64 +18,6 @@ cmake_minimum_required(VERSION 3.20)
 # advanced_network_common will be generated and used by the operators below.
```
agirault, line 18:
```suggestion
# advanced_network_common will be generated and used by operators.
```

operators/advanced_network/Dockerfile
```diff
@@ -15,7 +15,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE
+ARG BASE_IMAGE=nvcr.io/nvidia/clara-holoscan/holoscan:v2.9.0-dgpu
```
agirault, line 18: Revert, should let dev_container set it, and latest is 3.1 now

operators/advanced_network/advanced_network/managers/gpunetio/adv_network_doca_mgr.cpp
```diff
@@ -1743,7 +1743,8 @@ std::optional<uint16_t> DocaMgr::get_port_from_ifname(const std::string& name) {
   return -1;
 }
 
-Status DocaMgr::get_rx_burst(BurstParams** burst) {
+Status DocaMgr::get_rx_burst(BurstParams** burst, int port, int q) {
+  // Update to allow multiple ports and queues!
```
agirault, line 1747: I assume that's a TODO? That would be a regression if we don't do this right? We need this for PSD pipeline at the very least.

agirault:
FYI @JohnMoon-VTS @dylan-eustice

dylan-eustice:
[heart]         Dylan Eustice reacted to your message:
________________________________

applications/adv_networking_bench/README.md
```diff
@@ -1,48 +1,54 @@
...
-This is a simple application to measure a lower bound on performance for the advanced networking operator
-by receiving packets, optionally doing work on them, and freeing the buffers. While only freeing the packets is
-an unrealistic workload, it's useful to see at a high level whether the application is able to keep up with
-the bare minimum amount of work to do. The application contains both a transmitter and receiver that are
-designed to run on different systems, and may be configured independently.
+> [!TIP]
+> Review the [High Performance Networking tutorial](/tutorials/high_performance_networking/README.md) for guided
+> instructions to configure your system and test the Advanced Network Library.
+
+This is a sample application to measure a lower bound on performance for the [Advanced Network
```
cliffburdick, line 7: Library is capitalized above but not here

applications/adv_networking_bench/README.md
```diff
@@ -1,48 +1,54 @@
...
 
-The receiver receives the UDP packets in either CPU-only mode or header-data split mode. CPU-only mode
-will receive the packets in CPU memory, copy the payload contents to a host-pinned staging buffer, and
-freed. In header-data split mode the user may configure a split point where the bytes before that point
-are sent to the CPU, and all bytes afterwards are sent to the GPU. Header-data split should achieve higher
-rates than CPU mode since the amount of data to the CPU can be orders of magnitude lower compared to running
-in CPU-only mode.
+The receiver receives the UDP packets in either CPU-only mode, header-data split mode, or GPU-only mode.
+- CPU-only mode will receive the packets in CPU memory, copy the payload contents to a host-pinned staging buffer, and
+  freed.
```
cliffburdick, line 33: "freed" sounds odd here. Maybe "free the buffers"?

agirault: 100%, though that was written before. Not sure why I changed this section.

applications/adv_networking_bench/README.md
```diff
@@ -1,48 +1,54 @@
...
+  rates than CPU mode since the amount of data to the CPU can be orders of magnitude lower compared to running
+  in CPU-only mode.
+- GPU-only mode: all bytes of the packets are received in GPU memory.
 
 ### Configuration
 
 The application is configured using a separate transmit and receive file. The transmit file is called
 `adv_networking_bench_tx.yaml` while the receive is named `adv_networking_bench_rx.yaml`. Configure the
-advanced networking operator on both transmit and receive per the instructions for that operator.
+Advanced Network library on both transmit and receive per the instructions for that library.
```
cliffburdick, line 44: This isn't quite right since there's only one library now, but in the past there were multiple operators. I think we should reword this so it's something like "for the YAML":

applications/adv_networking_bench/README.md
agirault: keeping all lowercase, ty

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
```diff
@@ -78,6 +77,7 @@ advanced_network:
             # ipv4_len: 1050
 
 bench_rx:
+  port_name: "rx_port"    # Name of rx_port from ANO config
```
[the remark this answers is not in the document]
agirault: also, we call this `interfaces` on line 44, can we name this `interface_name` instead ?

operators/advanced_network/README.md
```diff
@@ -43,45 +56,41 @@ The limitations below will be removed in a future release.
...
 
-NVIDIA DOCA brings together a wide range of powerful APIs, libraries, and frameworks for programming and accelerating modern data center infrastructures​. The DOCA SDK composed by a variety of C/C++ API for different purposes​, exposing all the features supported by NVIDIA hardware and platforms. [DOCA GPUNetIO](https://docs.nvidia.com/doca/sdk/doca+gpunetio/index.html) is one of the libraries included in the SDK and it enables the GPU to control, from a CUDA kernel, network communications directly interacting with the network card and completely removing the CPU from the critical data path.
+If the application wants to enable GPU communications, it must chose `gpunetio` as backend. The behaviour of the GPUNetIO backend is similar to the DPDK one except that the receive and send are executed by CUDA kernels. Specifically:
```
cliffburdick, line 84: British behavior

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
```diff
@@ -95,4 +95,4 @@ bench_tx:
   ip_dst_addr: <5.6.7.8>  # Destination IP address - required on layer 3 network
   udp_src_port: 4096      # UDP source port
   udp_dst_port: 4096      # UDP destination port
-  address: <0000:00:00.0> # Source NIC Bus ID. Should match the address of the Tx interface above
+  address: <0000:00:00.0> # Source NIC Bus ID. Should match the address of the Tx interface above
```
agirault, line 98: so in a way this `address` is the equivalent of `port_name` (which I think we should rename `interface`) in the Rx operator? If yes, should we do the same?

operators/advanced_network/README.md
```diff
@@ -412,81 +341,56 @@ Too low means risk of dropped packets from NIC having nowhere to write (Rx) or h
...
 - **`queues`**: List of queues on NIC
 	type: `list`
 	full path: `cfg\interfaces\tx\queues`
 	- **`name`**: Name of queue
   		- type: `string`
- 	- **`id`**: Integer ID used for flow connection or lookup in operator compute method
+	- **`id`**: Integer ID used for flow connection or lookup in operator compute method
   		- type: `integer`
-	- **`cpu_core`**: CPU core ID. Should be isolated when CPU polls the NIC for best performance.. <mark>Not in use for Doca GPUNet IORivermax manager</mark>
+	- **`cpu_core`**: CPU core ID. Should be isolated when CPU polls the NIC for best performance.. <mark>Not in use for Doca GPUNetIO</mark>
```
cliffburdick, line 354: Not related to this review but I thought this would still apply. Even with GPUNetIO you need the CPU waiting to see if packets arrived so it could do something.

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
cliffburdick: Address was intended to be the PCIe address, linux network name, or mlx5 name of the device. the name was supposed to be a separate name to describe the interface.

operators/advanced_network/README.md
agirault: idk. let's discuss offline with elena. she mentioned in the past some things to not about the cpu core (like it would ignore if you give it different values?)

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
agirault: right, but the goal of this `address` field here is the same as the `port_name` field for rx, correct? which is to simply know what port/interface to send from/receive to.

Seems like we should have a consistent interface for rx and tx to define which port to interact with, either the custom name from the config, of the device system info (address, linux, mlx5). I like the former in this case, i.e. leave device system info to the advanced_network config, and the app operators just need info for headers and to know how to talk to advanced_network, so use the advanced_network interface name

applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -69,6 +69,15 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
...
     holoscan::Operator::initialize();
 
+    port_id_ = address_to_port(port_name_.get());
+    if (port_id_ == -1) {
+      HOLOSCAN_LOG_ERROR("Invalid RX port {} specified in the config", port_name_.get());
+      exit(1);
+    }
+
+    queue_ids_.fill(-1);
+    queue_ids_[0] = 0;  // Only poll one queue for now
```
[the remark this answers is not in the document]
cliffburdick: This is entirely up to the user app. They could loop through queues with about the same amount of code as it would take to check each queue coming from the old method.

applications/adv_networking_bench/cpp/doca_bench_op_rx.h
```diff
@@ -110,12 +109,12 @@ class AdvNetworkingBenchDocaRxOp : public Operator {
     free_batch_queue();
 
     // Get new input burst (ANO batch of packets)
-    auto burst_opt = op_input.receive<BurstParams*>("burst_in");
-    if (!burst_opt) {
-      HOLOSCAN_LOG_ERROR("No burst input");
+    BurstParams *burst;
+    auto status = get_rx_burst(&burst, 0, 0);
```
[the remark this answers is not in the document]
cliffburdick: I will change the port ID to be dynamic, but the queue is 0 in this example (per the config)

applications/adv_networking_bench/cpp/default_bench_op_rx.h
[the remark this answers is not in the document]
agirault: Per our discussion, we can either do that to make it generic-ish, or put a comment for now that makes it ultra clear this is specific to align with the yaml config.

But if we do the latter, I would get rid of `queue_ids_` altogether in this example, and just pass 0.

applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -358,15 +372,17 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
...
   // Holds burst buffers that cannot be freed yet and CUDA event indicating when they can be freed
   struct BatchAggregationParams {
-    std::array<BurstParams*, MAX_ANO_BURSTS> bursts;
+    std::array<BurstParams*, MAX_BURSTS_PER_BATCH> bursts;
     int num_bursts;
     cudaEvent_t evt;
   };
 
+  int port_id_;                                    // Port ID to poll on
+  std::array<int, 16> queue_ids_;                  // Queue IDs to poll on
```
agirault, line 385: if you want to stick to hardcoded for queue 0, let's just remove it. If we keep it, let's at least loop through all the non -1 queues?

applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -69,6 +69,15 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
     HOLOSCAN_LOG_INFO("AdvNetworkingBenchDefaultRxOp::initialize()");
     holoscan::Operator::initialize();
 
+    port_id_ = address_to_port(port_name_.get());
+    if (port_id_ == -1) {
+      HOLOSCAN_LOG_ERROR("Invalid RX port {} specified in the config", port_name_.get());
+      exit(1);
```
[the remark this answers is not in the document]
cliffburdick: Good question. I don't know enough about holoscan to exit gracefully from the init function. Returning will crash harder.

applications/adv_networking_bench/cpp/default_bench_op_rx.h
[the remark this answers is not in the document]
cliffburdick: Added a comment to clarify

applications/psd_pipeline/README.md
```diff
@@ -111,7 +111,6 @@ ID 1, we see:
                 id: 1
                 cpu_core: 5
                 batch_size: 12500
-                output_port: "bench_rx_out"
```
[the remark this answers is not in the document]
cliffburdick: Done

applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -367,6 +381,8 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
     cudaEvent_t evt;
   };
 
+  int port_id_;                                    // Port ID to poll on
+  std::array<int, 16> queue_ids_;                  // Queue IDs to poll on
```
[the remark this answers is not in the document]
cliffburdick: Port or queues?

operators/advanced_network/advanced_network/managers/gpunetio/adv_network_doca_mgr.cpp
cliffburdick: The GPUNetIO backend hasn't supported multiple ports or queues yet, so it's not a regression.

operators/advanced_network/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
```diff
@@ -155,6 +155,7 @@ bool DpdkMgr::set_config_and_initialize(const NetworkConfig& cfg) {
 int DpdkMgr::address_to_port(const std::string& addr) {
   for (const auto& intf : cfg_.ifs_) {
     if (intf.address_ == addr) { return intf.port_id_; }
+    if (intf.name_ == addr) { return intf.port_id_; }
```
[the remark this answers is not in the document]
cliffburdick: Is this one outdated from your updates?

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
cliffburdick: Good point. will update

applications/adv_networking_bench/cpp/default_bench_op_rx.h
cliffburdick: Switched this to loop over available queues

applications/adv_networking_bench/cpp/CMakeLists.txt
```diff
@@ -38,8 +38,7 @@ add_executable(adv_networking_bench
 target_link_libraries(adv_networking_bench
   PRIVATE
   holoscan::core
-  holoscan::ops::advanced_network_rx
```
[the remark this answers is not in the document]
cliffburdick: Done

operators/advanced_network/advanced_network/managers/gpunetio/adv_network_doca_mgr.cpp
agirault: ty, hadn't caught that.

applications/adv_networking_bench/cpp/default_bench_op_rx.h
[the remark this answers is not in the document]
agirault: queue ids. port_id is inferred from `Parameter<std::string> port_name_` already.

applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -186,185 +183,190 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
...
-      return;
-    }
 
-    auto burst_size = get_num_packets(burst);
+    const auto num_rx_queues = get_num_rx_queues(port_id_);
+    for (int q = 0; q < num_rx_queues; q++) {
 
-    // Count packets received
-    ttl_pkts_recv_ += burst_size;
+      auto status = get_rx_burst(&burst, port_id_, q);
```
agirault, line 189: should we skip if q id is -1 ?

agirault: Nevermind, just saw you removed queue_ids_

applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -186,185 +183,190 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
...
 
     BurstParams *burst;
-    auto status = get_rx_burst(&burst, port_id_, queue_ids_[0]);
-    if (status != Status::SUCCESS) {
-      HOLOSCAN_LOG_DEBUG("No RX burst available");
-      return;
-    }
 
-    auto burst_size = get_num_packets(burst);
+    const auto num_rx_queues = get_num_rx_queues(port_id_);
```
agirault, line 187: Can we add a comment above similar to this?

```suggestion
    // In this example, we'll loop through all the rx queues of the interface
    // assuming we want to process the packets the same way for all queues
    const auto num_rx_queues = get_num_rx_queues(port_id_);
```

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -226,14 +226,11 @@ void Vita49ConnectorOpRx::compute(
         InputContext& op_input,
         OutputContext& op_output,
         ExecutionContext& context) {
-  auto burst_opt = op_input.receive<in_t>("in");
-  if (!burst_opt) {
-    return;
-  }
-  auto burst = burst_opt.value();
+  BurstParams *burst;
+  auto status = get_rx_burst(&burst, 0, CONTEXT_QUEUE_ID);
```
agirault, line 230: we should call `address_to_port` here as well to retrieve the port id given an interface name, since we lost the holoscan output port connection

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -287,8 +284,15 @@ void Vita49ConnectorOpRx::compute(
   }
 
   // Assumes that channel 0 is queue 1, channel 1 is queue 2, etc.
-  uint16_t channel_num = get_q_id(burst) - 1;
-  process_channel_data(op_output, burst, channel_num);
+  const auto num_rx_queues = get_num_rx_queues(0);
```
agirault, line 287: and use that id here

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -287,8 +284,15 @@ void Vita49ConnectorOpRx::compute(
...
+  for (uint16_t q = CONTEXT_QUEUE_ID + 1; q < num_rx_queues; q++) {
+
+    auto status = get_rx_burst(&burst, 0, q);
```
agirault, line 289: and here

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -226,14 +226,11 @@ void Vita49ConnectorOpRx::compute(
         InputContext& op_input,
         OutputContext& op_output,
         ExecutionContext& context) {
-  auto burst_opt = op_input.receive<in_t>("in");
-  if (!burst_opt) {
-    return;
-  }
-  auto burst = burst_opt.value();
+  BurstParams *burst;
```
agirault, line 229:
```suggestion
  // Check for a context packet first
  BurstParams *burst;
```

applications/adv_networking_bench/cpp/default_bench_op_rx.h
```diff
@@ -394,7 +395,7 @@ class AdvNetworkingBenchDefaultRxOp : public Operator {
   std::array<void**, num_concurrent> h_dev_ptrs_;  // Host-pinned list of device pointers
   std::array<void*, num_concurrent> full_batch_data_d_;  // Device aggregated batch
   std::array<void*, num_concurrent> full_batch_data_h_;  // Host aggregated batch
-  Parameter<std::string> port_name_;                     // Port name from advanced_network config
+  Parameter<std::string> interface_name_;                     // Port name from advanced_network config
```
agirault:
```suggestion
  Parameter<std::string> interface_name_;                // Port name from advanced_network config
```

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
```diff
@@ -77,14 +77,15 @@ advanced_network:
...
-  port_name: "rx_port"    # Name of rx_port from the advanced_network config
+  interface_name: "rx_port"    # Name of rx_port from the advanced_network config
   gpu_direct: true        # Set to true if using a GPU region for the Rx queues.
   split_boundary: false   # Whether header and data is split (Header to CPU)
   batch_size: 10240
   max_packet_size: 1064
   header_size: 64
 
 bench_tx:
+  interface_name: "tx_port"    # Name of tx_port from the advanced_network config
```
agirault, line 88: Thank you that's great.

We need to update the docs/tutorials accordingly. Happy to do it myself

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
cliffburdick: They don't specify the interface name in their config and always assume 1 port, which is why I did it this way.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
cliffburdick: See above

operators/advanced_network/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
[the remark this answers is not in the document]
agirault: idk. but we can rename in another MR to minimize changes, up to you.

operators/advanced_network/advanced_network/manager.h
```diff
@@ -77,6 +77,7 @@ class Manager {
   virtual Status get_mac_addr(int port, char* mac) = 0;
   virtual int address_to_port(const std::string& addr) final;  // NOLINT(readability/inheritance)
   virtual bool validate_config() const;
+  virtual uint16_t get_num_rx_queues(int port_id) const;
```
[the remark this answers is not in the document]
cliffburdick: This was intentionally not final so that another operator could override it if they wanted more internal queues for management vs what was in the config.

applications/adv_networking_bench/cpp/doca_bench_op_rx.h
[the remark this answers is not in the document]
agirault: Change still needed right? for port id.

And queue is 0 is the same for the dpdk backend where we still get the number of queues and loop through them?

operators/advanced_network/advanced_network/managers/dpdk/adv_network_dpdk_mgr.h
```diff
@@ -222,6 +225,7 @@ class DpdkMgr : public Manager {
   std::array<std::string, MAX_IFS> if_names;
   std::array<std::string, MAX_IFS> pcie_addrs;
   std::array<struct rte_ether_addr, MAX_IFS> mac_addrs;
+  RxRings rx_rings_;
   struct rte_ether_addr conf_ports_eth_addr[RTE_MAX_ETHPORTS];
   struct rte_ring* rx_ring;
```
agirault, line 230: Should this go away?

cliffburdick: Removed the singular version

operators/advanced_network/advanced_network/managers/dpdk/adv_network_dpdk_mgr.h
```diff
@@ -137,6 +137,8 @@ class DpdkLogLevelCommandBuilder : public ManagerLogLevelCommandBuilder {
   DpdkLogLevel::Level level_;  ///< The DPDK log level.
 };
 
+using RxRings = std::array<std::array<struct rte_ring*, MAX_NUM_RX_QUEUES>, MAX_INTERFACES>;
```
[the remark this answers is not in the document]
cliffburdick: This is specific to the DPDK implementation and it's guaranteed that they are continuous starting at 0

operators/advanced_network/advanced_network/manager.h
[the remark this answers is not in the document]
agirault: want to comment that in the implementation code?

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
agirault: but with dpdk we know the port could not be 0 depending on which one you use from your NIC, no?

operators/advanced_network/advanced_network/managers/dpdk/adv_network_dpdk_mgr.cpp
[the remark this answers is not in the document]
cliffburdick: let's do that. this one is getting massive

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
cliffburdick: DPDK always starts at 0 regardless of which port you have in the config

operators/advanced_network/advanced_network/managers/dpdk/adv_network_dpdk_mgr.h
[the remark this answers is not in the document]
cliffburdick: You cannot have a port ID of 1 and not 0 since DPDK auto-assigns the numbers when it scans. Queues are a bit different and you can technically give them non-contiguous numbers, but it will likely have all kinds of problems down the line. This is something we should check in `validate_config` since there's no reason to do that.

operators/advanced_network/advanced_network/manager.h
```diff
@@ -69,16 +69,15 @@ class Manager {
   virtual void print_stats() = 0;
   virtual uint64_t get_burst_tot_byte(BurstParams* burst) = 0;
   virtual BurstParams* create_tx_burst_params() = 0;
-
-  /* Internal functions used by ANO operators */
-  virtual Status get_rx_burst(BurstParams** burst) = 0;
+  virtual Status get_rx_burst(BurstParams** burst, int port, int q) = 0;
```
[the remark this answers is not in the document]
cliffburdick: @ronyrad do you mind taking a look at the implementation here? It looks like there's a single queue here for all bursts, but we'd want to separate this per queue.

operators/advanced_network/advanced_network/managers/dpdk/adv_network_dpdk_mgr.h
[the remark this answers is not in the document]
agirault: > You cannot have a port ID of 1 and not 0 since DPDK auto-assigns the numbers when it scans. 


Gotcha, misunderstood where the id came from, thought it was from the NIC.

>  Queues are a bit different and you can technically give them non-contiguous numbers, but it will likely have all kinds of problems down the line. 

is there a cost of doing unordered_map?

> This is something we should check in validate_config since there's no reason to do that.

Instead, would it make more sense to have names to identify queues (just like we do for ports, memory regions) in our user interface, and we internally assign contiguous ids for the queues (just like we do for ports)?

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
```diff
@@ -95,4 +96,3 @@ bench_tx:
   ip_dst_addr: <5.6.7.8>  # Destination IP address - required on layer 3 network
   udp_src_port: 4096      # UDP source port
   udp_dst_port: 4096      # UDP destination port
-  address: <0000:00:00.0> # Source NIC Bus ID. Should match the address of the Tx interface above
```
agirault, line 98: Still need to update all other configs.

Also need to update the `##### Configure the application` section in the tutorial.

I'm happy to do it.

ronyrad (commented):
> @ronyrad do you mind taking a look at the implementation here? It looks like there's a single queue here for all bursts, but we'd want to separate this per queue. 

I have a concern about this change. While adding port/queue parameters enables direct queue access, it risks enforcing a rigid round-robin pattern that could disrupt natural traffic distribution ratios between queues. In scenarios where queues receive packets at different rates (e.g., 5:1 ratio), mandatory per-queue polling would artificially equalize processing of inherently unequal flows and force empty queue checks  and might delay packets in busy queues.

operators/advanced_network/advanced_network/manager.h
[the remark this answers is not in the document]
ronyrad: I have a concern about this change. While adding port/queue parameters enables direct queue access, it risks enforcing a rigid round-robin pattern that could disrupt natural traffic distribution ratios between queues. In scenarios where queues receive packets at different rates (e.g., 5:1 ratio), mandatory per-queue polling would artificially equalize processing of inherently unequal flows and force empty queue checks and might delay packets in busy queues.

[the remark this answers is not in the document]
agirault: >  mandatory per-queue polling 

I might be wrong, but the api doesn't seem to enforce that? If we know

Or is the recommendation to do per port only check only, and then check the queue for the bursts received from the burst itself?

Ex:
```cpp
auto success = get_rx_burst(burst, port_id);
if (status != Status::SUCCESS) {
  HOLOSCAN_LOG_DEBUG("No RX burst available");
  return;
}
switch (burst.queue_id) {
 ...
```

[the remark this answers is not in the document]
cliffburdick: I agree that we can have a port override for the function where it doesn't take a queue, and maybe even another override where it looks at all ports and queues. These would be increasingly slower the less you specify. @ronyrad I think your concern is something like RSS where the queue doesn't matter as much, and that's fair, although we don't support that right now. There are definitely use cases for all 3 of these, so we can just make overrides for each.

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
cliffburdick: Sure, if you can that would be helpful.

cliffburdick:
> I have two general questions
> 
> 1. Should the code move to some other location (after it is no operator anymore).
> 2. Don't we want to switch to C++ API.

1) This is eventually the plan, but it's staying in holohub/operators right now
2) What do you mean? Did you mean getting a pointer to the ANO manager and just calling functions from that?

applications/adv_networking_bench/adv_networking_bench_default_tx_rx.yaml
agirault: if you wanted to review: https://github.com/nvidia-holoscan/holohub/compare/1cb11f265ddfbaf65376e830dad399c31b8cb15b..fc4603c739fde3a1c9faddae15d12c55bb7983b5

operators/advanced_network/advanced_network/managers/dpdk/adv_network_dpdk_mgr.h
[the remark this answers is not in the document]
agirault: so it seems for this MR to be safe, we either need to update `validate_config` for queue ids, or switch the inner array (queue level) to a `unordered_map`.

In contrast, Tx has underdered_map with single key: `(intf.port_id_ << 16) | q.common_.id_`

operators/advanced_network/advanced_network/manager.h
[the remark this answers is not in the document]
agirault: Should we in this MR? Otherwise we need to update the doca and rivermax managers accordingly to support that API also in this MR.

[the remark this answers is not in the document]
ronyrad: I will explain. Let's say we have and application that wants to dispatch bursts arriving from multiple queues to different output ports. (a mapping from port/queue to output). In that case it is more efficient that get burst without specifying in the API which port/queue id to use and to read it from burst header and dispatch. For that case we can have one output queue (should be configurable?). But I agree that there application might want to pull from specific queue. The question if we want to add a dedicated config for that. Rivermax manager can support both options.

ronyrad:
> 2\. to



> > I have two general questions
> > 
> > 1. Should the code move to some other location (after it is no operator anymore).
> > 2. Don't we want to switch to C++ API.
> 
> 1. This is eventually the plan, but it's staying in holohub/operators right now
> 2. What do you mean? Did you mean getting a pointer to the ANO manager and just calling functions from that?

Yes A class. Almost all C functions are forwarding a call to the manager. Thus with a small change it can be a top level C++ API

cliffburdick:
> > 2. to
> 
> > > I have two general questions
> > > 
> > > 1. Should the code move to some other location (after it is no operator anymore).
> > > 2. Don't we want to switch to C++ API.
> > 
> > 
> > 
> > 1. This is eventually the plan, but it's staying in holohub/operators right now
> > 2. What do you mean? Did you mean getting a pointer to the ANO manager and just calling functions from that?
> 
> Yes A class. Almost all C functions are forwarding a call to the manager. Thus with a small change it can be a top level C++ API

I did that exact thing here: https://github.com/nvidia-holoscan/holohub/compare/main...cliffburdick:holohub:direct_access

However, I did not see any improvement in performance over the C API. I thought I would, but the shared_ptr removal caused the only performance increase.

agirault:
Seems like these two discussions are out of scope for this MR.

Only discussions I think need to be addressed before merging:
1. [get_rx_burst per queue](https://github.com/nvidia-holoscan/holohub/pull/743#discussion_r2037537221): support in rivermax and gpunetio managers if current design acceptable
2. [ensure valid queue ids](https://github.com/nvidia-holoscan/holohub/pull/743#discussion_r2038208308): either validate config, or use map like Tx Rings.

operators/advanced_network/advanced_network/manager.h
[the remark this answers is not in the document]
ronyrad: I agree, let's either have overload. Or in a case -1 is passed as queue id, each manager will interpret it either as error or request for "any queue"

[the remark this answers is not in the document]
cliffburdick: Let's still with overloads to keep it more C++-ish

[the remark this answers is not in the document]
agirault: for the overload, is there a concern of which port/queue are polled first? If I have queue 0 and queue 1, with queue 0 getting a lot of packets continuously vs queue 1 rarely, a simple loop would only pick bursts from queue 0 and maybe never get to check queue 1, unless we return a list of bursts and not a single burst. But maybe that was already a concern before? Not sure how the single ring with multiple queues was prioritizing distribution of bursts from separate queues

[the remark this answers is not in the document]
cliffburdick: We talked about this on slack, but keeping an iterator of the last place polled would ensure fairness. If they don't want fairness they can use the port/queue directly.

operators/advanced_network/advanced_network/manager.cpp
```diff
@@ -260,4 +260,67 @@ uint16_t Manager::get_num_rx_queues(int port_id) const {
...
+Status Manager::get_rx_burst(BurstParams** burst, int port_id) {
+  // Check if the port_id is valid
+  if (port_id < 0 || port_id >= static_cast<int>(cfg_.ifs_.size())) {
+    HOLOSCAN_LOG_ERROR("Invalid port_id {} provided to get_rx_burst", port_id);
+    return Status::INVALID_PARAMETER;
+  }
+
+  const auto& queues = cfg_.ifs_[port_id].rx_.queues_;
+  if (queues.empty()) {
+    // No queues configured for this port
```
agirault, line 272: should I log error for requesting a port with no rx queue?

operators/advanced_network/advanced_network/manager.cpp
```diff
@@ -260,4 +260,67 @@ uint16_t Manager::get_num_rx_queues(int port_id) const {
...
+    }
+  }
+
+  // If we checked all queues and none had data
+  return Status::NULL_PTR;
+}
+
+Status Manager::get_rx_burst(BurstParams** burst) {
+  if (cfg_.ifs_.empty()) {
+    // No interfaces configured
```
agirault, line 303: same as above

operators/advanced_network/advanced_network/manager.h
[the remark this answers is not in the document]
agirault: attempted in https://github.com/nvidia-holoscan/holohub/pull/743/commits/e91ae94f04eb9d70246ecebcc965e3ed456219e3

operators/advanced_network/advanced_network/manager.cpp
cliffburdick: I think that's a good idea

PR #747, opened 2025-03-18
Rename platform names in metadata

agirault:
arm64 -> aarch64
amd64 -> x86_64


tbirdso (approved):
Thanks @agirault ! LGTM.

Follows internal alignment to `x86_64` and `aarch64` architecture naming across Holoscan SDK docs. May update fields from `platforms` to `architectures` in the future.

PR #748, opened 2025-03-18
Minor infra fixes

agirault:
- Fix `run:get_app_dockerfile` to not ignore language parameter (Regression introduced in #712)
- Update `CONTRIBUTING.md` to reflect new platform name conventions
- Call linting tools prefixed with `python3 -m` when appropriate, to circumvent scenarios where tools installation paths are not in the PATH


tbirdso (approved):
Thank you @agirault !

PR #756, opened 2025-03-23
Fix tao peoplenet app

wendell-hom:


jjomier (approved):

tbirdso:
@whom3  is metadata disabled to address a runtime issue? Or is disabling optional to reflect that the app does not rely on metadata?

wendell-hom:
> @whom3 is metadata disabled to address a runtime issue? Or is disabling optional to reflect that the app does not rely on metadata?

The app does not rely on metadata, but also it's being disabled here to address the runtime issue.

tbirdso:
@whom3  could you please share the runtime issue either here or internally?

tbirdso:
Merging following internal discussion. Metadata conflicts from V4L2 to be more broadly addressed in the upcoming v3.1 release.

PR #759, opened 2025-03-25
ANO: fix port id set

agirault:
# Context

Port Id logic was changed in #715. Some issues came up:
- dpdk checks port ids twice with `rte_eth_dev_get_port_by_name`
- gpunetio and rivermax backends don't have initialized port ids
- identified redundancy between `address_to_port` and `get_port_from_ifname`
- `address_to_port` is generic to all managers and shouldntt get overriden

# Changes

- [x] DPDK: remove duplicated `rte_eth_dev_get_port_by_name`
- [x] Rivermax: initialize port IDs
- [x] GPUNetIO: initialize port IDs
- [x] Consolidate `address_to_port` in base class
- [x] Remove `get_port_from_ifname` 

# Results

- [x] DPDK tests still pass (apart from jumbo frames, failing already on `main`)

```bash
./dev_container launch \
  --img holohub:adv_networking_bench \
  --docker_opts "-u 0 --privileged" \
  -- bash -c "cd build/adv_networking_bench/applications/adv_networking_bench/cpp/ && python3 -m pytest -v -k 'dpdk' "
```
```pytest
==================================================================================== test session starts =====================================================================================
platform linux -- Python 3.10.12, pytest-8.3.5, pluggy-1.5.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /workspace/holohub
configfile: pyproject.toml
collected 10 items / 6 deselected / 4 selected        

testing/test_ano_bench.py::test_multi_if_loopback[dpdk-64-6.0-0.1-0.0] PASSED                                                                                                          [ 25%]
testing/test_ano_bench.py::test_multi_if_loopback[dpdk-512-55.0-0.1-0.0] PASSED                                                                                                        [ 50%]
testing/test_ano_bench.py::test_multi_if_loopback[dpdk-1500-94.0-0.1-0.0] PASSED                                                                                                       [ 75%]
testing/test_ano_bench.py::test_multi_if_loopback[dpdk-9000-96.0-0.1-0.0] FAILED                                                                                                       [100%]
```

### DIFF
--- a/operators/advanced_network/advanced_network/manager.cpp
+++ b/operators/advanced_network/advanced_network/manager.cpp
@@ -192,6 +192,22 @@ Status Manager::allocate_memory_regions() {
   return Status::SUCCESS;
 }
 
+/**
+ * @brief Generic implementation of address_to_port that looks up port in config
+ * This is a final method that cannot be overridden by subclasses.
+ *
+ * @param addr Address string to look up
+ * @return int Port ID or -1 if not found
+ */
+int Manager::address_to_port(const std::string& addr) {
+  for (const auto& intf : cfg_.ifs_) {
+    if (intf.address_ == addr) {
+      return intf.port_id_;
+    }
+  }
+  return -1;
+}
+
 bool Manager::validate_config() const {
   bool pass = true;
   std::set<std::string> mr_names;
--- a/operators/advanced_network/CHANGELOG.md
+++ b/operators/advanced_network/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Breaking
 
--
+- Removed `get_port_from_ifname`. Use `address_to_port` instead.
 
 ### Non-Breaking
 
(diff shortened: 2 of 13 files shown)


operators/advanced_network/advanced_network/manager.cpp
```diff
@@ -192,6 +192,22 @@ Status Manager::allocate_memory_regions() {
...
 }
 
+/**
+ * @brief Generic implementation of address_to_port that looks up port in config
+ * This is a final method that cannot be overridden by subclasses.
+ *
+ * @param addr Address string to look up
+ * @return int Port ID or -1 if not found
+ */
+int Manager::address_to_port(const std::string& addr) {
```
cliffburdick, line 202: I don't know if we want a more generic name for this or have more functions. This could be called `interface_to_port` where we take in mlx5, linux names, pcie, or config name. As it is right now this only means PCIe address to me.

agirault: right, that's a fair idea, but I think out of scope: here we keep the current method (which takes the pcie address). We just implement it once for all managers instead of specific to each of them

PR #760, opened 2025-03-25
Fix Rivermax build following ANO file structure refactor

ronyrad:
- Update include path
- Remove _.tar.gz_ from _.dockerignore_ , (Rivermax Docker copies tar.gz during docker build).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -23,7 +23,6 @@ ci/
 *.so
 *.deb
 *.rpm
-*.tar.gz
 *.tar.xz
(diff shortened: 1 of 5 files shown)


agirault:
@ronyrad seems like there is a whitespace in the commit description before the `Signed-off-by` which makes the DCO fail?

agirault (approved):
@ronyrad seems like there is a whitespace in the commit description before the Signed-off-by which makes the DCO fail?

Otherwise LGTM

PR #763, opened 2025-03-25
Prevent returning multiple IDs from get_group_id()

tbensonatl:
Grepping /etc/group for a group name can match multiple entries. For example, with the following entries in /etc/group:

video:x:GID1:...
video-data:x:GID2:...

`grep video /etc/group` will match both lines. Append a colon to the search group name to prevent these extra matches. In case there are still multiple matches, add head -1 to keep only the first entry.

### DIFF
--- a/dev_container
+++ b/dev_container
@@ -256,7 +256,7 @@ get_group_id() {
     # a single ID. Thus, append a colon to the group name to match the exact
     # group (e.g., video: and not video-users:) and include head -1 to get a
     # single ID if there are still multiple matches.
-    cat /etc/group | grep $group: | cut -d: -f3 | head -1
+    cat /etc/group | grep ^${group}: | cut -d: -f3 | head -1
 }
 
 get_ucx_options() {
--- a/dev_container
+++ b/dev_container
@@ -251,7 +251,12 @@ run_docker() {
 
 get_group_id() {
     local group=$1
-    cat /etc/group | grep $group | cut -d: -f3
+    # It it possible to match multiple lines in /etc/group for a given
+    # search pattern, but the expectation is that this function returns
+    # a single ID. Thus, append a colon to the group name to match the exact
+    # group (e.g., video: and not video-users:) and include head -1 to get a
+    # single ID if there are still multiple matches.
+    cat /etc/group | grep $group: | cut -d: -f3 | head -1
 }
 
 get_ucx_options() {
--- a/dev_container
+++ b/dev_container
@@ -256,7 +256,7 @@ get_group_id() {
     # a single ID. Thus, append a colon to the group name to match the exact
     # group (e.g., video: and not video-users:) and include head -1 to get a
     # single ID if there are still multiple matches.
-    cat /etc/group | grep ^${group}: | cut -d: -f3 | head -1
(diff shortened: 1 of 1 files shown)


dev_container
```diff
@@ -251,7 +251,12 @@ run_docker() {
 
 get_group_id() {
     local group=$1
-    cat /etc/group | grep $group | cut -d: -f3
+    # It it possible to match multiple lines in /etc/group for a given
+    # search pattern, but the expectation is that this function returns
+    # a single ID. Thus, append a colon to the group name to match the exact
+    # group (e.g., video: and not video-users:) and include head -1 to get a
+    # single ID if there are still multiple matches.
+    cat /etc/group | grep $group: | cut -d: -f3 | head -1
```
agirault, line 259: thanks! Using `:` as closing make a lot of sense, seems we could push it further with ensuring we have no prefix also like so?

```suggestion
    grep "^${group}:" /etc/group | cut -d: -f3 | head -1
```

dev_container
```diff
@@ -251,7 +251,12 @@ run_docker() {
 
 get_group_id() {
     local group=$1
-    cat /etc/group | grep $group | cut -d: -f3
+    # It it possible to match multiple lines in /etc/group for a given
+    # search pattern, but the expectation is that this function returns
+    # a single ID. Thus, append a colon to the group name to match the exact
+    # group (e.g., video: and not video-users:) and include head -1 to get a
+    # single ID if there are still multiple matches.
```
agirault, line 258: Why it's safe and won't hurt to pipe to `head -1`, also curious to know how you got multiple matches. I assume group names should be unique? and they all seemed to match to the same `44` value in the log you shared, was it multiple entries for the same group? Just curious.

dev_container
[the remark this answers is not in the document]
agirault: Gemini is actually correct for once. We have `set -e` globally so I initially thought we'd be fine, but the error is ignored in pipes unless you use `set -o pipefail`. 

I don't think we want to handle it globally without risking to break a bunch of other things in this script.

@tbensonatl your MR already provides a net benefit and self-contained scope so the suggestion can be ignored here, unless you felt like picking it up.

cc @tbirdso for general visibility of error handling in piped commands, though less relevant with the refactoring work in the python CLI.

dev_container
tbensonatl: There were multiple entries with suffixes like video-1, video-2, etc. and the same GID as the video group. I assume this was done to prevent having too many users from being listed in a single entry. One mechanism is the MAX_MEMBERS_PER_GROUP field in login.defs (https://www.man7.org/linux/man-pages/man5/login.defs.5.html).

dev_container
tbensonatl: Sure, I can add that. We could also use getent to get the information more directly, but I don't know if that is more universally available than grep. Gemini suggests adding a default return value as well, but I don't know of a reasonable default.

tbensonatl: Updated to match the full group name (exclude prefixes) and switch directly to using grep as Gemini suggested (rather than cat file | grep).

agirault (approved):

PR #764, opened 2025-03-26
Refactor volume_rendering_xr and xr_hello_holoscan to move tight coupling operators into applications

mimiliaogo:
This PR is the first step of our holoscan-xr migration. We'll refactor these two apps once we have migrated/developed a more general use of xr operators. 

1. Move XrFrameOp and XrTransformOp from `operators/` into `volume_rendering_xr` application subfolders to highlight tight coupling among implementations.
2. Similar process on `xr_hello_holoscan`


tbirdso:
Thank you @mimiliaogo ! I suggest we remove duplicate XR operators from `xr_hello_holoscan` and reuse the implementation in `volume_rendering_xr` with the infra update at: https://github.com/nvidia-holoscan/holohub/pull/767

Suggest the following steps:
1. (here) Migrate app-specific XR operators here to `volume_rendering_xr`, referenced by `xr_hello_holoscan`
2. (planned) Add updated, generalized XR operators to HoloHub and update `xr_hello_holoscan` to use that impl
3. (if time allows) Update XR Volume Rendering to use updated, generalized XR operators

applications/volume_rendering_xr/main.cpp
```diff
@@ -15,12 +15,12 @@
  * limitations under the License.
  */
 
-#include "convert_depth_to_screen_space_op.hpp"
+#include "operators/XrFrameOp/convert_depth/convert_depth_to_screen_space_op.cpp"
```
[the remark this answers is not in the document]
tbirdso: Seems like existing paths should remain valid with `target_include_directories`, was there an issue with include paths that prompted this change?

applications/xr_hello_holoscan/CMakeLists.txt
```diff
@@ -29,13 +29,26 @@ FetchContent_Declare(
...
+  URL https://github.com/nlohmann/json/releases/download/v3.11.2/json.tar.xz
+  DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+)
+FetchContent_MakeAvailable(nlohmann_json)
+
 find_package(holoscan 2.0 REQUIRED CONFIG)
 
+# Add local XR operators
+add_subdirectory(operators/XrFrameOp)
+add_subdirectory(operators/XrTransformOp)
```
tbirdso, line 44: Suggest removing in favor of `volume_rendering_xr` impl unless there is a need to fork operator code

applications/volume_rendering_xr/utils/xr_hello_holoscan/README.md
```diff
@@ -24,9 +24,31 @@ The mixed reality demonstration scene includes:
...
+```bash
+./dev_container build_and_run volume_rendering_xr
+```
+This will launch the main volume rendering application by default.
+
+### Running the Test Utility
+To run this test utility instead:
+
+1. Open `/applications/volume_rendering_xr/metadata.json`
+2. Modify the `run` command to use `xr_hello_holoscan`:
```
tbirdso, line 39: Rather than modifying `metadata.json`, could we handle trailing arguments from the [run launch --extra_args](https://github.com/nvidia-holoscan/holohub/blob/main/dev_container#L927) flag in `run.sh` out of the box?

A new user should be able to run `xr_hello_holoscan` without modifying files, sample command might look like this:
```bash
./run launch volume_rendering_xr --extra_args "xr_hello_holoscan"
```

applications/volume_rendering_xr/scripts/run.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+APP_NAME=$1
+shift  # Remove first argument, leaving remaining args
```
tbirdso, line 4: Not sure if we can always prepend with `./run launch --extra_args <APP>`. Rather than assuming a positional argument you might instead set a flag on whether `xr_hello_holoscan` is detected as an argument anywhere or not.

mimiliaogo: Fixed in https://github.com/nvidia-holoscan/holohub/pull/764/commits/5515d91dfed98097127871a274edebb9f02c9379.

Thanks for pointing out!

applications/volume_rendering_xr/utils/xr_hello_holoscan/xr_basic_render/README.md
tbirdso, line 1: Seems like the reference link to `xr_hello_holoscan` here needs to be updated

tbirdso (approved):
Changes look good, verified locally with the debug GUI. Nice work!

There is one broken link to fix, once that is updated we should move forward with merge.

tbirdso (approved):
🎉

PR #767, opened 2025-04-02
Permit CMake application dependencies on other applications

tbirdso:
Extend `add_holohub_application` to allow applications to specify dependencies on other applications.

Intended to support LIMITED dependencies among applications. For instance, `volume_rendering_xr` may define operators tightly coupled with that app, while `xr_hello_holoscan` may at first depend on those tightly coupled components and then refine by forking components or consolidating in a reusable module. Prior to this change there is no mechanism to allow one application to enable another application to be built in the same build tree.

Note that the order of `add_holohub_application` calls matters, dependent applications much be listed first in order to enable subsequent dependency applications.

Intended to help mitigate operator code duplication among two related applications in https://github.com/nvidia-holoscan/holohub/pull/764. For instance, `xr_hello_holoscan` may be updated as follows to access XR volume rendering components:
```
add_holohub_application(xr_hello_holoscan DEPENDS
                        APPLICATIONS volume_rendering_xr)
```

### DIFF
--- a/cmake/HoloHubConfigHelpers.cmake
+++ b/cmake/HoloHubConfigHelpers.cmake
@@ -56,7 +56,14 @@ function(add_holohub_application NAME)
 
     # If we have dependencies make sure they are built
     if(APP_DEPENDS)
-      cmake_parse_arguments(DEPS "" "" "EXTENSIONS;OPERATORS" ${APP_DEPENDS})
+      cmake_parse_arguments(DEPS "" "" "EXTENSIONS;OPERATORS;APPLICATIONS" ${APP_DEPENDS})
+      message(DEBUG "${pkgname} exts = ${DEPS_EXTENSIONS}")
+      message(DEBUG "${pkgname} ops = ${DEPS_OPERATORS}")
+      message(DEBUG "${pkgname} apps = ${DEPS_APPLICATIONS}")
+
+      foreach(dependency IN LISTS DEPS_APPLICATIONS)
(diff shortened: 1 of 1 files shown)


mimiliaogo (approved):

agirault (changes requested):
I would advocate for not doing so: as soon as an operator is to be used by multiple applications, that's probably the right signal to move it from under a single application folder into its own operators folder? I think otherwise it goes against sensible expectations from users about where to find dependencies.

Open to be convinced otherwise as always.

mimiliaogo:
> I would advocate for not doing so: as soon as an operator is to be used by multiple applications, that's probably the right signal to move it from under a single application folder into its own operators folder? I think otherwise it goes against sensible expectations from users about where to find dependencies.
> 
> Open to be convinced otherwise as always.

@tbirdso As discussed, we can move the hello world example into volume rendering xr. Then we probably don't need this change anymore?

agirault:
@mimiliaogo per discussion with @tbirdso , I think that would be the right stopgap solution to a specific and temporary problem, and you can move the application in its own folder when generic operators are available to move out also.

tbirdso:
Definitely valid concerns, @mimiliaogo  I agree, let's move `xr_hello_holoscan` under `volume_rendering_xr` as a utility app for testing and document it in the README there. Thanks @agirault  @mimiliaogo !

Closing as not aligned with the current HoloHub structure, executables that share common, narrow dependencies should be contained within a single app subdirectory.

PR #769, opened 2025-04-04
Website improvements including Versioning support

agirault:
## Context

github.io page: https://nvidia-holoscan.github.io/holohub/

- As it stands, any change to the README of an holohub component will override any historical version of the docs on github.io page. This is a concern for components that want to provide for a certain revision in the past, such as the 0.1 holoscan networking package. That project is undergoing interface changes which we want to expose for users of the source repo, while keeping available the interface documentation for the 0.1 stable release that is publicly distributed.
- The website currently exhibits some other issues, all tightly intermingled in the website generation code:
  - Duplication of READMEs when shared by a cpp and python implementation
  - Inconsistencies between paths on the website and github, causing broken relative links and 404s
  - Lack of sensible navigation sorting 
- The website generation script itself lacks coherency, separation of concerns, and has  redundancies as well as unreliable handling of path strings and objects (per Claude 3.7 sonnet thinking)

## Changes

> Recommended to review [commit per commit](https://github.com/nvidia-holoscan/holohub/pull/769/commits).

### New features

- Support for per-component archival versioning

### Enhancement and fixes

- Proper alphabetical sorting of all components by title instead of by path
name, and of singletons before dirs
- Documentation relative paths are consistent between website and git repo, making relative links between them work instead of 404s.
- No duplication of C++ and Python pages for projects with a single README
  - _caveat_: currently metadata headers only appear for a single one of the metadata that share the README. I think the better approach to handling this is not some extra code for this use case, but eventually changing the metadata structure to support multi languages (instead of one metadata per language).
- Relative links in markdown now will send to the right path on GitHub, or to another page on the website if it was generated from a README. Addresses 404s.
- Improved identification of the top page headers to solve missing metadata header in 31 pages.

### Refactoring

> Made to enable clearer implementation of features and enhancements above.

- More robust path definitions
  - reduce redundancy
  - using Path object consistently
  - better handling of relative vs absolute paths
  - better README path discovery and handling
  - clearer language specific directory handling
- Cleaner architecture with better separation of concerns through functions
- Better component organization
  - global definition of COMPONENT_TYPES
  - metadata search only on supported types
  - statistics infered from components tracked in dictionary

### Style

> Can be moved to another MR if requested, though minimal changes, isolated in separate commits.

- Fix some titles and README locations for some projects
- Hide missing navigation in tags page
- ~~Show number of components in top navigation~~
- Add missing url to workflows page
- Rename Home page header to site title and avoid redundant title in top site banner

### Infra

> Can be moved to another MR if requested, though minimal change, isolated in a separate commit.

- Watch input files for live-reload on edit


doc/website/scripts/generate_pages.py
```diff
@@ -20,109 +20,96 @@
...
+        rel_file_path = file_path.relative_to(git_repo_path)
+        cmd = ["git", "-C", str(git_repo_path), "show", f"{git_ref}:{rel_file_path}"]
+        result = subprocess.run(cmd, capture_output=True, text=True, check=True)
+        return result.stdout
+    except (subprocess.CalledProcessError, ValueError) as e:
+        if isinstance(e, subprocess.CalledProcessError):
+            logger.error(f"Git error: {e.stderr}")
+        else:
+            logger.error(f"Path {file_path} is not within the Git repository")
+        raise e
```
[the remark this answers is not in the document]
agirault: yes there is, we raise right below

AndreasHeumann:
I noticed that the images for the Holoviz examples are missing, probably because the images in the app directory are links to the images in the generator directory. This was also the case before but I wonder if this could be fixed with this changed.

The webrtc apps use mermaid for flowcharts, do you think this can be supported as well? Currently only the code is shown.

agirault:
Thank you @AndreasHeumann . There is already more in this MR that I was hoping for given the original scope, I'm actually more enticed to remove some commits than adding more, but I fully agree.

Since this is not a regression but is an issue in main also, I can help with that in a follow-up MR. I think there are simple fixes and improvements to references of files only hosted on github that I could make, though it should have already been supported for images and works for some of them, so that's another bug, though closely related.

tbirdso (commented):
@agirault many thanks for the enhancements here. Responses to the PR overview:

> Features: Support for per-component archival versioning

I see support added in [ed8c40f7](https://github.com/nvidia-holoscan/holohub/pull/769/commits/ed8c40f7ae27517c17a56173896c907a1256229a) but I'm not sure where to find versioned docs in practice. Could you please point me towards where to find both the versioned (0.1.0) and latest high performance networking docs in your internal landing page? I might just be looking past it..

> Proper alphabetical sorting of all components by title instead of by path
name, and of singletons before dirs

Looks much better 👍 

> Documentation relative paths are consistent between website and git repo, making relative links between them work instead of 404s.

Path consistency is good, but note that 404s are not fully addressed under the current state as seen at http://10.28.226.146:8000/holohub/tutorials/pretrained_foundational_models/self_supervised_training/Docker/Dockerfile . Is this something you feel like should be addressed in this PR or in a followup?

> No duplication of C++ and Python pages for projects with a single README. caveat: currently metadata headers only appear for a single one of the metadata that share the README ... I think the better approach to handling this is not some extra code for this use case, but eventually changing the metadata structure to support multi languages (instead of one metadata per language).

My instinct is that we should leave READMEs as separate until a unified metadata strategy is formed so that we avoid the possibility of missing/misleading metadata for one language or another, but I don't feel strongly about it either way. Multi-language metadata structure is a nontrivial lift.

> Refactoring, Style, Infra

Looks good. As you mentioned before, we are still baking in both filesystem organizations and metadata assumptions together, but that is a larger HoloHub issue beyond the scope of this PR.

jjomier:
Thanks for the improvements. Couple of comments:

1. The menu is confusing with the number of each category - we should remove the counts
2. There is a regression with directories containing only one implementation : http://10.28.226.146:8000/holohub/applications/velodyne_lidar_app/cpp/ which should not be shown in a sub-section
3. Workflow should go first in the menu and tags last

agirault:
> 1. The menu with the count is a little bit confusing at first, I'm not sure displaying the count helps with the navigation and a little bit distracting.

Fair, reverted.

> 2. There is a regression with directories containing only one implementation : /holohub/applications/velodyne_lidar_app/cpp/ which should not be shown in a sub-section

That's not a regression. The logic to change the path was creating a lot of issues with cross-references (ex: @huvers issues with tutorials and relative paths to other files).  I strongly believe he proper way to handle it would be to keep alignment between the path in github and the website. In that case, a simple no-code solution that prevents high complexity changes to the script is to have the README one level above (or just not have a cpp folder).

I did it for ano in [this commit](https://github.com/nvidia-holoscan/holohub/pull/769/commits/7f716f487289cd665196a38401b19458150899eb). I will update velodyne also and amend, all the other apps/components look good.

> 3. The menu should not be alphabetized: we want workflows, applications, operators, tutorials, tags in that order

Thank you, updated. I assume you meant benchmarks and not tags? tags currently not in the nav.

> 4. The permalink icon is weird

I don't get that part 

> 5. Seems like the link to images is failing for some: /holohub/applications/stereo_vision/

Thank you, I had lost the HTML image urls, patched.

doc/website/scripts/generate_pages.py
```diff
@@ -136,143 +153,420 @@ def parse_metadata_file(metadata_file: Path, statistics) -> None:
...
-        readme_parent = 1
-    if not os.path.exists(readme_path):
-        readme_path = str(metadata_file.parent.parent) + "/README.md"
-        readme_parent = 2
-
-    if os.path.exists(readme_path):
-        with open(readme_path, "r") as readme_file:
-            readme_text = readme_file.read()
+    header_text = (
+        ":octicons-person-24: **Authors:** "
```
agirault, line 157: Add function for metadata line format

doc/website/scripts/generate_pages.py
```diff
@@ -136,143 +153,420 @@ def parse_metadata_file(metadata_file: Path, statistics) -> None:
...
+            rel_repo_target_path = _get_path_relative_to_repo(
+                original_img_path, git_repo_path, readme_dir
             )
-            header_text += (
-                ":octicons-sparkle-fill-24: **Contribution metric:** " + metric_str + "<br>"
+            if not rel_repo_target_path:
+                continue
+
+            # Create raw GitHub URL
+            encoded_path = "/".join(urlencode(part, safe="") for part in rel_repo_target_path.parts)
```
agirault, line 251: Add function encode_url_parts

Issue #776, opened 2025-04-10
Operator builds fail with latest Holoscan


JohnMoon-VTS:
Hi, I'm trying to rebuild our `psd_pipeline` app with the 3.1.0 Holoscan container and I'm running into a build error in several operators. Here's what it looks like for the FFT operator:

```
[4/29] Building CUDA object operators/fft/CMakeFiles/fft.dir/fft.cu.o
FAILED: operators/fft/CMakeFiles/fft.dir/fft.cu.o
/usr/local/cuda/bin/nvcc -forward-unknown-to-host-compiler -DFMT_HEADER_ONLY=1 -DMATX_DISABLE_CUB_CACHE -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CUDA -DTHRUST_DISABLE_ABI_NAMESPACE -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_CPP -DTHRUST_IGNORE_ABI_NAMESPACE_ERROR -I/workspace/holohub/build/psd_pipeline/_deps/matx-src/include -I/workspace/holohub/build/psd_pipeline/_deps/matx-src/include/matx/kernels -I/workspace/holohub/build/psd_pipeline/_deps/cccl-src/thrust/thrust/cmake/../.. -I/workspace/holohub/build/psd_pipeline/_deps/cccl-src/libcudacxx/lib/cmake/libcudacxx/../../../include -I/workspace/holohub/build/psd_pipeline/_deps/cccl-src/cub/cub/cmake/../.. -isystem=/opt/nvidia/holoscan/include -isystem=/opt/nvidia/holoscan/include/3rdparty -isystem=/opt/nvidia/holoscan/include/gxf -isystem=/usr/local/cuda/include -isystem=/opt/nvidia/holoscan/include/3rdparty/ucx -O3 -DNDEBUG --generate-code=arch=compute_70,code=[compute_70,sm_70] --generate-code=arch=compute_80,code=[compute_80,sm_80] --generate-code=arch=compute_90,code=[compute_90,sm_90] --expt-relaxed-constexpr -std=c++17 -MD -MT operators/fft/CMakeFiles/fft.dir/fft.cu.o -MF operators/fft/CMakeFiles/fft.dir/fft.cu.o.d -x cu -c /workspace/holohub/operators/fft/fft.cu -o operators/fft/CMakeFiles/fft.dir/fft.cu.o
[...]

PR #778, opened 2025-04-10
PSD pipeline updates

JohnMoon-VTS:
A few changes to update the `psd_pipeline` app to use a new packetizer and fix a build error introduced in Holoscan 3.1.0.

This PR works around #776, but there may be a better solution.


tbirdso (approved):
@JohnMoon-VTS thanks for the ping, no concerns about the updates. Neat to see the Rust integration here, I believe this is the first HoloHub contribution to do so.

We'll keep the DLPack versioning issue open for now while we investigate on our end.

PR #790, opened 2025-04-17
Integration Testing for AI Surgical Video Workflow

bhashemian:
This PR implements integration tests for AI Surgical Video Workflow, makes necessary changes, and adds all the required data.


tbirdso (approved):
Great, looks good!

jjomier:
Thanks @bhashemian - looks good.
Wondering if there is a way to still keep 10 frames to test but spread them over the length of the input video so we can test the out of body part of the video. Thoughts?

bhashemian:
> Thanks @bhashemian - looks good. Wondering if there is a way to still keep 10 frames to test but spread them over the length of the input video so we can test the out of body part of the video. Thoughts?

@jjomier that's an excellent point. I'll see what I can do.

bhashemian:
@jjomier @tbirdso, I’ve added a decimation operator to the workflow so that it can run the replayer at specific intervals of the input video frames. In this way, the test images reaches the portion of the video that probe gets out of the body. Could you please help to review and merge? Thanks

jjomier:
> @jjomier @tbirdso, I’ve added a decimation operator to the workflow so that it can run the replayer at specific intervals of the input video frames. In this way, the test images reaches the portion of the video that probe gets out of the body. Could you please help to review and merge? Thanks

Awesome! looks good to me.

workflows/ai_surgical_video/python/ai_surgical_video.py
```diff
@@ -74,6 +75,27 @@ def compute(self, op_input, op_output, context):
         op_output.emit(in_message, "out")
 
 
+class DecimationOp(Operator):
```
jjomier, line 78: Maybe call it something else than DecimationOp (makes me think of 3D decimation)?

bhashemian: Changed it to `FrameSamplerOp`, is it better?

workflows/ai_surgical_video/python/ai_surgical_video.py
```diff
@@ -651,7 +703,8 @@ def dynamic_flow_callback(op):
         elif self.source == "aja":
             self.add_flow(aja, source, {("video_buffer_output", "in")})
         else:
-            self.add_flow(replayer, source)
+            self.add_flow(replayer, frame_sampler)
+            self.add_flow(frame_sampler, source)
```
tbirdso, line 707: Is there an advantage to having `frame_sampler` alter input to the overall application as opposed to only affecting output as seen by the recorder?

If the motivation is solely to support testing, I would suggest running the application with normal input and then sample select output frames for recording and validation, rather than altering the input data source as seen by the application.

I.e. replace with:

```
add_flow(recorder_format_converter, frame_sampler)
add_flow(frame_sampler, recorder)
```

tbirdso: The update looks good, thanks!

workflows/ai_surgical_video/python/ai_surgical_video.py
jjomier: Perfect. Can you fix the linting and we merge

PR #798, opened 2025-04-22
Utility: Fix HoloHub CLI build default data download path

tbirdso:
Use the default data location `holohub/data` at CMake configuration time for consistency with existing app expectations.

Previously a clean build with the `holohub` CLI entrypoint would download data under the binary folder (`build/<app>`) by default, which did not match app expectations.


utilities/cli/holohub.py
```diff
@@ -393,6 +394,7 @@ def handle_run(self, args: argparse.Namespace) -> None:
                     "Ninja",
                     f"-DCMAKE_BUILD_TYPE={BUILD_TYPE}",
                     f"-DCMAKE_PREFIX_PATH={HoloHubCLI.DEFAULT_SDK_DIR}",
+                    f"-DHOLOHUB_DATA_DIR:PATH={HoloHubCLI.DEFAULT_DATA_DIR}",
```
agirault, line 397: out of scope, but can't we leverage/call whats in `handle_build` from `handle_run` ?

tbirdso: Yep, we should refactor this as part of @wyli 's suggestions to make CLI tooling more accessible for gen AI calls. Deferring for now.

PR #802, opened 2025-04-23
add ./holohub create ...

wyli:
add command:
```bash
$ ./holohub create my_proj
Successfully created new project: my_proj 
Project directory: /home/Documents/holohub/applications/my_proj

Possible next steps:
- Add your operators to /home/Documents/holohub/applications/my_proj/src/my_proj.cpp
- Update project metadata in /home/Documents/holohub/applications/my_proj/metadata.json
- Review source code license files and headers (e.g. /home/Documents/holohub/applications/my_proj/LICENSE)
- Build and run your application:
   ./holohub build my_proj
   ./holohub run my_proj

```

output an application folder such as:
```
/home/Documents/holohub/applications/my_proj
├── CMakeLists.txt
├── LICENSE
├── metadata.json
├── README.md
└── src
    └── my_proj.cpp

2 directories, 5 files
```


applications/template/cookiecutter.json
```diff
@@ -0,0 +1,14 @@
+{
+    "project_name": "My Project",
+    "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '_') }}",
+    "full_name": "John Doe",
+    "affiliation": "My Institution",
```
tbirdso, line 5: Could we add [human readable prompts](https://cookiecutter.readthedocs.io/en/stable/advanced/human_readable_prompts.html) for this information?

applications/template/cookiecutter.json
```diff
@@ -0,0 +1,14 @@
...
+    "language": "{{ cookiecutter.language }}",
+    "version": "0.1.0",
+    "holoscan_version": "2.3",
```
tbirdso, line 8: I wonder if we could parse and default to the latest Holoscan SDK release?

applications/template/{{cookiecutter.project_slug}}/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+3. Build the application:
+
+## Usage
+
+### Running the Application
+
+```bash
+./holohub build {{ cookiecutter.project_slug }}
+./holohub run {{ cookiecutter.project_slug }}
```
tbirdso, line 31: `run` always calls `build` first

```suggestion
./holohub run {{ cookiecutter.project_slug }}
```

applications/template/{{cookiecutter.project_slug}}/README.md
```diff
@@ -0,0 +1,79 @@
...
+│   └── operators/
+├── include/
+├── tests/
+└── docs/
+```
+
+### Adding New Operators
+
+1. Create a new operator class in `src/operators/`
+2. Register the operator in `src/main.cpp`
```
tbirdso, line 53:
```suggestion
2. Include the operator in `src/main.cpp`
```

utilities/cli/holohub.py
```diff
@@ -1025,6 +1062,79 @@ def handle_clear_cache(self, args: argparse.Namespace) -> None:
...
+
+        print(
+            Color.green(f"Successfully created new project: {args.project}"),
+            f"\nProject directory: {project_dir}\n\n"
+            f"Possible next steps:\n"
+            f"- Add your operators to {target_file}\n"
+            f"- Update project metadata in {project_dir / 'metadata.json'}\n"
+            f"- Review source code license files and headers (e.g. {project_dir / 'LICENSE'})\n"
+            f"- Build and run your application:\n"
+            f"   ./holohub build {canonical_project_name}\n"
```
tbirdso, line 1134:
```suggestion
```

applications/template/{{cookiecutter.project_slug}}/CMakeLists.txt.template
```diff
@@ -0,0 +1,21 @@
...
+{% if cookiecutter.language == "cpp" %}
+add_executable({{ cookiecutter.project_name }}
+    {{ cookiecutter.project_name }}.cpp
+)
+
+target_link_libraries({{ cookiecutter.project_name }}
+    PRIVATE
+        holoscan::core
+)
+{% endif %}
```
tbirdso, line 21: Should we add a dummy test to encourage coverage? Something like:

```suggestion
if(BUILD_TESTING)
  add_test({{ cookiecutter.project_name }}-test
    COMMAND echo "Add test coverage for {{ cookiecutter.project_name }} here"
endif()
{% endif %}
```

tbirdso (commented):
@wyli very nice! Do you have thoughts on how we could support other project types with various templates:
- C++ operators + Python wrappings
- Python operators

utilities/cli/holohub.py
wyli: Hi @tbirdso, after `./holohub create my_proj`, it seems I still need to add `add_holohub_application(my_proj)` to `applications/CMakeLists.txt`.

 otherwise `./holohub run my_poj` won't work. is there a way to avoid modifying `applications/CMakeLists.txt`?

wyli: (@jjomier mentioned that this could be a new feature to support `./holohub run my_proj` when `my_proj` is not in the standard directory.)

wyli:
> support other project types with various templates:
> 
> * C++ operators + Python wrappings
> * Python operators

yes, currently the default template is 'applications/template' https://github.com/wyli/holohub/blob/013e7f08583dccbbb19bb97d7a4949d3d235f361/utilities/cli/holohub.py#L73-L76 for the operators we can create different templates and reuse the `handle_create` logic.

applications/template/cookiecutter.json
tbirdso: There are a few places we can pull this information from, open to opinions on the best source:
- Agreed `holoscan.__version__` is simplest, though it requires the user has the latest `holoscan` package installed locally. Maybe there is a good way such as `pip search` to find the latest version on PyPI without installing it?
- We currently inline this information in HoloHub CLI [here](https://github.com/nvidia-holoscan/holohub/blob/main/utilities/cli/container.py#L48), could expose the expected container version as a string staticmethod that we continue to manually update
- If a `holoscan` C++ install tree is available in the local environment (apt/container) we could parse `holoscan-config-version.cmake`

utilities/cli/holohub.py
tbirdso: Is it possible to use `cookiecutter` or otherwise append to `applications/CMakeLists.txt` when `holohub create` is run? You're right, at the moment it is required that C++ projects (or any project leveraging the CMake build system) have an `add_holohub_application` entry.

applications/template/cookiecutter.json
wyli: thanks, for now I go for the 2nd option to add a unified variable to hold this info.

utilities/cli/holohub.py
wyli: added a function to modify the cmakelists.txt https://github.com/wyli/holohub/blob/2f7745f98b198c6c37a9d0dcfdf6e7ec2095cf58/utilities/cli/holohub.py#L1078-L1079 I think we can further improve 'build' and 'run' to process folders with local cmakelists.txt.

utilities/cli/holohub.py
```diff
@@ -1025,6 +1075,133 @@ def handle_clear_cache(self, args: argparse.Namespace) -> None:
...
+
+        import cookiecutter.main
+
+        if project_dir.exists():
+            holohub_cli_util.fatal(f"Project directory {project_dir} already exists")
+
+        try:
+            cookiecutter.main.cookiecutter(
+                str(template_dir),
+                no_input=True,
```
tbirdso, line 1158: Seems like `__prompts__` is not respected unless input is allowed (could control with a command parameter?)

```suggestion
                no_input=False,
```

utilities/cli/holohub.py
```diff
@@ -1025,6 +1075,133 @@ def handle_clear_cache(self, args: argparse.Namespace) -> None:
...
+                extra_context=context,
+                output_dir=str(args.directory),
+            )
+        except Exception as e:
+            holohub_cli_util.fatal(f"Failed to create project: {str(e)}")
+
+        # Copy the appropriate example source file
+        ext_name = "cpp" if args.language.lower() in ["c++", "cpp"] else "py"
```
tbirdso, line 1166: `__prompts__` also prompts for this information, but entering "python" still results in a C++-based project being created. Could we align to accept the language either from the config prompt (preferred) or from the command line, but not both?

I.e. after running the following:
```
$ ./holohub create mynewapp3
project_name [mynewapp3]: My New App
project_slug [mynewapp3]: 
full_name [tbirdsong]: myname
affiliation [Holoscan]: NVIDIA
language [cpp]: python
version [0.1.0]: 1.0.0
holoscan_version [3.1.0]: 
platforms [["x86_64", "aarch64"]]: 
tags [["keyword1", "keyword2", "keyword3"]]: 
description [A brief description of the project]: 
license [Apache-2.0]: 
year [2025]: 
Successfully created new project: mynewapp3 
Directory: /home/tbirdsong/repos/holohub-internal/applications/mynewapp3
```

I would expect to see Python stubs in `mynewapp3` but instead I see C++ stubs

utilities/cli/holohub.py
wyli: Yes, added --interactive or -i, default to true.

tbirdso (approved):
Updates LGTM, thanks! Please review merge conflict

PR #805, opened 2025-04-24
minor fixes CLI

wyli:
collection of some minor CLI fixes during development:
- dryrun broken commands e.g. `--device-cgroup-rule c 81:* rmw` to `--device-cgroup-rule "c 81:* rmw"`
- remove strict deps check when `dryrun` is true
- docs: run is actually build+run, removed some duplicated code
- _find_project internally handle the case of not found and exit, so no need to check it after _find_project 
- `run_cmd += " --dryrun"` not needed as `container.run` is already in dryrun mode: https://github.com/nvidia-holoscan/holohub/blob/965fa6604dfa65afbde86471edff0af6899c47ff/utilities/cli/holohub.py#L522-L523

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -259,40 +259,59 @@ class HoloHubCLI:
         return None
 
     def _make_project_container(
-        self, project_name: Optional[str] = None, language: Optional[str] = None
+        self,
+        project_name: Optional[str] = None,
+        language: Optional[str] = None,
+        dryrun: bool = False,
+        build_args: Optional[dict] = None,
     ) -> HoloHubContainer:
-        """Define a project container"""
-        if not project_name:
-            return HoloHubContainer(project_metadata=None)
+        """Define a project container and optionally build it
 
-        project_data = self._find_project(project_name=project_name, language=language)
-        if not project_data:
-            holohub_cli_util.fatal(
-                f"Project '{project_name}' {f'({language})' if language else ''} not found"
-            )
-        return HoloHubContainer(project_metadata=project_data)
+        Args:
+            project_name: Name of the project
+            language: Language implementation (cpp/python)
+            dryrun: Whether to run in dryrun mode
+            build_args: Optional arguments to pass to container.build()
+        """
+        if not project_name:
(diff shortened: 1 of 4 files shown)


utilities/cli/holohub.py
```diff
@@ -259,40 +259,55 @@ def _find_project(self, project_name: str, language: Optional[str] = None) -> di
...
 
-        project_data = self._find_project(project_name=project_name, language=language)
-        if not project_data:
-            holohub_cli_util.fatal(
-                f"Project '{project_name}' {f'({language})' if language else ''} not found"
-            )
-        return HoloHubContainer(project_metadata=project_data)
+        container.dryrun = dryrun
+        if build_args is not None:
+            container.build(**build_args)
```
tbirdso, line 284: I wonder about this heuristic change. Is there a clear motivation for changing to implicitly build, rather than always calling `container.build` explicitly on the resulting object?

utilities/cli/holohub.py
```diff
@@ -318,51 +333,58 @@ def get_buildtype_str(self, build_type: Optional[str]) -> str:
...
-                ["cmake", "--build", str(build_dir), "--config", build_type], dry_run=args.dryrun
-            )
-
         else:
             # Build in container
             container = self._make_project_container(
                 project_name=args.project,
                 language=args.language if hasattr(args, "language") else None,
+                dryrun=args.dryrun,
+                build_args={},  # Empty dict to trigger build but with no special args
```
tbirdso, line 386: Concerned that the default behavior here with `None` versus `{}` is less intuitive than calling `.build` explicitly

utilities/cli/holohub.py
```diff
@@ -318,51 +333,58 @@ def get_buildtype_str(self, build_type: Optional[str]) -> str:
         }
         return build_type_map.get(build_type.lower(), "Release")
 
+    def _build_project_locally(
```
tbirdso, line 336: Consolidation looks good!

tbirdso (approved):
LGTM, thanks for making those updates 👍

PR #806, opened 2025-04-24
CI Check for Title/Name Matching

bhashemian:
This PR

1. Implements a test to check if "name" field in the metadata does not contains "forbidden" terms like Application, Holoscan, Holohub, and its variations.
2. Implements a test to check if README's title matches "name" field in the metadata
3. Fixes minor title-name descrepancies in some applications
4. Adds individual README.md for Laser Detection Latency Applications
    - USB Camera Calibration
    - EVT Camera Calibration
    - Laser Detection

Note: This check is currently only active for applications but we can enable it for other entities like operators, benchmarks, etc. once we have resolved the existing discrepancies in those entities too.


applications/psd_pipeline/README.md
```diff
@@ -1,8 +1,3 @@
-<!--
```
tbirdso, line 1: Can we keep this header? This file is contributed by an external collaborator.

agirault: 100%

applications/psd_pipeline/data_writer/README.md
```diff
@@ -1,8 +1,3 @@
-<!--
-SPDX-FileCopyrightText: 2024 Valley Tech Systems, Inc.
-
-SPDX-License-Identifier: Apache-2.0
--->
```
agirault, line 5: please revert as well

applications/psd_pipeline/data_writer/metadata.json
```diff
@@ -1,6 +1,6 @@
 {
     "operator": {
-        "name": "data_writer",
+        "name": "Data Writer Operator",
```
agirault, line 3: whats wrong using this name? I believe it's the same as the cmake target name which makes it convenient to recognize and use quickly as that defines the name in the holohub website navigation.

agirault: most operators have that format: https://nvidia-holoscan.github.io/holohub/operators/

applications/psd_pipeline/README.md
bhashemian: Thanks for the feedback. I understand it’s an external contributor, but it doesn’t align with the Holohub style. Let’s break down its content:

- Copyright statement: I agree that we should keep it in the README. However, instead of keeping it at the top, let’s move it to the bottom of the README as it is the usual place for the documents to have copyright statement unlike the codes that have it at the top.

- License statement: Since all contributions to Holohub are licensed under Apache 2.0, this license statement is redundant. Therefore, it makes sense to remove it.

@tbirdso @agirault @jjomier do you agree?

applications/psd_pipeline/data_writer/metadata.json
bhashemian: @agirault I agree, It is fine to have the names like that but we should be consistent across different files. I will revert this as it is not in the scope of this PR but let's discuss about the operator naming online.

applications/psd_pipeline/README.md
tbirdso: Sorry but I disagree:
- The header here is in a comment block in the README source and doesn't currently render as part of documentation ([ref](https://github.com/nvidia-holoscan/holohub/blob/main/applications/psd_pipeline/README.md)). In that respect this is still "source code"
- The copyright reflects the external contributor, even if it is the same type of license

In order to encourage external contributions we should avoid touching copyright headers in external contributions unless changes are trivial and/or absolutely necessary for maintenance.

For least effort I would suggest that we add an exemption policy for the name check and allow the `psd_pipeline` README to be exempt to work around the copyright block here. Alternatively, it could be useful for the name check to be able to parse and ignore comment headers in markdown files when conducting checks. What do you think?

bhashemian: Alright, I’ve reverted the change and updated the title extractor to ignore the HTML header.

However, I believe we shouldn’t easily compromise on the consistency. In my opinion, this isn’t a restriction on contributors but rather a matter of different formatting, and we should enforce it. Regardless, this isn’t related to this PR, so we can discuss it at a later time. Thanks again for sharing your thought.

applications/psd_pipeline/data_writer/README.md
bhashemian: done!

PR #808, opened 2025-04-24
PSD Pipeline Fixes

JohnMoon-VTS:
A recent commit (b8b7c82) removed the ANO operator interface which the
psd_pipeline application uses.

Instead, the ANO is now exposed as a library which we initialize and
fetch bursts from.

Before this change, the VITA connector compute() function was only called
when individual batches were ready in the ANO. Now, the function is run
in a loop and can process any/all channels it wants.

This change requires a refactor in how the psd_pipeline's VITA connector
operator behaves; now, it kicks off copy/processing for all channels it
can and emits the channels one at a time (as one compute() call cannot
have multiple op_output.emit() calls).

At the same time, reduce some of the log spew in the pipeline to make
the output more friendly and reduce some CPU load.


cliffburdick (approved):

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -290,23 +296,20 @@ void Vita49ConnectorOpRx::compute(
...
   }
 }
 
 void Vita49ConnectorOpRx::process_channel_data(
         OutputContext& op_output,
         BurstParams *burst,
-        uint16_t channel_num) {
-  auto channel = channel_list.at(channel_num);
-
+        std::shared_ptr<struct Channel> channel) {
```
agirault, line 325: fully your call, just curious why you switched from passing a channel id to passing the channel directly? Would make sense to me if you did anything else with the channel in the calling function, but you don't.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -290,23 +296,20 @@ void Vita49ConnectorOpRx::compute(
...
-  // Assumes that channel 0 is queue 1, channel 1 is queue 2, etc.
-  const auto num_rx_queues = get_num_rx_queues(port_id_);
   for (uint16_t q = CONTEXT_QUEUE_ID + 1; q < num_rx_queues; q++) {
+    auto channel = channel_list.at(q - 1);
+    // If there's new data, start processing it
     auto status = get_rx_burst(&burst, port_id_, q);
     if (status == Status::SUCCESS) {
-      uint16_t channel_num = q - 1;  // Should never be negative since CONTEXT_QUEUE_ID is 0
-      process_channel_data(op_output, burst, channel_num);
+      process_channel_data(op_output, burst, channel);
```
agirault, line 317: You could technically only lookup your channel ptr if you have a burst:
```cpp
  // Check all data queues (id >= 1) for an available burst
  for (uint16_t q_id = CONTEXT_QUEUE_ID + 1; q < num_rx_queues; q_id++) {
    auto status = get_rx_burst(&burst, port_id_, q_id);
    // If there's new data, start processing it
    if (status == Status::SUCCESS) {
      auto channel = channel_list.at(q_id - 1);
      process_channel_data(op_output, burst, channel);
    }
```

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -341,13 +344,6 @@ void Vita49ConnectorOpRx::process_channel_data(
   if (channel->aggr_pkts_recv >= num_packets_per_batch) {
     HOLOSCAN_LOG_INFO("Aggregated {} packets on channel {} index {} - sending downstream",
                       channel->aggr_pkts_recv, channel->channel_num, channel->cur_idx);
-    do {
-      free_bufs_and_emit_arrays(op_output, channel);
-      if (channel->out_q.size() == num_concurrent) {
-        HOLOSCAN_LOG_ERROR("Fell behind in processing on GPU!");
-        cudaStreamSynchronize(channel->streams[channel->cur_idx]);
-      }
-    } while (channel->out_q.size() == num_concurrent);
```
agirault, line 350: 1. why not keep `free_bufs_and_emit_arrays`? wouldn't that allow you to check if you have more space to free up before trying to process this aggregated data? 
2. Your `h_dev_ptrs`, `streams` and `events` arrays are sized by `num_concurrent`, if your NIC is trying to write data faster than you can place it in `rf_data`, you'll run out of space and this current warning is how you'd catch it?

Though glancing a bit more, I see you have a `num_simul_batches_` param which seems maybe to add some redundancy to `num_concurrent`, and the default is 2 which is lower than the 4 for `num_concurrent` so you probably never encountered this (though could with a higher value as input?)

Might be more to look at here. No blocker of course, it's your code to merge as you see fit, just sharing some observations.

cc @cliffburdick 

If you're not concerned about being able to place your data fast enough in rf_data to free up space for the NIC, you can then get rid of the `num_concurrent` variable completely,

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -222,16 +218,26 @@ void Vita49ConnectorOpRx::free_bufs_and_emit_arrays(
...
         InputContext& op_input,
         OutputContext& op_output,
         ExecutionContext& context) {
+  const auto num_rx_queues = get_num_rx_queues(port_id_);
+  // Try to emit any waiting data on any channel that's ready (but
+  // only one "emit()" call per "compute()" call.
+  for (uint16_t q = CONTEXT_QUEUE_ID + 1; q < num_rx_queues; q++) {
+    auto channel = channel_list.at(q - 1);
+    if (free_bufs_and_emit_arrays(op_output, channel)) {
+        break;
```
agirault, line 237: I have not tested, but if you wanted to emit all arrays which are ready to go to reduce latency, you could try something like this in the future:

holochatgpt needed a lot of help, but I got to this at the end:

1. Operator Header

Define a Resource parameter in your operator class struct
```cpp
Parameter<Resource> transmitter_;
```

2. Operator Source

Use that parameter for your output port transmitter
```cpp
void Vita49ConnectorOpRx::setup(OperatorSpec& spec) {
  spec.output<out_t>("out").transmitter(transmitter_);
}
```

Register it for users to override.
```
void Vita49ConnectorOpRx::initialize() {
  register_parameter(transmitter_, "transmitter",
                     "Custom transmitter for output port",
                     make_resource<gxf::DoubleBufferTransmitter>("default_transmitter"));
}
```

3. Create Custom Transmitter in your app compose
```cpp
auto transmitter = make_resource<gxf::DoubleBufferTransmitter>(
  "transmitter", Arg("capacity") = 4
);
```

4. Pass Transmitter to Operator Constructor in your app compose
```cpp
make_operator<ops::Vita49ConnectorOpRx>(
  "vitaConnectorOp",
  from_config("vita_connector"),
  Arg("transmitter") = transmitter
);
```

Might even be able to just create the transmitter with the size dynamically set to `get_num_rx_queues(port_id_) - 1` in your operator `initialize` method so that users don't need to input it in the app. I just don't know if that's not too late to configure the output port in the holoscan lifecycle.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
```diff
@@ -222,16 +218,26 @@ void Vita49ConnectorOpRx::free_bufs_and_emit_arrays(
   meta->set("gain_stage_2_db", channel->current_context.gain_stage_2_db);
   meta->set("sample_rate_hz", channel->current_context.sample_rate_sps);
 
-  cudaStream_t stream = completed_msgs[0].stream;
```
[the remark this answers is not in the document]
JohnMoon-VTS: Yeah, I was a bit confused by this as well. I think in practice we didn't see more than one message being completed at once, so this never came up.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
JohnMoon-VTS: I think I did do something with it earlier while refactoring and it ended up not being used. You're right, this change probably isn't necessary.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
JohnMoon-VTS: > why not keep free_bufs_and_emit_arrays? 

Since we're doing it at the top of the `compute()` function now, I didn't want to hit the case where we call `emit()` twice during one `compute()`. If we end up using the custom transmitter you mention, maybe this issue goes away.

> and the default is 2 which is lower than the 4 for num_concurrent so you probably never encountered this 

Right, we've never encountered this, but it would be good to have a check. I'll noodle on this some more.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
JohnMoon-VTS: This is interesting! If I'm reading the docs around the `DoubleBufferTransmitter` correctly, this is already the default, but the big difference here is you're setting the `capacity` to `4`. Does that mean I can `emit()` 4 times in one compute? If so, do I need to set up a `DoubleBufferReceiver` on the downstream operator to take in 4 buffers? Or would the downstream operator just be scheduled 4 times?

cliffburdick: That's right. I asked this question to the HS developers a while ago and this is the solution I was told:

```
spec.input<holoscan::TensorMap>("in").connector(ConnectorType::kDoubleBuffer, holoscan::Arg("capacity", 5UL))
```

I think just that one change is needed.

JohnMoon-VTS: Okay. Just curious - is there a strong reason to prefer this method over say... adding another dimension to our output tensor?

cliffburdick: No, I think both ways are fine. This is mostly just an option if you want to get this to work without any other modifications. It might have scheduling differences though where if you send N items all at once it may schedule the other operator N times sequentially, but I'm not sure.

JohnMoon-VTS: Okay sounds good. We'll take this up internally and figure out how we want to proceed. For now, emitting once per compute is keeping up with our data rates, so I think we're okay. Thanks for the tips!

JohnMoon-VTS:
@agirault, I took some of your comments and updated. The larger items of emitting all 4 channels at once will require some more thought. For now, I think we should merge this so it's at least working on `main` and then we'll work out a path forward for reducing latency. Thanks for the detailed review!

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
agirault: I ran that by @grlee77 and he confirmed my fear that updating the transmitter after `setup` won't work. that means, if you want to allow multiple emits, you would be limited to something like this at present:

```cpp
  void setup(holoscan::OperatorSpec& spec) override {
    spec.output<out_t>("out", IOSpec::IOSize(MAX_NUM_RX_QUEUES));
    
    ...
```

JohnMoon-VTS:
@tbirdso, I think this is ready for merge?

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
agirault: didn't have to, was just curious. ty

JohnMoon-VTS: Right, but I also like to avoid spurious changes when possible. :)

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
agirault: and 100%, meant for future discussions. I believe it would just schedule downstream as many times as you emit.

applications/psd_pipeline/advanced_network_connectors/vita49_rx.cu
agirault: I'm in that team too!

PR #816, opened 2025-05-01
update xr_holoviz tag

wyli:
llm generated tags based on the new app

### DIFF
--- a/applications/xr_holoviz/metadata.json
+++ b/applications/xr_holoviz/metadata.json
@@ -29,12 +29,7 @@
 		"platforms": [
 			"x86_64"
 		],
-		"tags": [
-			"XR",
-			"OpenXR",
-			"Vulkan",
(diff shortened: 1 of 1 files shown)


applications/xr_holoviz/metadata.json
```diff
@@ -29,12 +29,7 @@
 		"platforms": [
 			"x86_64"
 		],
-		"tags": [
-			"XR",
-			"OpenXR",
-			"Vulkan",
-			"3D Visualization"
-		],
+		"tags": ["extended reality", "visualization", "stereoscopic projection matrices", "XR composition layer management", "OpenXR", "rendering", "HoloViz UI", "stereo", "XR"],
```
jjomier, line 32: Let's remove "HoloViz UI" and also XR and extended reality are the same.

PR #821, opened 2025-05-02
MONAI tool segmentation testing

jjomier:
Added support for Holohub CLI and testing for MONAI tool segmentation


wyli (approved):
thanks, looks good to me, I also verified the tests locally.

PR #823, opened 2025-05-03
managing multiple versions of holohub docs serving using mkdocs plugin

wyli:
currently `mike` was set up but not fully enabled, this causes page loading errors
Screenshot 2025-05-03 at 10 28 35

the pr enables mike build support different versions under the gh-pages branch
demo:
 - test version 0.1.0dev: https://wyli.github.io/holohub/0.1.0dev/
 - latest: https://wyli.github.io/holohub/latest/

build branch:
 - https://github.com/wyli/holohub/tree/gh-pages


jjomier:
I don't think we want versioning for the website. Why is this needed?

wyli:
> I don't think we want versioning for the website. Why is this needed?

I see mainly two reasons:
- the current mkdocs always automatically builds with the latest holohub commit and overwrites the previous builds. If I understand it correctly if the user is interested in a specific tag (from this tagging list https://github.com/nvidia-holoscan/holohub/tags), they would have to build the website manually. This pr persists the builds of the tagged holohub versions automatically using the mike plugin.
- the mike plugin has been configured already on the main branch https://github.com/nvidia-holoscan/holohub/blob/main/doc/website/mkdocs.yml#L59 but it's not enabled via the build commands. This causes a 404 error for the versions.json file during our front page loading because the plugin tries to manage multiple versions.

Just realised perhaps this pr's title is not accurate. It's more about managing multiple versions of website via mkdocs plugin..let me update it

jjomier:
Thanks for the clarification @wyli. We are not planning to maintain multiple versions of Holohub. The tags are mostly for building with older version of Holoscan. I believe @agirault has been using something different for managing versioning of the documentation for holoscan-networking. Alexis, what is your take on this?

wyli:
Ok sure, I'm closing this PR..

PR #825, opened 2025-05-05
Pinning version of NGC-CLI

jjomier:


tbirdso (approved):
Thanks, have we confirmed that 3.64.3 has expected download behavior?

PR #832, opened 2025-05-07
Enable X11 display container parameters by default

tbirdso:
Under previous behavior the `--ssh_x11` flag must be specified to mount the X11 socket directory and set the DISPLAY variable in the container, with headless as the default. (`./holohub run-container --ssh_x11`)

This fixes behavior so that X11 display behavior is enabled by default. (`./holohub run-container` or `./holohub run <app>`)


wyli (approved):
thanks, looks good to me

PR #840, opened 2025-05-14
group apt installs in setup

agirault:
Could do more by grouping the ones under install_cuda_dependencies_package as well.

There should not be these apt install in the dockerfile however.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ ARG DEBIAN_FRONTEND=noninteractive
 
 # --------------------------------------------------------------------------
 #
-# Holohub run setup 
+# Holohub run setup
 #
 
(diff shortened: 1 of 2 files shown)


wyli (approved):
Cool, looks good to me, we’ll probably need similar logic here https://github.com/nvidia-holoscan/holohub/blob/0f7bc6256b753ba15f79e2637a4151ccb8d29d20/utilities/cli/holohub.py#L799

agirault:
@wyli oh yeah that doesn't look nice.

@tbirdso @jjomier what should we do in general when we see infra issues in run and dev_container? Also update the python cli? the python cli seems to have a lot of gaps still just looking at the code

PR #844, opened 2025-05-14
Relax pybind11 ABI restriction

agirault:
## Context

When building on ubuntu 24.04 with gcc 13, and linking against the hsdk python bindings
built on ubuntu 22.04 with gcc 11, trying to import these bindings, would fail like so:

```py
ImportError: generic_type: type 'MyOp' referenced unknown base type 'holoscan::Operator'
```

Reference:
- https://github.com/pybind/pybind11/issues/3751
- https://github.com/pybind/pybind11/pull/2602

## Changes

The Holoscan SDK 3.3 will provide a new cmake target named `holoscan::pybind11` with
overriden `PYBIND11_` compiler definitions to disable strict ABI protection.

This change stays backwards compatible with the Holoscan SDK 3.2 and below by simply
warning if the `holoscan::pybind11` target is not found, in which case developers will
need to ensure they are using the same compiler and C++ library version as what was
used in the version of the Holoscan SDK they're linking against.

### hsdk >= 3.3

- CMake log:

   ```log
  -- my_op_python: Linking against holoscan::pybind11 to disable strict ABI protection in pybind11. See https://docs.nvidia.com/holoscan/sdk-user-guide/holoscan_create_operator_python_bindings.html#pybind11-abi-compatibility for details
   ```

- Runtime: no more `ImportError`

### hsdk < 3.3

- CMake log:

   ```log
     -- my_op_python: holoscan::pybind11 target not found, using pybind11's default ABI protection. See https://docs.nvidia.com/holoscan/sdk-user-guide/holoscan_create_operator_python_bindings.html#pybind11-abi-compatibility for details
   ```

- Runtime:  There are no errors if sticking to the same build environment as the hsdk 3.2 and below (build and dev containers) as is the case currently. If building holohub operator bindings with gcc 13 such as on ubuntu 24.04, the exception will now provide a relevant note:

  ```trace
    File "/workspace/holohub/applications/my_app/my_app.py", line 31, in <module>
      from holohub.my_op import MyOp
    File "/workspace/holohub/build/my_app/python/lib/holohub/my_op/__init__.py", line 42, in <module>
      raise e
    File "/workspace/holohub/build/my_app/python/lib/holohub/my_op/__init__.py", line 23, in <module>
      from ._my_op import MyOp
  ImportError: generic_type: type "MyOp" referenced unknown base type "holoscan::Operator"

  - Holoscan SDK >= 3.3.0: make sure to link your bindings against 'holoscan::pybind11'.
  - Holoscan SDK < 3.3.0: use the same compiler version as your installation of the Holoscan SDK.

  See https://docs.nvidia.com/holoscan/sdk-user-guide/holoscan_create_operator_python_bindings.html#pybind11-abi-compatibility for details.
  ```


operators/aja_source/python/__init__.py
```diff
@@ -1,22 +0,0 @@
-"""
```
tbirdso, line 1: Motivation for removal?

cmake/pybind11_add_holohub_module.cmake
```diff
@@ -49,6 +49,15 @@ function(pybind11_add_holohub_module)
             ${MODULE_CPP_CMAKE_TARGET}
     )
 
+    # Conditionally link to the ABI config target if it exists (for HSDK >= 3.3)
+    set(pybind11_abi_details_msg "See https://github.com/pybind/pybind11/pull/2602 for details")
+    if(TARGET holoscan::pybind11)
+        message(STATUS "${target_name}: Linking against holoscan::pybind11 to disable strict ABI protection. ${pybind11_abi_details_msg}.")
+        target_link_libraries(${target_name} PRIVATE holoscan::pybind11)
+    else()
+        message(WARNING "${target_name}: holoscan::pybind11 target not found, using pybind11's default ABI protection. Are you using Holoscan SDK < 3.3? You might need to build ${target_name} with the same compiler and C++ library version as what this version of the Holoscan SDK was built with. ${pybind11_abi_details_msg}.")
```
tbirdso, line 58: How should the user discover the compiler/stdlib used to build their HSDK version? Could we point to docs?

agirault, line 58: @jjomier are you fine having this as a warning?

agirault: Switched to message in cmake: https://github.com/nvidia-holoscan/holohub/pull/844/commits/2077f17c44a336b18fb0606136b2cc752a8155ac#diff-7331419371c0630f72747d4cd01eaa199222c61d33880dd09f09c15027154babR58

and extra info if `ImportError` occurs: https://github.com/nvidia-holoscan/holohub/pull/844/commits/2077f17c44a336b18fb0606136b2cc752a8155ac

The docs referenced will answer the questions you raised.

tbirdso (approved):
Thanks, looks good!

PR #845, opened 2025-05-15
VolumeRendering: fix GCC 13 build error

AndreasHeumann:
Output: `error: 'uint32_t' has not been declared`
Solution: Patch ClaraViz `Exception.h` file to include <cstdint>

### DIFF
--- a/operators/volume_renderer/clara_viz_gcc_13_fix.patch
+++ b/operators/volume_renderer/clara_viz_gcc_13_fix.patch
@@ -1,0 +1,11 @@
+diff -Naur a/src/claraviz/util/Exception.h b/src/claraviz/util/Exception.h
+--- a/src/claraviz/util/Exception.h	2025-05-15 09:09:08.863921942 +0200
++++ b/src/claraviz/util/Exception.h	2025-05-15 09:08:43.697934661 +0200
+@@ -18,6 +18,7 @@
+ 
+ #include <sstream>
+ #include <stdexcept>
++#include <cstdint>
+ 
+ namespace clara::viz
+ {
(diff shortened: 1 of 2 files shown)


operators/volume_renderer/clara_viz_gcc_13_fix.patch
```diff
@@ -0,0 +1,11 @@
+diff -Naur a/src/claraviz/util/Exception.h b/src/claraviz/util/Exception.h
+--- a/src/claraviz/util/Exception.h	2025-05-15 09:09:08.863921942 +0200
++++ b/src/claraviz/util/Exception.h	2025-05-15 09:08:43.697934661 +0200
+@@ -18,6 +18,7 @@
+ 
+ #include <sstream>
+ #include <stdexcept>
++#include <cstdint>
```
agirault, line 8: Could we update upstream instead and update the tag/sha in holohub?

agirault (changes requested):

operators/volume_renderer/clara_viz_gcc_13_fix.patch
AndreasHeumann: Yes, I will upstream this to ClaraViz. But creating and publishing a new release takes time, having the patch is quicker. Once there is a new ClaraViz version I'll remove the patch.

agirault: Thank you. Building from source is too slow?

AndreasHeumann: No, the operator already builds ClaraViz from source. But releasing a new version of ClaraViz *is* slow.

PR #848, opened 2025-05-15
adv_net: fix build for 24.04 base

agirault:
- add missing <array> include
- fix dependencies in Dockerfile

Bonus:
- cleanup dockerfile deps
- cleanup gpunetio cmakelists.txt

### DIFF
--- a/operators/advanced_network/advanced_network/managers/gpunetio/CMakeLists.txt
+++ b/operators/advanced_network/advanced_network/managers/gpunetio/CMakeLists.txt
@@ -16,31 +16,56 @@ cmake_minimum_required(VERSION 3.20)
 
 message(STATUS "PROJECT_NAME: ${PROJECT_NAME}")
 
-target_include_directories(${PROJECT_NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
-target_sources(${PROJECT_NAME} PRIVATE
-  adv_network_doca_mgr.cpp
-  adv_network_doca_mgr_obj.cpp
-  adv_network_doca_kernels.cu
-)
+# --- Dependencies ---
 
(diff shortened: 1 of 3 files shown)


operators/advanced_network/advanced_network/managers/gpunetio/CMakeLists.txt
```diff
@@ -16,31 +16,56 @@ cmake_minimum_required(VERSION 3.20)
...
+    ALLOW_EXPERIMENTAL_API
+    DOCA_ALLOW_EXPERIMENTAL_API
+)
 
-set_target_properties(${PROJECT_NAME} PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
-set_target_properties(${PROJECT_NAME} PROPERTIES CUDA_RESOLVE_DEVICE_SYMBOLS ON)
-set_target_properties(${PROJECT_NAME} PROPERTIES CUDA_ARCHITECTURES "80;90")
+set_target_properties(${PROJECT_NAME}
+  PROPERTIES
+    CUDA_SEPARABLE_COMPILATION ON
```
cliffburdick, line 54: this is what turns rdc on. I missed this before but I'd rather not have this enabled.

cliffburdick (changes requested):
let's discuss the RDC line

operators/advanced_network/advanced_network/managers/gpunetio/CMakeLists.txt
agirault: that's just for gpunetio though, not propagated otherwise. also, this commit doens't add it, it just makes the cmakelists more readable and removes the flags not needed.

Agreed we can discuss rdc, but that's not in this MR's scope.

PR #852, opened 2025-05-16
Fix python-dev install

jjomier:
Checking for Python max version only if already installed and meeting the minimum requirements.


agirault (approved):
🙏

PR #854, opened 2025-05-19
Reverting default config for endoscopy tool tracking

jjomier:
Disabling RDMA and enabling it for overlay is not possible with AJA.
For simplicity and consistency reverting back to original default configuration.
Note: the modification was introduced by Deltacast for testing.


tbirdso (approved):
Thanks @jjomier . FYI @laurent-radoux  @Nathoufresh

PR #855, opened 2025-05-19
Remove ubuntu default user from devcontainer base image

agirault:
Avoid conflict with vscode user creation in devcontainer


tbirdso (approved):
@agirault could you please add a note (could be in the squash commit message) that this is required for Holoscan >= v3.3.0 due to upgraded (24.04) dev container?

mocsharp (approved):

PR #858, opened 2025-05-20
Pinning version of HSDK 3.2 for Qt video replayer

jjomier:


tbirdso (approved):

tbirdso:
FYI @AndreasHeumann

AndreasHeumann:
@jjomier Could we add a comment on why we pin to v3.2?

PR #861, opened 2025-05-20
Update CV-CUDA version

jjomier:
Update CV-CUDA to 0.15 to support Python 3.12


grlee77 (approved):
Looks good, thanks!

I see 0.15.0 removed the "b0" beta designation that were on prior wheel releases so a `--pre` is no longer required in the `pip install` command.

PR #862, opened 2025-05-20
Update generated bar graph in release benchmarking

sohamm17:
Updated graph includes unit in ms and graph title.


tbirdso (approved):
Thanks @sohamm17 !

PR #863, opened 2025-05-21
Adding test command for holohub CLI

jjomier:
Adding` ./holohub test` to run ctest script for testing Holohub applications


utilities/testing/holohub.ctest
```diff
@@ -0,0 +1,77 @@
...
+  -DFLOW_BENCHMARKING=ON
+  # Define the default applications to test
+  -DAPP_colonoscopy_segmentation=1
+  -DAPP_endoscopy_tool_tracking=1
+  -DAPP_endoscopy_out_of_body_detection=1
+  -DAPP_multiai_endoscopy=1
+  -DAPP_ultrasound_segmentation=1
+  -DAPP_volume_rendering=1
+)
+
```
agirault, line 67: Should we not pass CONFIGURE_OPTIONS and APP here as well?

utilities/testing/holohub.ctest
```diff
@@ -0,0 +1,77 @@
...
+# Copy the CDash configurations
+file(COPY
+     ${CMAKE_CURRENT_LIST_DIR}/CTestCustom.cmake
+     DESTINATION ${CTEST_BINARY_DIRECTORY})
+ctest_read_custom_files("${CTEST_BINARY_DIRECTORY}")
+
+set(configure_options
+  -DCMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc
+  -Dholoscan_DIR=${HOLOSCAN_BASE_DIR}/public/install/lib/cmake/holoscan
```
tbirdso, line 56: Slightly confused on how the `holoscan` installation is obtained, is this always from a source build+install tree? Where does the container mount occur?

`public` folder is an internal detail, remove?

utilities/testing/holohub.ctest
```diff
@@ -0,0 +1,77 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tbirdso, line 1: It's not immediately obvious how the `.ctest` files should be used, could you please add a brief description either as an inline header or in a `utilities/testing/README.md` file?

Based on first read I think these are:
- `holohub.container.ctest`: generic, can use with any individual application, intended to run in app container
- `holohub.ctest`: specific to our test machine configuration, tests multiple apps known to have matching environment requirements, targeting bare metal (or default HoloHub container?). Recommend that we keep this internal

utilities/cli/holohub.py
```diff
@@ -237,6 +237,24 @@ def _create_parser(self) -> argparse.ArgumentParser:
         )
         install.set_defaults(func=self.handle_install)
 
+        # Add test command
+        test = subparsers.add_parser("test", help="Test a project")
```
tbirdso, line 241: Nice!

tbirdso (changes requested):

utilities/cli/holohub.py
```diff
@@ -336,6 +355,73 @@ def handle_run_container(self, args: argparse.Namespace) -> None:
...
+
+        # Construct the ctest command line
+        # If we should run without xvfb
+        xvfb = "xvfb-run -a"
+        if args.no_xvfb:
+            xvfb = ""
+
+        base_img = args.base_img
+        if not base_img:
+            base_img = container.default_base_image()
```
tbirdso, line 388: Nitpick, could simplify:

```suggestion
        base_img = args.base_img or container.default_base_image()
```

utilities/testing/holohub.ctest
tbirdso: Following offline discussion I think we are aligned on removing this file as it's not currently exercised/tested, let me know if you disagree

tbirdso (approved):
LGTM, thanks @jjomier !

PR #864, opened 2025-05-21
Added note why we are pinning the container to HSDK 3.2

jjomier:


AndreasHeumann (approved):
Looks good, thanks for adding the comments.

PR #872, opened 2025-05-27
Isaac Sim Holoscan bridge

AndreasHeumann:
The Isaac Holoscan Bridge is a sophisticated application that demonstrates the integration between NVIDIA's Isaac Simulator and the Holoscan framework.

image

It creates a bridge where simulated data, including camera and robotic joint data, from Isaac Sim is processed through a Holoscan pipeline. The application features a virtual camera in Isaac Sim that captures video frames, which are then streamed to a Holoscan pipeline for real-time processing. The processed data is then again displayed in Isaac Sim. The application also streams joint positions from a robot to a Holoscan pipeline. The pipeline modifies these joint positions and feeds them back into the simulation moving the robotic arm.


applications/isaac_holoscan_bridge/README.md
```diff
@@ -0,0 +1,61 @@
+# Isaac Holoscan Bridge
+
+The Isaac Holoscan Bridge is a sophisticated application that demonstrates the integration between NVIDIA's Isaac Simulator and the Holoscan framework.
+
+<img src="IsaacSimHoloscanBridge.png" alt="isolated" width="800"/>
+
+It creates a bridge where simulated data, including camera and robotic joint data, from Isaac Sim is processed through a Holoscan pipeline. The application features a virtual camera in Isaac Sim that captures video frames, which are then streamed to a Holoscan pipeline for real-time processing. The processed data is then again displayed in Isaac Sim. The application also streams joint positins from a robot to a Holoscan pipeline. The pipeline modifies these joint positions and feeds them back into the simulation moving the robotic arm.
```
tbirdso:
```suggestion
It creates a bridge where simulated data, including camera and robotic joint data, from Isaac Sim is processed through a Holoscan pipeline. The application features a virtual camera in Isaac Sim that captures video frames, which are then streamed to a Holoscan pipeline for real-time processing. The processed data is then again displayed in Isaac Sim. The application also streams joint positions from a robot to a Holoscan pipeline. The pipeline modifies these joint positions and feeds them back into the simulation moving the robotic arm.
```

applications/isaac_holoscan_bridge/README.md
```diff
@@ -0,0 +1,61 @@
...
+The pipeline includes several specialized operators:
+* an AsyncDataPushOp for handling data streaming from Isaac Sim to Holoscan
+* a CallbackOp which is handling data transfer from Holoscan to Isaac Sim
+* a SobelOp for image processing (edge detection)
+* a ControlOp for managing arm joint positions
+* and HolovizOp for visualization.
+
+A unique feature of this application is its bidirectional callback system, where data flows between the simulation and Holoscan pipeline through carefully orchestrated callbacks, enabling real-time interaction between the two systems.
+
+## Flowdiagram
```
tbirdso:
```suggestion
## Flow Diagram
```

applications/isaac_holoscan_bridge/Simulator.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+        material.CreateSurfaceOutput().ConnectToSource(shader.ConnectableAPI(), "surface")
+
+        # Create a plane for the dynamic texture
+        plane_path = "/World/plane"
+        plane_geom = UsdGeom.Plane.Define(stage, plane_path)
+        # Set the size, position and rotation of the plane
+        plane_geom.AddTranslateOp().Set(Gf.Vec3f(2.5, 0.5, 1.5))
+        plane_geom.AddRotateXYZOp().Set(Gf.Vec3f(90.0, 0.0, 26.0))
+        plane_geom.AddScaleOp().Set(Gf.Vec3f(self._image_size[1] / self._image_size[0], 1, 2.0))
```
tbirdso, line 154: How were these derived?

applications/isaac_holoscan_bridge/app.py.save
```diff
@@ -0,0 +1,159 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tbirdso: Could you comment on the purpose of this file? Backup file accidentally added to tracking?

applications/isaac_holoscan_bridge/Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+RUN ./python.sh -m pip install holoscan==${HOLOSCAN_VERSION}
+
+# Isaac Sim is using its own Python installation, so we install alternatives to use that with the normal `python` and `python3` commands.
+RUN update-alternatives --install /usr/bin/python python /isaac-sim/python.sh 100 && \
+    update-alternatives --install /usr/bin/python3 python3 /isaac-sim/python.sh 100
+
+# update the SCRIPT_DIR to use the realpath which handles symlinks
+RUN sed -i 's/SCRIPT_DIR="\$( cd "\$( dirname "\${BASH_SOURCE\[0\]}" )" && pwd )"/SCRIPT_DIR=$( dirname -- "\$( readlink -f -- "\$0"; )"; )/' /isaac-sim/python.sh
+
+# patch the camera device to return data in CUDA memory
```
tbirdso, line 59: Do you know if this will be enabled out of the box in some future Isaac Sim release, or will a patch always be required?

applications/laser_detection_latency/CMakeLists.txt
```diff
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 3.24)
```
tbirdso, line 16: Related to this app?

dev_container
```diff
@@ -925,6 +925,7 @@ build_and_run_desc() { c_echo 'Build and run a requested application in a Docker
             If `--img` is not specified then a custom image tag will be defined.
         --img : Specify fully qualified output container name
         --container_args: Additional Docker run args.
+        --container_launch_args: Additional Docker launch args.
```
tbirdso:
```suggestion
        --container_launch_args: Launch args to forward to "dev_container launch" script command
```

tbirdso (commented):
Very nice! Left a few minor comments

applications/isaac_holoscan_bridge/Simulator.py
ghost: By interactively rotating and placing the plane which shows the image in Isaac Sim. I then noted the translation and rotation parameters and applied them here. The scale is set to keep the image aspect ratio.

applications/isaac_holoscan_bridge/app.py.save
ghost: Ops, should not be here. I will remove it.

applications/isaac_holoscan_bridge/Dockerfile
ghost: This is fixed and will part of IsaacSim 5.0.

applications/laser_detection_latency/CMakeLists.txt
ghost: Yes, sort of. The problem is that the docker file is based on Ubuntu 22.04 which has CMake 3.22 (I think) and compilation fails. We have a few applications which have side effects although they are not enabled. This can be seen when deleting the build directory and building just one application. After building just this one application there are several directories for applications created in the build directory.

tbirdso: > We have a few applications which have side effects although they are not enabled.

Ah gotcha. Agreed with this fix to address the side effect, thanks!

ghost:
Thanks @tbirdso. I updated the change.

Nic-Ma:
Hi @AndreasHeumann ,

Thanks for putting together this `IsaacHoloscanBridge` workflow.
Our typical use case is: Run GR00T N1 model and IsaacLab simulation together, so we need to pass the camera and robot joint positions to GR00T N1 model through this bridge, and send back the next step actions.
Currently, we are using DDS communication to connect these 2 parts in 2 Linux processes.
I think we need to change the `IsaacSim` logic in this PR to `IsaacLab` APIs in our use case.
Is there any other gap?

Thanks.

ghost:
Hi @Nic-Ma,
> I think we need to change the IsaacSim logic in this PR to IsaacLab APIs in our use case.

Could you please explain that in more detail? I understand that you run GROOT N1 and IsaacLab in separate processes and use DDS for communication. How do you envision Holoscan to help in that case?
Could you please point me to the GROOT N1 - IsaacLab use case?
This PR shows how to implement data exchange between Isaac SIM and Holoscan when both run in the same process.

Nic-Ma:
> Hi @Nic-Ma,
> 
> > I think we need to change the IsaacSim logic in this PR to IsaacLab APIs in our use case.
> 
> Could you please explain that in more detail? I understand that you run GROOT N1 and IsaacLab in separate processes and use DDS for communication. How do you envision Holoscan to help in that case? Could you please point me to the GROOT N1 - IsaacLab use case? This PR shows how to implement data exchange between Isaac SIM and Holoscan when both run in the same process.

Hi @AndreasHeumann ,

Let's sync offline about the details.

Thanks.

dev_container
jjomier, line 1: dev_container is deprecated, we should use the holohub script instead.

applications/laser_detection_latency/CMakeLists.txt
jjomier: I'm not sure I understand the issue. What are the steps to reproduce?

tbirdso: The parent [CMakeLists.txt](https://github.com/nvidia-holoscan/holohub/blob/main/applications/laser_detection_latency/CMakeLists.txt#L16) is always included on each config regardless of whether the `laser_detection` project is enabled. This adds a check for CMake >=3.24 to every HoloHub build for every project.

This hasn't typically been an issue because the Holoscan SDK dev container uses CMake 3.24. This project has now encountered the issue because it is using a different container with CMake 3.22.

The fix simply removing the check allows configuration to succeed with CMake 3.22. The check was redundant for the laser application and is retained in the subdirectory [CMakeLists.txt](https://github.com/nvidia-holoscan/holohub/blob/main/applications/laser_detection_latency/laser_detection/CMakeLists.txt#L16).

applications/isaac_holoscan_bridge/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required(VERSION 3.20)
+
+project(isaac_holoscan_bridge LANGUAGES NONE)
```
jjomier, line 18: Can we perform an automated integration test?

dev_container
ghost: Ok, I'll look into that. Does that mean that in the README all new applications should use the holohub script for launching the app?

applications/isaac_holoscan_bridge/CMakeLists.txt
ghost: Sure, I'll add one.

dev_container
ghost: Just asking because the top-level Readme is still using the dev_container script.

ghost: It seems there is no equivalent to the `./dev_container build_and_run isaac_holoscan_bridge`. If I use `./holohub run-container isaac_holoscan_bridge` it builds the container, builds the app, and launches the container. But it does not launch the app. And since the container I use has no default entry point, the launch fails.
@jjomier Any suggestions?

applications/isaac_holoscan_bridge/CMakeLists.txt
ghost: I added a test but unfortunately it's not that useful since there is segfault in Kit when shutting down. I've found no way in CMake to ignore the segfault so I added `SKIP_RETURN_CODE 1`.

dev_container
tbirdso: Hi @AndreasHeumann , please try `./holohub run isaac_holoscan_bridge` as the approximate equivalent of `./dev_container build_and_run`. `./holohub run` will build+run the container as setup steps by default if running on host before then building+running the app in side the container. For more commands and help: `./holohub --help`, `./holohub run --help`

We may need to expose an additional argument in `./holohub run` or `./holohub run-container` to support launching the container as root, to reflect the `dev_container` addition in this PR ([ref](https://github.com/nvidia-holoscan/holohub/pull/872/files#diff-3f43fb9b7168437cd50168bbf18a034e3d02b7f6c74026dcd8d1739656f60a5cR47)). @wyli  maybe you could comment on whether this is feasible short term?

@wyli  is working on `./holohub run` support for [mode presets](https://github.com/nvidia-holoscan/holohub/pull/901) which will support this information out of the box and help abstract details from new users. This contribution seems like a good candidate to use that feature in a future update.

ghost: Hi @tbirdso 
Okay, I tried `./holohub`. It sort of worked but there are some issues. The command I used was
```
./holohub run --docker-opts="-u root -e ACCEPT_EULA=Y -e PRIVACY_CONSENT=Y \
-v ${HOME}/docker/isaac-sim/cache/kit:/isaac-sim/kit/cache:rw \
-v ${HOME}/docker/isaac-sim/cache/ov:/root/.cache/ov:rw \
-v ${HOME}/docker/isaac-sim/cache/pip:/root/.cache/pip:rw \
-v ${HOME}/docker/isaac-sim/cache/glcache:/root/.cache/nvidia/GLCache:rw \
-v ${HOME}/docker/isaac-sim/cache/computecache:/root/.nv/ComputeCache:rw \
-v ${HOME}/docker/isaac-sim/logs:/root/.nvidia-omniverse/logs:rw \
-v ${HOME}/docker/isaac-sim/data:/root/.local/share/ov/data:rw \
-v ${HOME}/docker/isaac-sim/documents:/root/Documents:rw" isaac_holoscan_bridge
```
* `-u root` is used for the missing `--as_root` option (@wyli)
* `./holhub` hard coded the use of `Ninja` in `_build_project_locally()`. `Ninja` is not installed in the Isaac Sim contaienr and should be optional and only enabled when Ninja is installed. (@wyli)
* The `dev_container` script used to call `xhost` which is needed for X11. This is not called by `./holohub` and should be added to `get_display_options()`. (@wyli)

With these changes the launch worked.

tbirdso: Thanks @AndreasHeumann  for the notes! We've added these tasks to internal tracking (cc @wyli ). Let's move forward with `./dev_container build_and_run` for now to unblock.

tbirdso (approved):
Approved, confirmed internally that we are ready to check in this contribution.

dev_container
wyli: thanks @AndreasHeumann (somehow I missed these useful feedback earlier)

applications/isaac_holoscan_bridge/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+- On a [Holohub supported platform](../../README.md#supported-platforms)
+- Python 3.10+
+- Nvidia GPU with at least 8GB memory
+
+## Run Instructions
+
+```bash
+./dev_container build_and_run --container_launch_args "--as_root" --container_args "-e ACCEPT_EULA=Y -e PRIVACY_CONSENT=Y \
+-v ${HOME}/docker/isaac-sim/cache/kit:/isaac-sim/kit/cache:rw \
```
wyli, line 48: I might have missed some docs here -- these commands seem to assume some local contents such as `${HOME}/docker/isaac-sim/cache/kit` before the app?

ghost: Hi @wyli,
These directories are used to cache compiled shaders and reduce the startup delay. The application writes to them but they don't contain any data needed by the application.
Docker automatically creates these directories if they don't already exist.

PR #877, opened 2025-05-28
add quick run section

wyli:
the quick run has been merged into 'mode' as discussed

> an example of adding support of quick run in an application's metadata.json , and using the information to automatically update its readme and webpage.
> 
> to update the readme (this could be done via github action to ensure single-source truth from the metadata.json):
> ```
> python utilities/metadata/metadata_to_readme.py -p applications/body_pose_estimation
> ```
> 
> output example:
> Screenshot 2025-05-29 at 09 02 48
> 
>

### DIFF
--- a/applications/body_pose_estimation/metadata.json
+++ b/applications/body_pose_estimation/metadata.json
@@ -58,7 +58,7 @@
 			"dds": {
 				"description": "Run with DDS video stream input",
 				"build": {
-					"op_depends": [
+					"target_depends": [
 						"dds_video_subscriber",
 						"dds_video_publisher"
 					]


applications/body_pose_estimation/metadata.json
```diff
@@ -34,14 +34,50 @@
...
 				}
 			]
 		},
-		"run": {
-			"command": "python3 <holohub_app_source>/body_pose_estimation.py",
-			"workdir": "holohub_bin"
+		"modes": {
+			"default": {
+				"description": "Run with camera input (default)",
+				"hardware": ["camera"],
```
tbirdso, line 44: To confirm, is this field just for user reference for the time being? The alternative of doing some automatic check for hardware sounds nice, but seems both difficult to implement consistently across hardware and out of scope.

applications/body_pose_estimation/metadata.json
```diff
@@ -34,14 +34,50 @@
...
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"args": "--source replayer"
+				}
+			},
+			"dds": {
+				"description": "Run with DDS video stream input",
+				"build": {
+					"op_depends": [
```
tbirdso: Updated internal doc to suggest `target_depends`

```suggestion
					"target_depends": [
```

applications/body_pose_estimation/metadata.json
```diff
@@ -34,14 +34,50 @@
...
+			},
+			"video_device": {
+				"description": "Run with specific camera device, replace 1 with the device number accordingly",
+				"hardware": ["camera"],
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"args": "--video_device /dev/video1"
+				}
+			}
```
tbirdso, line 80: For general parameterization, I would suggest that the app documents `--run_args` in its README rather than enumerating args in mode presets. I.e. in README:

> To run with a specific camera device, pass the video device as an argument to the default app mode: `./holohub run body_pose_estimation --run_args "--video_device /dev/video1"`

```suggestion
```

wyli: My understanding is that the args here will enable convenient 'mode run' such as `./holohub run body_pose_estimation video_device`, or do you expect something different user experience?

applications/body_pose_estimation/metadata.json
wyli: Yes, the idea is mainly for user reference and it could be a flag for the automated integration test as well.

Issue #879, opened 2025-05-29
dev_container does not work


kimullaa:
When I tried run demo, error was happened.(`/bin/dash: 1: ffmpeg: not found`)

```
2025-05-29 05:34:29 $./dev_container launch --img holohub:yolo_model_deployment --docker_opts --entrypoint=bash  -- -c ./run build yolo_model_deployment
2025-05-29 05:34:29 $ xhost +local:docker
non-network local connections being added to access control list
[...]

PR #880, opened 2025-05-29
Add set -euo pipefail for safer error handling

kimullaa:
Fixes: #879


agirault (commented):
Fully in favor in terms of intent. The only risk is if some variables were not tested and we're missing some handling of "allowed to fail". IMO I prefer to fail loudly and we fix that as it comes, than fail quietly and it will go wrong later, so the change is worth the risk, but would like to defer to @jjomier and @tbirdso to approve since they're the main authors.

jjomier:
Note that the `run `script is going to go away and replaced by `holohub` script instead.

kimullaa:
@agirault thank you for a comment!

> Fully in favor in terms of intent. The only risk is if some variables were not tested and we're missing some handling of "allowed to fail". 

good points! Besides I tested some case of cli subcommand.but maybe it is not enough.

@jjomier Any plans on when to replace run to holohub command? At least some dockerfiles still seem to use run script(main branch)

jjomier:
@kimullaa we are working on it. Hopefully without a couple of weeks.

run
```diff
@@ -831,27 +823,21 @@ build() {
...
   if [ $benchmark == true ] && [ $is_app == true ]; then
     app_source_root_path=$(get_app_source_root_dir $1)
-    run_command benchmarks/holoscan_flow_benchmarking/restore_application.sh ${app_source_root_path}
+    run_command benchmarks/holoscan_flow_benchmarking/restore_application.sh ${app_source_root_path} || true
   fi
 
   if [ $is_pkg == true ]; then
     for cpack_config in $(find ${build_path}/pkg -name "CPackConfig-*.cmake"); do
-      run_command cpack --config "$cpack_config" -G "${pkg_generator}"
+      run_command cpack --config "$cpack_config" -G "${pkg_generator}" || true
```
tbirdso: Seems like we should error out if packaging is requested and fails

```suggestion
      run_command cpack --config "$cpack_config" -G "${pkg_generator}"
```

tbirdso (commented):
Thanks for the update, `set -euo pipefail` looks reasonable to me, have not tested locally yet

bhashemian:
Hi @kimullaa, thanks for your contribution!

As you may already know, `./run` has been deprecated in favor of Holohub’s new CLI (`./holohub`). Therefore, I’m closing this PR.

If your problem persists and you’d like to contribute to resolving it, please open a new PR. Thanks!

PR #882, opened 2025-05-29
Added lazy import of modules for endoscopy tool tracking python

jjomier:
The endoscopy tool tracking application has some modules that can be compiled/enabled optionally.
Adding lazy_import helps with loading the modules only when used.

### DIFF
--- a/applications/endoscopy_tool_tracking/python/endoscopy_tool_tracking.py
+++ b/applications/endoscopy_tool_tracking/python/endoscopy_tool_tracking.py
@@ -13,7 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import importlib.util
 import os
+import sys
 from argparse import ArgumentParser
 
 from holoscan.core import Application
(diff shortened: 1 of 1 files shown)


applications/endoscopy_tool_tracking/python/endoscopy_tool_tracking.py
```diff
@@ -23,27 +25,41 @@
...
+        ImportError: If the specified module cannot be found
+    """
+    spec = importlib.util.find_spec(module_name)
+    if spec is None:
+        raise ImportError(f"Module {module_name} not found")
+    module = importlib.util.module_from_spec(spec)
+    loader = importlib.util.LazyLoader(spec.loader)
+    loader.exec_module(module)
+    sys.modules[module_name] = module
+    return module
```
agirault, line 62: Oh that's cool, I didn't know about `LazyLoader`. This is what I did back then for the SDK, which works fine but using LazyLoader could be helpful too: https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/python/holoscan/__init__.py.in#L103-L133

Issue #558, opened 2024-10-29
`./run build --benchmark` is a no-op for some Python apps


tbirdso:
## Overview

`./run build --benchmark` is a no-op for pure Python apps executing from the application source folder.

## Expected Behavior

Building the Multi-AI Ultrasound application with `./run build --benchmark` prepares it for latency logging with Holoscan Flow Benchmarking tools. The following command sequence logs latency results:
```
./run build multiai_ultrasound python --benchmark
python3 ./benchmarks/holoscan_flow_tracking/benchmark.py -a multiai_ultrasound --language python
```

## Observed Behavior

- Building a C++ app with `./run build --benchmark` successfully patches and compiles the app to inherit from `BenchmarkedApplication` for latency logging, then reverts source changes without rebuilding
- Building a Python app with `./run build --benchmark` patches the source file, runs CMake, and then reverts the source file. In the case of Multi-AI Ultrasound, `./run launch` [references the reverted source file](https://github.com/nvidia-holoscan/holohub/blob/main/applications/multiai_ultrasound/python/metadata.json#L32). As a result, `benchmark.py -a multiai_ultrasound` is not prepared for latency logging.

## Suggested Changes
- Consider updating `./run build --benchmark` to detect whether the app is a Python app. If so, skip the source revert step to allow execution from the source folder.
- In addition or as an alternative, update Multi-AI Ultrasound and possibly other Python apps to copy source files to the build folder and then execute from the build folder. Similar to C++ apps, this would preserve patches for benchmarking after reverting source files.
- Document desired behavior in [`./run build_desc`](https://github.com/nvidia-holoscan/holohub/blob/main/run#L538) and possibly the Holoscan Flow Benchmarking project README.

sohamm17:
Great point, I prefer the second option to the first one, as this standardizes the behavior across C++ and python. One downside is that it would be a big change, as all apps might need to change.

PR #887, opened 2025-05-30
fixes python benchmarking script patching

wyli:
the benchmarking of python source code current doesn't work as expected, the root cause is that the command `./run build <app> --benchmark` will patch and then restore the source code, after running this command, the python application is not ready for the benchmarking command: `python benchmarks/holoscan_flow_benchmarking/benchmark.py`.

fixes https://github.com/nvidia-holoscan/holohub/issues/558

the PR provides additional patching of the python files within the benchmark.py on the fly.

possible alternative solutions:
- remove the auto restore step after build:
https://github.com/nvidia-holoscan/holohub/blob/6adefbef01740517db66261588b730e89b2c8bde/run#L847-L850
the user will have to manually restore the application code.

- patch the metadata.json run command when benchmarking, it's tricky as the app entry points are currently defined in various ways with different folder structure assumptions:

https://github.com/nvidia-holoscan/holohub/blob/6adefbef01740517db66261588b730e89b2c8bde/applications/colonoscopy_segmentation/metadata.json#L37

https://github.com/nvidia-holoscan/holohub/blob/6adefbef01740517db66261588b730e89b2c8bde/applications/adv_networking_bench/python/metadata.json#L36

https://github.com/nvidia-holoscan/holohub/blob/6adefbef01740517db66261588b730e89b2c8bde/applications/aja_video_capture/python/metadata.json#L32

https://github.com/nvidia-holoscan/holohub/blob/6adefbef01740517db66261588b730e89b2c8bde/applications/endoscopy_out_of_body_detection/python/metadata.json#L38


benchmarks/holoscan_flow_benchmarking/patch_application.sh
```diff
@@ -47,27 +47,5 @@ for file in $cpp_files; do
...
-    import_line=$(grep -nE -- "import" "$file" | head -n 1 | awk -F ':' '{print $1}')
-    sed -i "$import_line i import sys\nsys.path.append(\"$(dirname "$(readlink -f "$0")")\")\nfrom benchmarked_application import BenchmarkedApplication\n" "$file"
-
-    # Find "(Application)" and change it to "(BenchmarkedApplication)"
-    sed -i 's/(Application)/(BenchmarkedApplication)/g' "$file"
-
-    echo "Patched $file. Original file is backed up in $file.bak. diff -u $file $file.bak to see the changes."
-done
+# Use the common Python utility for patching instead of duplicating logic here
+python3 "$(dirname "$0")/patch_python_sources.py" "$dir"
```
tbirdso, line 51: Is there redundancy between patching Python sources here and also in `benchmark.py` that could be reduced?

tbirdso (commented):
Thanks @wyli , is the recommended approach for benchmarking still `./run <app_name> --benchmark` following these changes, or what is the current dev workflow?

benchmarks/holoscan_flow_benchmarking/patch_application.sh
sohamm17: @wyli Thanks for the update and fix! I agree that we can remove the python application patching from `patch_application.sh` and keep it dynamic in the `benchmark.py`. Then, the workflow for python application benchmarking would not do anything when we compile the application with `--benchmark`, but will dynamically patch the app in `benchmark.py`.
The instructions in `holoscan_flow_benchmarking` needs to updated accordingly.

wyli: I think patch_application.sh may still have to handle .py files because some python codes will be launched in `holohub_bin`

e.g. https://github.com/nvidia-holoscan/holohub/blob/6adefbef01740517db66261588b730e89b2c8bde/applications/aja_video_capture/python/metadata.json#L32-L33
(I somehow missed your comments earlier...)

tbirdso: > because some python codes will be launched in holohub_bin

Ok, I think I understand your point that `benchmark.py` currently always patches Python scripts in the application source dir whereas the actual Python script that runs might be in the build dir. Note that the script that runs might be in either <holohub_app_source> or <holohub_app_bin> in "command" (line 32) regardless of the workdir (line 33).

Would it be possible for `benchmark.py` to intelligently find and patch the Python script that will actually run, whether it is in the source or build directory?

Not sure how difficult that would be with the latest HoloHub CLI module, OK to keep the existing `patch_application.sh` and `benchmark.py` patching redundancy if it is the shortest path to cover both the source and build folders.

wyli: exactly, this PR shows the shortest path and the user experience is kept the same:
`./run build app --benchmark` then `python benchmark.py app ...` and benchmark.py internally calls `./run launch`.

we can probably rethink about it when we deprecate `./run` and design `./holohub app --benchmark` cc @jjomier

tbirdso: Thanks, maintaining status quo for now seems OK then:
- `./run build app --benchmark` patches the build dir (by way of patch source -> build may copy sources to build dir -> revert source)
- `benchmark.py` patches the source dir again before running

tbirdso (approved):
LGTM, let's wait for @sohamm17 's approval as well before merge

benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -42,24 +55,14 @@ $ ./benchmarks/holoscan_flow_benchmarking/patch_application.sh applications/endo
...
-$ ./run build <application name> <other options> --benchmark
-```
-
-Please make sure to test that the application runs correctly after building it, and before going to
-next steps of performance evaluation. You may also make sure that all the necessary TensorRT engine
-files are generated by running the application at least once, for example, for the endoscopy tool
-tracking application.
 
-3. **Run the performance evaluation**
+2. **Run the performance evaluation**
```
sohamm17, line 69: Kindly change the Step 4 to Step 3 below, as well. @wyli

sohamm17: Step 5 could be made optional.

sohamm17 (changes requested):
@wyli minor comments.

benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -42,24 +55,14 @@ $ ./benchmarks/holoscan_flow_benchmarking/patch_application.sh applications/endo
 ```
 This script saves the original `cpp` files in a `*.cpp.bak` file.
```
tbirdso:
```suggestion
This script saves the original source files (`.cpp` or `.py`) in a corresponding `*.bak` file.
```

benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -28,8 +28,21 @@ Extra dependencies might be needed to be installed to satisfy the requirements.
...
+```
+$ ./run build <application name> <other options> --benchmark
+```
+
+Please make sure to test that the application runs correctly after building it, and before going to
+next steps of performance evaluation. You may also make sure that all the necessary TensorRT engine
+files are generated by running the application at least once, for example, for the endoscopy tool
+tracking application.
+
+The command above patches the application for benchmarking.
```
tbirdso:
```suggestion
The command above patches the application for benchmarking. The benchmarking patch is applied to sources during the build, then source directory files are reverted after the build completes.
```

benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -42,24 +55,14 @@ $ ./benchmarks/holoscan_flow_benchmarking/patch_application.sh applications/endo
...
+2. **Run the performance evaluation**
 
 ```
 $ python benchmarks/holoscan_flow_benchmarking/benchmark.py -a <application name> <other options>
 ```
 
-The above command will run an application which is executed normally by 
+The above command will run an application which is executed normally by
 `./run launch <application name> cpp`. If an application is executed differently, then use the
 `--run-command` argument to specify the command to run an application.
```
tbirdso, line 77:
```suggestion
`--run-command` argument to specify the command to run an application.

**note**: If the target application is a Python file, `benchmark.py` will patch the application source in the local Holohub source directory for benchmarking before running the application.
```

tbirdso (commented):
Suggesting extra documentation following internal feedback from @sohamm17

benchmarks/holoscan_flow_benchmarking/README.md
sohamm17: LGTM, thanks!

benchmarks/holoscan_flow_benchmarking/README.md
wyli: sounds good, I've merged step 5 into the patching section so that it's easier to get the idea of both patching and restoring...

sohamm17 (approved):
LGTM

wyli:
fixes https://github.com/nvidia-holoscan/holohub/issues/558

PR #892, opened 2025-06-03
fixes CLI -- TypeError: 'staticmethod' object is not callable

wyli:
- fixes
```
main/utilities/cli/util.py", line 62, in Color
    red = _create_color_method(RED)
TypeError: 'staticmethod' object is not callable
``` 

- make `check_nvidia_ctk ` less strict so that we only need minimal setup to run the CLI and tests

### DIFF
--- a/.github/workflows/check_cli.yml
+++ b/.github/workflows/check_cli.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
   push:
-    branches: [main, holoscan-sdk-lws2, fixes-cli]
+    branches: [main, holoscan-sdk-lws2]
 
 jobs:
   check-cli-py:
--- a/.github/workflows/check_cli.yml
+++ b/.github/workflows/check_cli.yml
@@ -1,0 +1,30 @@
+name: Check CLI
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches: [main, holoscan-sdk-lws2, fixes-cli]
+
+jobs:
+  check-cli-py:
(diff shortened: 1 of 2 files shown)


.github/workflows/check_cli.yml
```diff
@@ -0,0 +1,30 @@
+name: Check CLI
```
wyli, line 1: since we're actively working on the CLI, I think it'll be useful to have extensive tests...is this a suitable frequency/setup to test these? @tbirdso @jjomier

.github/workflows/check_cli.yml
```diff
@@ -0,0 +1,30 @@
...
+  push:
+    branches: [main, holoscan-sdk-lws2]
+
+jobs:
+  check-cli-py:
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.9', '3.10', '3.11', '3.12']
```
jjomier, line 15: Should we test python 3.8?

utilities/cli/util.py
```diff
@@ -110,7 +109,8 @@ def check_nvidia_ctk() -> None:
     recommended_version = "1.14.1"
 
     if not shutil.which("nvidia-ctk"):
-        fatal("nvidia-ctk not found. Please install the NVIDIA Container Toolkit.")
```
jjomier, line 113: Why not keep fatal()?

.github/workflows/check_cli.yml
wyli: 3.8 is officially unsupported https://devguide.python.org/versions/ I think if there's issues we should encourage users to upgrade asap...

PR #896, opened 2025-06-04
Infra: Add `./holohub run --docker-opts` flag for container arguments

tbirdso:
Add support for forwarding container arguments to underlying container run when doing a containerized build/run with the Holohub CLI.

Supports projects requiring certain environment variables or volume mounts at runtime. Previous approach required two steps:
1. `./holohub run-container <myproject> --docker-opts <my_opts>`
2. `./holohub run <myproject>`

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -159,7 +159,7 @@ class HoloHubCLI:
             "--as_root", action="store_true", help="Run the container with root permissions"
         )
         run_container.add_argument(
-            "--docker_opts", help="Additional options to pass to the Docker launch"
+            "--docker_opts", default="", help="Additional options to pass to the Docker launch"
         )
         run_container.add_argument(
             "--language", choices=["cpp", "python"], help="Specify language implementation"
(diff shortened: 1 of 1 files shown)


wyli (approved):
Thanks! (We have mixed usage of underscore and hyphen in the CLI argument names, may need to unify them soon...)

utilities/cli/holohub.py
```diff
@@ -210,6 +210,11 @@ def _create_parser(self) -> argparse.ArgumentParser:
             dest="with_operators",
             help="Optional operators that should be built, separated by semicolons (;)",
         )
+        run.add_argument(
+            "--docker_opts",
```
bhashemian: @tbirdso, thanks so much for the update. I'm just wondering if there is any reason we are combining underscore with dashes. I think it makes more sense to have `--docker-opts` instead of `--docker_opts`. IMO, it would be easier to remember and type arguments which only have dashes (no underscores). Is there any benefit of mixing them?

tbirdso: Good point, underscores were carried over from legacy `run` and `dev_container` scripts. Updated here to use `-` instead of `_` in all commands except `test` to avoid CI breakage, will follow up with that change in a separate PR.

bhashemian: Thanks for the update.

PR #898, opened 2025-06-05
fixes cli, add tests

wyli:
addressed recently found cli issues:
- py39 TypeError: 'type' object is not subscriptable
- `--build_with` -> `--build-with`
- cmakelint usage in `./holohub lint`
- undefined variable `cuda_version`
- check_nvidia_ctk depends on 'packaging'
- add check_cli tests on github workflow

### DIFF
--- a/.github/workflows/check_cli.yml
+++ b/.github/workflows/check_cli.yml
@@ -1,0 +1,48 @@
+name: Check CLI
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+  push:
+    branches: [main, holoscan-sdk-lws2]
+
+jobs:
+  check-cli-py:
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.9', '3.10', '3.11', '3.12']
+
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Run Python tests
+        run: |
+          python -m unittest utilities.cli.tests.test_cli
+          python -m unittest utilities.cli.tests.test_container
+
+  check-cli-ctest:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Install CMake
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake
+
+      - name: Configure CMake with testing enabled
+        run: |
+          cmake -B build -DBUILD_TESTING=ON -DBUILD_HOLOHUB_TESTING=ON .
+
+      - name: Run CMake tests
+        run: |
+          cd build
+          ctest --verbose
--- a/.github/workflows/check_cli.yml
+++ b/.github/workflows/check_cli.yml
@@ -6,6 +6,9 @@ on:
   push:
     branches: [main, holoscan-sdk-lws2]
 
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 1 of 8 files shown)


utilities/cli/tests/test_cli.py
```diff
@@ -218,9 +228,12 @@ def test_lint_fix_command(self):
         args = self.cli.parser.parse_args("lint --fix".split())
         try:
             args.func(args)
+        except FileNotFoundError as e:
+            self.assertIn("ruff", str(e))  # if not installed, it complains about ruff
```
tbirdso: Could you explain this check? Do we really want to test linting without all lint dependencies installed?

utilities/cli/util.py
```diff
@@ -119,11 +116,25 @@ def check_nvidia_ctk() -> None:
...
         if match:
             version = match.group(1)
-            from packaging import version as ver
 
-            if ver.parse(version) < ver.parse(min_version):
+            try:
+                from packaging import version as ver
+
+                version_check = ver.parse(version) < ver.parse(min_version)
+            except ImportError:
```
tbirdso, line 124: Thanks for updating this!

tbirdso (commented):
Very nice! Added one question, otherwise LGTM

.github/workflows/check_cli.yml
```diff
@@ -0,0 +1,48 @@
...
+          python-version: ${{ matrix.python-version }}
+
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+      - name: Run Python tests
+        run: |
+          python -m unittest utilities.cli.tests.test_cli
+          python -m unittest utilities.cli.tests.test_container
+
+  check-cli-ctest:
```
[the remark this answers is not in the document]
tbirdso: @wyli thoughts on this?

utilities/cli/tests/test_cli.py
wyli: I think both Filenotfounderror and systemexit==0 are expected behaviours, depending on whether the environment has ruff installed.

.github/workflows/check_cli.yml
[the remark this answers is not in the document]
wyli: looks like it's suggesting a more secure permission, I've updated to follow it

utilities/cli/tests/test_cli.py
tbirdso: Agreed that ruff is [required](https://github.com/nvidia-holoscan/holohub/blob/main/utilities/requirements.lint.txt#L7) for linting. I'm concerned that in its current state there exists a scenario where we fail to detect a real issue where:
1. `lint` dependencies including `ruff` are properly installed
2. `ruff` returns an unexpected `FileNotFoundError` during a `--fix` operation (i.e. a true failure)
3. The test assumes the issue is due to `ruff` not being installed and returns a false negative result

What do you think?

wyli: good point, I've added a condition in the test case to be more specific, let me know any other suggestions.

utilities/cli/tests/test_cli.py
```diff
@@ -218,9 +237,13 @@ def test_lint_fix_command(self):
         args = self.cli.parser.parse_args("lint --fix".split())
         try:
             args.func(args)
+        except FileNotFoundError as e:
+            if not is_ruff_available():
+                self.assertIn("ruff", str(e))  # if not installed, it complains about ruff
```
tbirdso: We should emit the error if ruff is available and `FileNotFoundError` occurs:

```suggestion
            if is_ruff_available():
                raise e
            else:
                self.assertIn("ruff", str(e))  # if not installed, it complains about ruff
```

tbirdso (approved):
Thank you @wyli !

PR #900, opened 2025-06-05
Added way to rename parent script for Holohub CLI

jjomier:
This allow to have other holohub repository with a script different from ./holohub

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -66,6 +66,11 @@ class HoloHubCLI:
         parser = argparse.ArgumentParser(
             description="HoloHub CLI tool for managing Holoscan applications and containers"
         )
+        parser.add_argument(
+            "--script_name",
+            default="./holohub",
+            help="Name of the script to run (default: ./holohub)",
+        )
         subparsers = parser.add_subparsers(dest="command", required=True)
 
         # Add create command
@@ -509,7 +514,7 @@ class HoloHubCLI:
             container.build()
 
             # Build command with all necessary arguments
-            build_cmd = f"./holohub build {args.project} --local"
+            build_cmd = f"{args.script_name} build {args.project} --local"
             if args.build_type:
                 build_cmd += f" --build-type {args.build_type}"
             if args.with_operators:
@@ -571,7 +576,7 @@ class HoloHubCLI:
                 if not build_dir.is_dir() and not args.dryrun:
                     holohub_cli_util.fatal(
                         f"The build directory {build_dir} for this application does not exist.\n"
-                        f"Did you forget to './holohub build {args.project}'?"
+                        f"Did you forget to '{args.script_name} build {args.project}'?"
                     )
 
             # Handle workdir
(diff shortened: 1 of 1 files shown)


utilities/cli/holohub.py
```diff
@@ -66,6 +66,11 @@ def _create_parser(self) -> argparse.ArgumentParser:
         parser = argparse.ArgumentParser(
             description="HoloHub CLI tool for managing Holoscan applications and containers"
         )
+        parser.add_argument(
+            "--script_name",
+            default="./holohub",
+            help="Name of the script to run (default: ./holohub)",
+        )
```
wyli, line 70: in addition to the argument input, there're a few ways detecting the name suggested:
```py
    def _detect_command_name(self) -> str:
        """
        Detect the actual command name being used by examining the command line.
        This allows the same holohub.py to work with different wrapper script names.

        Returns the command name (e.g., './holohub', './myproject') to use for
        internal container commands.
        """
        # Method 1: Check environment variable set by wrapper script
        if 'HOLOHUB_COMMAND_NAME' in os.environ:
            return os.environ['HOLOHUB_COMMAND_NAME']

        # Method 2: Check if there's a parent process that looks like our wrapper script
        try:
            import psutil
            current_process = psutil.Process()
            parent = current_process.parent()
            if parent and parent.name() in ['bash', 'sh']:
                # Get the command line of the parent bash process
                parent_cmd = parent.cmdline()
                if len(parent_cmd) >= 2:
                    # Find the script name from the bash command line
                    for arg in parent_cmd:
                        if arg.startswith('/') or arg.startswith('./'):
                            script_path = Path(arg)
                            if script_path.exists() and script_path.is_file():
                                # Check if this script calls holohub.py
                                try:
                                    with open(script_path) as f:
                                        content = f.read()
                                        if 'holohub.py' in content:
                                            return str(script_path)
                                except:
                                    pass
        except (ImportError, Exception):
            pass

        # Method 3: Look for wrapper scripts in the holohub root that call holohub.py
        try:
            potential_wrappers = []
            for file_path in self.HOLOHUB_ROOT.iterdir():
                if file_path.is_file() and not file_path.name.startswith('.'):
                    try:
                        # Check if file is executable or a script
                        if file_path.stat().st_mode & 0o111:  # executable bit
                            with open(file_path) as f:
                                content = f.read()
                                if 'holohub.py' in content and 'SCRIPT_DIR' in content:
                                    potential_wrappers.append(file_path)
                    except (PermissionError, UnicodeDecodeError):
                        continue

            # If we found wrapper scripts, prefer the one that matches common patterns
            if potential_wrappers:
                # Sort by preference: holohub first, then alphabetically
                potential_wrappers.sort(key=lambda x: (x.name != 'holohub', x.name))
                return f"./{potential_wrappers[0].name}"
        except Exception:
            pass

        # Method 4: Default fallback - return './holohub' for backward compatibility
        return "./holohub"
```

jjomier: Very good point. I like 1. it's cleaner and avoid adding an extra arg.
I don't think 2 and 3 are reliable. @tbirdso  thoughts?

tbirdso: I'm in favor of (1) as well.

PR #901, opened 2025-06-06
mode preset config

wyli:
this is to discuss the updated metadata mode:
- this is a demo metadata.json showing the possible args: https://github.com/nvidia-holoscan/holohub/blob/mode-preset/applications/body_pose_estimation/metadata.json
- implemented `./holohub list-modes body_pose_estimation `, output:
```
Metadata file: /home/wenqil/Documents/holohub/applications/body_pose_estimation/metadata.json

Available modes for project body_pose_estimation:

default: Body pose estimation with camera input (default)
  Hardware: camera
  Command: python3 <holohub_app_source>/body_pose_estimation.py

replayer: Body pose estimation with pre-recorded video (no camera required)
  Command: python3 <holohub_app_source>/body_pose_estimation.py

dds: Body pose estimation with DDS video stream input
  Build dependencies: dds_video_subscriber, dds_video_publisher
  Command: python3 <holohub_app_source>/body_pose_estimation.py

video_device: Body pose estimation with specific camera device, replace 'video1' with device accordingly
  Hardware: camera
  Command: python3 <holohub_app_source>/body_pose_estimation.py
```

- `./holohub run body_pose_estimation dds` will build and run the mode preset
- static default docker arguments are extracted to a json for further customization flexibility https://github.com/nvidia-holoscan/holohub/blob/mode-preset/utilities/cli/docker_config.json

### DIFF
--- a/applications/body_pose_estimation/metadata.json
+++ b/applications/body_pose_estimation/metadata.json
@@ -68,8 +68,9 @@
 					"docker_args": [
 						"-e ACCEPT_EULA='Yes with whitespace'",
 						"-e PRIVACY_CONSENT=Y",
-						"-v ~/docker/rti:/root/rti:ro",
-						"-v /dev:/dev"
+						"-v /tmp/docker/rti:/root/rti:ro",
+						"-v /dev:/dev",
+						"--detach"
 					]
 				}
 			},
--- a/applications/body_pose_estimation/metadata.json
+++ b/applications/body_pose_estimation/metadata.json
@@ -34,14 +34,50 @@
 			"hardware": [
 				{
 					"name": "camera",
-					"description": "This application requires a Video4Linux (V4L2) compatible device as input.",
-					"required": true
+					"description": "This application requires a Video4Linux (V4L2) compatible device as input."
 				}
 			]
 		},
-		"run": {
-			"command": "python3 <holohub_app_source>/body_pose_estimation.py",
-			"workdir": "holohub_bin"
+		"modes": {
+			"default": {
+				"description": "Run with camera input (default)",
+				"hardware": ["camera"],
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin"
+				}
+			},
+			"replayer": {
+				"description": "Run with pre-recorded video (no camera required)",
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"args": "--source replayer"
+				}
+			},
+			"dds": {
+				"description": "Run with DDS video stream input",
+				"build": {
+					"op_depends": [
+						"dds_video_subscriber",
+						"dds_video_publisher"
+					]
+				},
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"args": "--source dds"
+				}
+			},
+			"video_device": {
+				"description": "Run with specific camera device, replace 1 with the device number accordingly",
+				"hardware": ["camera"],
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"args": "--video_device /dev/video1"
+				}
+			}
 		}
 	}
 }
--- a/applications/body_pose_estimation/metadata.json
+++ b/applications/body_pose_estimation/metadata.json
@@ -40,7 +40,7 @@
 		},
 		"modes": {
 			"default": {
-				"description": "Run with camera input (default)",
+				"description": "Body pose estimation with camera input (default)",
 				"hardware": ["camera"],
 				"run": {
 					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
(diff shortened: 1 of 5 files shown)


applications/body_pose_estimation/metadata.json
```diff
@@ -34,14 +34,55 @@
...
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"cmd_args": ["--source replayer"]
+				}
+			},
+			"dds": {
+				"description": "Body pose estimation with DDS video stream input",
+				"build": {
+					"depends": ["dds_video_subscriber", "dds_video_publisher"],
```
tbirdso, line 61: Suggested `target_depends` in design doc discussion out of concern that `depends` is too open-ended, user might try to add PyPI/apt dependencies and see errors, etc. cc @agirault  for discussion

utilities/cli/container.py
```diff
@@ -246,6 +274,7 @@ def run(
         nsys_location: str = "",
         as_root: bool = False,
         docker_opts: str = "",
+        docker_args: List[str] = None,
```
tbirdso: Should we remove `docker_opts` in favor of `docker_args`? Seems redundant

utilities/cli/docker_config.json
```diff
@@ -0,0 +1,70 @@
+{
```
tbirdso, line 1: Could you comment on the pros and cons of maintaining docker configs outside of `container.py`? In particular:
1. Is the main goal readability and/or abstraction between CLI args and docker flags? Seems useful, just clarifying intent
1. Will developers and users ever provide their own configs in the future, or will `container.py` always reference this file?
2. Would we accept dev contributions to `docker_config.json` adding app-specific configuration options? My guess is no, we would encourage devs to add their configurations to their own app metadata.json modes instead
3. Would we accept dev contributions to `docker_config.json` tweaking existing options or expanding arguments for broadly used configs? My guess is again no, or only in limited cases where there are clear issues
4. This config file references the "modes" field, is there an advantage to implementing the "mode" schema such as inheritance? Or should we define a separate schema for flexibility? It looks like `container.py` is currently treating these fields as a list of settings for CLI-specific logic rather than inheritable properties for combination with project "modes", IMO we should pick one or the other

utilities/cli/docker_config.json
```diff
@@ -0,0 +1,70 @@
...
+		"display": {
+			"description": "Display server options for X11",
+			"run": {
+				"docker_args": [
+					"-v /tmp/.X11-unix:/tmp/.X11-unix",
+					"-e DISPLAY"
+				]
+			}
+		},
+		"conditional": {
```
tbirdso, line 59: The arguments provided in "conditional" don't seem to match the pattern laid out by other presets above, should each of these be a separate preset?

utilities/cli/docker_config.json
```diff
@@ -0,0 +1,70 @@
+{
+	"modes": {
```
tbirdso, line 2: Project metadata.json modes are exclusive, but here it seems like we will typically apply most of the docker arg collections below to a given container launch command

utilities/cli/holohub.py
```diff
@@ -349,12 +365,33 @@ def handle_build_container(self, args: argparse.Namespace) -> None:
...
+            mode_config = self._get_mode_config(project_data, args.mode)
+            build_config = mode_config.get("build", {})
+
+            # Apply mode-specific docker build args
+            build_docker_args = build_config.get("docker_args", [])
+            if isinstance(build_docker_args, list):
+                mode_build_args = " ".join(build_docker_args)
+
+        # Combine user-provided build args with mode-specific args
+        combined_build_args = args.build_args or ""
```
tbirdso, line 388: Nitpick, should we set build_args with `default=''` instead of trying to handle the `None` default?

utilities/cli/holohub.py
```diff
@@ -364,14 +401,44 @@ def handle_run_container(self, args: argparse.Namespace) -> None:
...
+            if isinstance(run_docker_args, list):
+                mode_docker_opts = " ".join(run_docker_args)
+
+        # Build container with mode-specific args
+        container.build(build_args=mode_build_args if mode_build_args else None)
+
+        # Combine user-provided docker opts with mode-specific opts
+        combined_docker_opts = args.docker_opts
+        if mode_docker_opts:
+            combined_docker_opts = f"{args.docker_opts} {mode_docker_opts}".strip()
```
tbirdso: Can we standardize with "docker opts" or "docker args" naming? Seems like they refer to the same thing?

utilities/cli/holohub.py
```diff
@@ -482,6 +571,17 @@ def _build_project_locally(
         if with_operators:
             cmake_args.append(f'-DHOLOHUB_BUILD_OPERATORS="{with_operators}"')
 
+        # Add mode-specific build dependencies
+        mode_config = self._get_mode_config(project_data, mode)
+        build_config = mode_config.get("build", {})
+
+        # Handle mode-specific build dependencies
+        if "depends" in build_config:
+            depends = build_config["depends"]
```
tbirdso, line 585: Suggest `target_depends` or some other name to reflect more specific nature of this field reflecting op/app/ext dependencies only, open for discussion
```suggestion
        if "target_depends" in build_config:
            depends = build_config["target_depends"]
```

applications/body_pose_estimation/metadata.json
```diff
@@ -34,14 +34,55 @@
...
+					"docker_args": ["--ssh default"]
+				},
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"cmd_args": ["--source dds"],
+					"docker_args": [
```
tbirdso, line 68: Some discussion with @jjomier  on whether `docker_args` belong in `build`/`run` field or should be a parent field to avoid possible duplication across build/run. I.e. many apps will have build container == run container, so the following format:
```
"docker": {
  "launch_args": ["-e MYVAR", "-e MYVAR2"]
}, "build": { ... },
  "run": { ... },
}
```

Would be more streamlined and less error-prone than re-definition under build and run:
```bash
"build": { 
  "docker_args": ["-e MYVAR", "-e MYVAR2"],
...
 },
  "run": { 
  "docker_args": ["-e MYVAR", "-e MYVAR2"],
...
 },
}
```

However, we still want to enable the second case as well in the event that the build container != run container.

What do we think about enabling both cases, such that docker args can be defined at the parent level (preferred) if applicable to both cases or at the build/run level if not common to both configs? Or does that risk introducing too much complexity? Open to thoughts.

tbirdso (commented):
Thanks @wyli ! Added comments for discussion.

utilities/cli/holohub.py
wyli: thanks! this is now addressed. probably there's some bugs I'll fix when writing tests.

utilities/cli/docker_config.json
wyli: mainly I'm thinking about the tool developer to provide their own config as the 'base config'. I also feel some overriding/inheritance mechanisms are needed otherwise the current metadata modes will be a bit rigid.

utilities/cli/docker_config.json
wyli: nice comment, I couldn't find a proper way to include these in a json, perhaps these type of 'dynamic' options should stay in the python script...

wyli:
thanks @tbirdso, another iteration on this is at https://github.com/nvidia-holoscan/holohub/pull/972

PR #902, opened 2025-06-06
fixes CLI setup pipe, default cmd

wyli:
- fixes support of `|` in run_command setup 
- fixes default HOLOHUB_CMD_NAME, `holohub` -> `./holohub`


utilities/cli/util.py
```diff
@@ -83,16 +83,19 @@ def fatal(message: str) -> None:
 
 
 def run_command(
-    cmd: List[str], dry_run: bool = False, check: bool = True, **kwargs
+    cmd, dry_run: bool = False, check: bool = True, **kwargs
```
tbirdso: How about `from typing import Union`, then:
```suggestion
    cmd: Union[str, List[str]], dry_run: bool = False, check: bool = True, **kwargs
```

tbirdso (approved):
Looks good! Minor type hint suggestion

tbirdso (approved):
🙏

PR #904, opened 2025-06-09
implement build package

wyli:
`./run build <package>` -> 

```
./holohub build pkg_name --local
./holohub build pkg_name
```

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -588,26 +588,24 @@ class HoloHubCLI:
             # Handle workdir
             workdir = run_config.get("workdir", "holohub_app_bin")
             if workdir == "holohub_app_source":
+                print(
+                    holohub_cli_util.format_cmd(
+                        "cd " + str(project_data.get("source_folder", "")), is_dryrun=args.dryrun
+                    )
+                )
                 if not args.dryrun:
                     os.chdir(project_data.get("source_folder", ""))
-                print(
-                    f"{Color.blue(holohub_cli_util.get_timestamp())} {Color.white('$')} {Color.green('cd ' + str(project_data.get('source_folder', '')))}"
-                )
             elif workdir == "holohub_bin":
+                print(holohub_cli_util.format_cmd("cd " + str(build_dir), is_dryrun=args.dryrun))
                 if not args.dryrun:
                     os.chdir(build_dir)
-                print(
-                    f"{Color.blue(holohub_cli_util.get_timestamp())} {Color.white('$')} {Color.green('cd ' + str(build_dir))}"
-                )
             else:  # default to app binary directory
                 target_dir = (
                     build_dir if language == "cpp" else project_data.get("source_folder", "")
                 )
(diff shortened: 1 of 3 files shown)


utilities/cli/holohub.py
```diff
@@ -460,26 +447,34 @@ def _build_project_locally(
...
             "cmake",
             "-B",
             str(build_dir),
             "-S",
             str(HoloHubCLI.HOLOHUB_ROOT),
-            "-G",
-            "Ninja",
+            "--no-warn-unused-cli",
+            f"-DPython3_EXECUTABLE={sys.executable}",
+            f"-DPython3_ROOT_DIR={os.path.dirname(os.path.dirname(sys.executable))}",
```
wyli, line 469: saw these from https://github.com/nvidia-holoscan/holohub/blob/a4095f840989f765d76be0c3ee979f24d6152542/run#L821 were they intentionally ignored when migrating from ./run to ./holohub?

tbirdso: Not intentional, makes sense to add these back, thanks

utilities/cli/holohub.py
```diff
@@ -58,7 +42,7 @@ class HoloHubCLI:
     DEFAULT_SDK_DIR = "/opt/nvidia/holoscan/lib"
 
     def __init__(self):
-        self.script_name = os.environ.get("HOLOHUB_CMD_NAME", "holohub")
+        self.script_name = os.environ.get("HOLOHUB_CMD_NAME", "./holohub")
```
tbirdso: Did previous "holohub" without "./" work, or resulted in an error?

tbirdso (approved):
LGTM!

utilities/cli/holohub.py
wyli: It's an error..

tbirdso: Ack. Thanks for fixing.

tbirdso: I guess CLI tests did not cover this case -- can we update them accordingly? Can be in a separate PR to avoid blocking this one. https://github.com/nvidia-holoscan/holohub/actions/runs/15535221215/job/43732552101

wyli: sure, let me try to add a test in https://github.com/nvidia-holoscan/holohub/pull/902 as it's trying to fix some misc issues as well. 

Mainly this error will be triggered when launching `holohub build`  in a container, example: https://github.com/nvidia-holoscan/holohub/blob/a4095f840989f765d76be0c3ee979f24d6152542/utilities/cli/holohub.py#L514-L527
it's not tested by the github ci because the github setup doesn't currently support nvidia-ctk.

wyli: including a literal check in https://github.com/nvidia-holoscan/holohub/pull/902 https://github.com/nvidia-holoscan/holohub/pull/902/commits/9aea872f702df5b6f0beaddd33b7a4cc20be89e9

Issue #888, opened 2025-05-30
Holohub CLI lacks Installation Inside the Container


bhashemian:
Currently, the only Holohub command that can install an app/op is `./dev_container build_and_install <APP|OP>`, which does not work when running inside the container.
We don't have `./run install` and `./holohub install` has not been implemented yet.

It would be great if we can prioritize implementation of `./holohub install`. Thanks

PR #905, opened 2025-06-09
implement install

wyli:
fixes https://github.com/nvidia-holoscan/holohub/issues/888

Port functionality from "dev_container build_and_install" to "holohub install" command

See also: "run build --install" legacy command


tbirdso (approved):
LGTM, verified with `./holohub install endoscopy_tool_tracking`

tbirdso:
FYI @mocsharp  we are migrating from `dev_container` and `run` scripts to a modernized `holohub` CLI, recommend migrating Holoscan CLI scripts to use `./holohub install` implemented here

PR #910, opened 2025-06-09
Fix an Error in Holohub Dockerfile

bhashemian:
Fix #909

PR #914, opened 2025-06-10
CLI docker build enhancements

wyli:
- support consistent docker build options such as `--base-img, --build-args` for various ./holohub commands ` build-container, run-container, run, build`
- dockerfile includes a python3 installation if not present. (the dockerfile seems to already assume python such as `/tmp/scripts/holohub setup` and `pip install`)
- add github ci tests for `./holohub build-container --base-img ubuntu:22.04` (so that we can verify dockerfile such as https://github.com/nvidia-holoscan/holohub/pull/910)
- improve argparse help message: the current `./holohub build --invalid_arg` doesn't print subparser usage
- fix an issue of passing `--run_args` to container's `--run-args` parser

### DIFF
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -1,0 +1,24 @@
+name: Check Docker
+
+on:
+  push:
+    branches: [main, holoscan-sdk-lws2]
+
+permissions:
+  contents: read
+
+jobs:
+  check-docker-build:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+
+      - name: Run Docker build
+        run: |
+          ./holohub build-container --base-img ubuntu:22.04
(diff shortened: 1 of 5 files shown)


tbirdso (approved):
Looks good for enabling extra granularity. If options prove to be too verbose/burdensome, we might consider moving a few of them into "metadata.json" mode presets and removing from CLI at a later time. Let's move forward with this to enable dev granularity for now.

.github/workflows/check_docker.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  push:
+    branches: [main, holoscan-sdk-lws2]
+
+permissions:
+  contents: read
+
+jobs:
+  check-docker-build:
+    runs-on: ubuntu-22.04
```
tbirdso, line 12: Suggest we add an extra test for ubuntu 24.04 as well as a followup

Issue #911, opened 2025-06-09
Holohub New CLI Does Not Support Benchmarking


bhashemian:
The new Holohub CLI `./holohub` does not support building for benchmarking as the legacy CLI:
```
./run build <application_name> [options] --benchmark
```

PR #916, opened 2025-06-10
Implement benchmarking CLI and various build/run args

wyli:
- enabling the benchmarking steps:

```
./holohub run-container endoscopy_tool_tracking
```
then within the container:
```bash
./holohub build endoscopy_tool_tracking --benchmark --language cpp
python benchmarks/holoscan_flow_benchmarking/benchmark.py -a endoscopy_tool_tracking --sched greedy -d myoutputs
```
or
```
python benchmarks/holoscan_flow_benchmarking/benchmark.py -a endoscopy_tool_tracking --sched greedy -d myoutputs --language python
```

- matching some dev_container flags
- the PR currently requires the changes in https://github.com/nvidia-holoscan/holohub/pull/914
- fix https://github.com/nvidia-holoscan/holohub/issues/911
- add `--no-docker-build` and `--no-local-build` and `HOLOHUB_ALWAYS_BUILD` to speed up running by skipping build

### DIFF
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2, impl-bench]
+    branches: [main, holoscan-sdk-lws2]
 
 permissions:
   contents: read
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -10,9 +10,6 @@ permissions:
 jobs:
   check-docker-build:
     runs-on: ubuntu-22.04
-    strategy:
-      matrix:
-        ubuntu-version: ["22.04"]
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2, impl-bench]
+    branches: [main, holoscan-sdk-lws2, impl-bench, temp-test]
 
 permissions:
   contents: read
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -16,9 +16,10 @@ jobs:
 
       - name: Run Docker build
         run: |
+          export ubuntu_version=22.04
           ./holohub build-container --dryrun --no-cache
           ./holohub build-container --dryrun --build-args "--build-arg TEST=value"
-          ./holohub build-container --base-img ubuntu:${{ matrix.ubuntu-version }}
+          ./holohub build-container --base-img ubuntu:${ubuntu_version}
           if ! command -v nvidia-ctk >/dev/null 2>&1; then
             sed -i 's/check_nvidia_ctk()/print("no check")/g' utilities/cli/container.py
             sed -i 's/            "--runtime=nvidia",/            # "--runtime=nvidia",/' utilities/cli/container.py
@@ -26,7 +27,7 @@ jobs:
             sed -i 's/            "all",/            # "all",/' utilities/cli/container.py
             echo "Replaced check_nvidia_ctk() call and commented out NVIDIA runtime args since nvidia-ctk is not available"
           fi
-          ./holohub run-container --base-img ubuntu:${{ matrix.ubuntu-version }} --no-docker-build
+          ./holohub run-container --base-img ubuntu:${ubuntu_version} --no-docker-build
           ./holohub run-container --dryrun --docker-opts "--memory 4g"
           ./holohub run-container --dryrun --add-volume "/tmp"
           ./holohub run-container --docker-opts "--memory 4g" --no-docker-build
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -34,4 +34,4 @@ jobs:
 
           ./holohub run-container --base-img ubuntu:${ubuntu_version} --no-docker-build | grep -q "docker run" || { echo "run-container failed"; exit 1; }
           ./holohub run-container --docker-opts "--memory 4g" --no-docker-build | grep -q "memory 4g" || { echo "docker-opts test failed"; exit 1; }
-          ./holohub run-container --no-x11 --no-docker-build --add-volume "/tmp" | grep -q "/tmp" || { echo "add-volume test failed"; exit 1; }
+          ./holohub run-container --no-docker-build --add-volume "/tmp" | grep -q "/tmp" || { echo "add-volume test failed"; exit 1; }
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, impl-bench]
 
 permissions:
   contents: read
@@ -10,15 +10,27 @@ permissions:
 jobs:
   check-docker-build:
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        ubuntu-version: ["22.04"]
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
 
       - name: Run Docker build
         run: |
-          ./holohub build-container --base-img ubuntu:22.04
-          if command -v nvidia-ctk >/dev/null 2>&1; then
-            ./holohub run-container --base-img ubuntu:22.04
-          else
-            echo "Skipping run-container: NVIDIA Container Toolkit not available"
+          ./holohub build-container --dryrun --no-cache
+          ./holohub build-container --dryrun --build-args "--build-arg TEST=value"
+          ./holohub build-container --base-img ubuntu:${{ matrix.ubuntu-version }}
+          if ! command -v nvidia-ctk >/dev/null 2>&1; then
+            sed -i 's/check_nvidia_ctk()/print("no check")/g' utilities/cli/container.py
+            sed -i 's/            "--runtime=nvidia",/            # "--runtime=nvidia",/' utilities/cli/container.py
+            sed -i 's/            "--gpus",/            # "--gpus",/' utilities/cli/container.py
+            sed -i 's/            "all",/            # "all",/' utilities/cli/container.py
+            echo "Replaced check_nvidia_ctk() call and commented out NVIDIA runtime args since nvidia-ctk is not available"
           fi
+          ./holohub run-container --base-img ubuntu:${{ matrix.ubuntu-version }} --no-docker-build
+          ./holohub run-container --dryrun --docker-opts "--memory 4g"
+          ./holohub run-container --dryrun --add-volume "/tmp"
+          ./holohub run-container --docker-opts "--memory 4g" --no-docker-build
+          ./holohub run-container --no-x11 --no-docker-build --add-volume "/tmp"
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -10,6 +10,9 @@ permissions:
 jobs:
   check-docker-build:
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        ubuntu_version: [22.04, 24.04]
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
@@ -17,7 +20,7 @@ jobs:
       - name: Run Docker build
         run: |
           set -o pipefail
-          export ubuntu_version=22.04
+          export ubuntu_version=${{ matrix.ubuntu_version }}
 
           ./holohub build-container --base-img ubuntu:${ubuntu_version} 2>&1 | tee /tmp/build.log
           grep -qE "writing image.* done" /tmp/build.log || { echo "Build failed - no success pattern"; cat /tmp/build.log; exit 1; }
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2, impl-bench, temp-test]
+    branches: [main, holoscan-sdk-lws2, impl-bench]
 
 permissions:
   contents: read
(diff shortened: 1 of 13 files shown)


benchmarks/holoscan_flow_benchmarking/benchmark.py
```diff
@@ -264,10 +325,15 @@ def main():
     if args.num_messages != 100:
         env["HOLOSCAN_NUM_SOURCE_MESSAGES"] = str(args.num_messages)
 
+    # Use the new holohub CLI command format
     if args.run_command == "":
-        app_launch_command = "./run launch " + args.holohub_application + " " + args.language
+        app_launch_command = (
+            f"./holohub run {args.holohub_application} --local --language {args.language}"
```
wyli, line 336: will we need a generic script name here from HOLOHUB_CMD_NAME?

.github/workflows/check_docker.yml
```diff
@@ -2,23 +2,26 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, impl-bench]
```
wyli: remove this temporary branch config 'impl-bench' before merging
```suggestion
    branches: [main, holoscan-sdk-lws2]
```

.github/workflows/check_docker.yml
```diff
@@ -2,23 +2,26 @@ name: Check Docker
...
+      matrix:
+        ubuntu-version: ["22.04", "24.04"]
     steps:
       - name: Checkout Repository
         uses: actions/checkout@v4
 
       - name: Run Docker build
         run: |
-          ./holohub build-container --base-img ubuntu:22.04
+          ./holohub build-container --base-img ubuntu:${{ matrix.ubuntu-version }}
```
wyli: enabling 24.04 gives an error from the [dockerfile](https://github.com/nvidia-holoscan/holohub/blob/main/Dockerfile):

```
#20 [12/17] RUN pip install meson
#20 0.330 error: externally-managed-environment
#20 0.330 
#20 0.330 × This environment is externally managed
#20 0.330 ╰─> To install Python packages system-wide, try apt install
#20 0.330     python3-xyz, where xyz is the package you are trying to
#20 0.330     install.
#20 0.330     
#20 0.330     If you wish to install a non-Debian-packaged Python package,
#20 0.330     create a virtual environment using python3 -m venv path/to/venv.
#20 0.330     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
#20 0.330     sure you have python3-full installed.
#20 0.330     
#20 0.330     If you wish to install a non-Debian packaged Python application,
#20 0.330     it may be easiest to use pipx install xyz, which will manage a
#20 0.330     virtual environment for you. Make sure you have pipx installed.
#20 0.330     
#20 0.330     See /usr/share/doc/python3.12/README.venv for more information.
#20 0.330 
#20 0.330 note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
#20 0.330 hint: See PEP 668 for the detailed specification.
#20 ERROR: process "/bin/sh -c pip install meson" did not complete successfully: exit code: 1
```

is there a preferred fix for this? cc @tbirdso @agirault  https://github.com/nvidia-holoscan/holohub/actions/runs/15562827974/job/43819290002

tbirdso: Could explore:
- Align to Holoscan SDK Dockerfile python3.12 installation: https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/Dockerfile#L316
- Probably OK to set [PIP_BREAK_SYSTEM_PACKAGES](https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-break-system-packages) for the scope of this installation

Is this in scope for benchmarking? Suggest exploring in a separate PR if possible without blocking benchmarking coverage

.github/workflows/check_docker.yml
```diff
@@ -2,23 +2,35 @@ name: Check Docker
...
 
 permissions:
   contents: read
 
 jobs:
   check-docker-build:
     runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        ubuntu-version: ["22.04"]
```
tbirdso: Redundant?

.github/workflows/check_docker.yml
```diff
@@ -2,23 +2,35 @@ name: Check Docker
...
+            sed -i 's/            "--runtime=nvidia",/            # "--runtime=nvidia",/' utilities/cli/container.py
+            sed -i 's/            "--gpus",/            # "--gpus",/' utilities/cli/container.py
+            sed -i 's/            "all",/            # "all",/' utilities/cli/container.py
+            echo "Replaced check_nvidia_ctk() call and commented out NVIDIA runtime args since nvidia-ctk is not available"
           fi
+          ./holohub run-container --base-img ubuntu:${{ matrix.ubuntu-version }} --no-docker-build
+          ./holohub run-container --dryrun --docker-opts "--memory 4g"
+          ./holohub run-container --dryrun --add-volume "/tmp"
+          ./holohub run-container --docker-opts "--memory 4g" --no-docker-build
+          ./holohub run-container --no-x11 --no-docker-build --add-volume "/tmp"
```
tbirdso: Any thoughts on how we could validate that the command (container) ran successfully rather than failing out silently?

utilities/cli/holohub.py
```diff
@@ -382,52 +449,95 @@ def _make_project_container(
         """Define a project container"""
         if not project_name:
             return HoloHubContainer(project_metadata=None)
-        project_data = self._find_project(project_name=project_name, language=language)
+        project_data = self.find_project(project_name=project_name, language=language)
         return HoloHubContainer(project_metadata=project_data)
 
+    def build_container_for_project(
```
tbirdso: Concerned that behavior of `build_container_for_project(..., skip_build=True)` is not intuitive. What do you think about splitting into `get` / `build` methods, or otherwise removing in favor of inlining at call sites? Or, thoughts on a different name to clarify get+build scope?

utilities/cli/holohub.py
```diff
@@ -1298,37 +1445,27 @@ def handle_install(self, args: argparse.Namespace) -> None:
...
                 install_cmd += f" --build-type {args.build_type}"
-            if getattr(args, "language", None):
-                install_cmd += f" --language {args.language}"
-            if getattr(args, "with_operators", None):
+            if hasattr(args, "with_operators") and args.with_operators:
                 install_cmd += f' --build-with "{args.with_operators}"'
-            if getattr(args, "parallel", None):
+            if hasattr(args, "language") and args.language:
+                install_cmd += f" --language {args.language}"
+            if hasattr(args, "parallel") and args.parallel:
```
tbirdso: Could you clarify why the `getattr` -> `hasattr` change is required?

utilities/cli/container.py
```diff
@@ -73,50 +75,113 @@ def device_args() -> List[str]:
         """Get docker run arguments for mounting devices in the container"""
         options = []
 
-        # Add device mounts
-        device_paths = [
+        for video_dev in glob.glob("/dev/video[0-9]*"):
+            options.extend(["--device", video_dev])
```
tbirdso, line 158: Is there an advantage to mounting individual devices over mounting `/dev`? General security?

Are the updated device mounts below based on known application needs? Not sure if we have audio general requirements?

benchmarks/holoscan_flow_benchmarking/benchmark.py
```diff
@@ -98,6 +99,50 @@ def run_command(app_launch_command, env):
...
+        source_folder = project_metadata.get("source_folder", "")
+        if source_folder and os.path.isdir(source_folder):
+            directories_to_patch.append(source_folder)
+        return directories_to_patch
+    command = run_config.get("command", "")
+    source_folder = project_metadata.get("source_folder", "")
+    if "<holohub_app_source>" in command and source_folder and os.path.isdir(source_folder):
+        directories_to_patch.append(source_folder)
+        logger.info(f"Will patch source directory: {source_folder}")
+    if "<holohub_app_bin>" in command:
```
tbirdso, line 130: Should we check `workdir` as well in case of relative execution? I.e. a project could have
```bash
		"run": {
			"command": "python3 body_pose_estimation.py",
			"workdir": "holohub_app_bin"
		}
```

.github/workflows/check_docker.yml
wyli: yes, let me remove this, we can add it back when testing both 22.04 and 24.04.

utilities/cli/holohub.py
wyli: different implementations exist for the similar checking logic, I was trying to consolidate a bit. after reviewing the overall 'args' attributes reading, now I think I'll just revert and we can discuss about consolidation later if needed.

utilities/cli/holohub.py
wyli: ok, removing looks good..

utilities/cli/container.py
wyli: mainly ported from the dev_container script: https://github.com/nvidia-holoscan/holohub/blob/98745dae2b8e0e7a9c19f05548158ba20ccda0c0/dev_container#L385-L395

.github/workflows/check_docker.yml
wyli: so I mainly added some logs checking in the updated config, there are additional ctests to check the dryrun commands, so I think their combination can detect some container errors now.

benchmarks/holoscan_flow_benchmarking/benchmark.py
wyli: added a `build_holohub_path_mapping` in `utilities/cli/util.py` to resolve the template variables

utilities/cli/container.py
```diff
@@ -453,7 +525,7 @@ def get_display_options(self, enable_x11: bool, ssh_x11: bool) -> List[str]:
                 )
 
         # Handle X11 forwarding
-        if enable_x11 or ssh_x11:
+        if not no_x11 or ssh_x11:
```
tbirdso: Preference to keep bool args as affirmative (`enable_x11`) rather then negative (`no_x11`) throughout to avoid double negatives, could be convinced otherwise, what do you think?

utilities/cli/util.py
```diff
@@ -467,3 +467,57 @@ def list_cmake_dir_options(script_dir: Path, cmake_function: str) -> List[str]:
...
+        path_mapping["holohub_app_source"] = str(app_source_path)
+    if build_dir:
+        path_mapping["holohub_bin"] = str(build_dir)
+        if app_source_path:
+            try:
+                app_build_dir = build_dir / Path(app_source_path).relative_to(holohub_root)
+                path_mapping["holohub_app_bin"] = str(app_build_dir)
+            except ValueError:
+                # Handle case where app_source_path is not relative to holohub_root
+                path_mapping["holohub_app_bin"] = str(build_dir)
```
tbirdso, line 500: Have we validated this outside of the main HoloHub repo, i.e. in a mini-Holohub repo? Can we?

benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -62,7 +62,7 @@ These python dependencies include:
 #### Automatic Build with Benchmarking
 
 ```bash
-./run build <application_name> [options] --benchmark
+./holohub build <application_name> [options] --benchmark
```
bhashemian, line 65: Thanks @wyli for enhancing the CLI! I just want to confirm that this is compatible with what we discussed with @tbirdso and @jjomier. While I prefer the current format with `--benchmark`, I want to ensure that it doesn’t conflict with our discussion of using `./holohub run <APP> benchmark`, which defines benchmark as a mode when running the application. Thanks

benchmarks/holoscan_flow_benchmarking/benchmark.py
```diff
@@ -107,26 +108,30 @@ def find_python_files_to_patch(project_metadata, holohub_root_path):
...
+    source_folder = project_metadata.get("source_folder", "")
     if not run_config:
         # If no run config, fall back to source directory only
-        source_folder = project_metadata.get("source_folder", "")
         if source_folder and os.path.isdir(source_folder):
             directories_to_patch.append(source_folder)
         return directories_to_patch
+
+    project_name = project_metadata.get("project_name", "")
+    build_dir = Path(holohub_root_path) / "build" / project_name if project_name else None
```
tbirdso: Should this be a failure if there is no project name to benchmark?

benchmarks/holoscan_flow_benchmarking/README.md
tbirdso: Fair question, IMO there is no conflict, the "benchmark" preset can be responsible for calling `./holohub build <app>` with `--benchmark` as a default build flag or otherwise handling the special case of patching the application for benchmarking at build time. @wyli  to confirm

benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -89,6 +94,7 @@ This command:
 ### Important notes
 
 - Verify the application runs correctly after building and before proceeding with performance evaluation.
+  For example, run the app `./holohub run endoscopy_tool_tracking --language python` (append `--local` if you are using the bare-metal installation or inside a container)
```
sohamm17: Is `--local` for bare-metal or inside a container? It's a bit confusing where to run the command?

wyli: good point, actually there's no need to have it when inside a container, I'll update it.

benchmarks/holoscan_flow_benchmarking/README.md
wyli: yes, no conflict, I feel the mode is flexible and compact, would potentially enable different benchmarking settings easily..

utilities/cli/util.py
wyli: not yet, I feel we can validate more when the use case is clearer? currently all the `project_data` that we are dealing with are from this 'find_project' function, which only support the primary repo or a repo in a very similar folder structure... https://github.com/nvidia-holoscan/holohub/blob/5db60cfc20d737c0dbd3863d981624223db5b8ab/utilities/cli/holohub.py#L385

utilities/cli/container.py
wyli: also changed the CLI option `--no-x11` to `--enable-x11` default true

utilities/cli/util.py
tbirdso: Agreed this doesn't need to block the PR. We should validate support in the near future, probably with CI coverage. Opened an internal ticket to track.

wyli:
Hi @sohamm17 before merging this, please let us know any comments on the updated benchmarking commands. The PR is mostly updating the entry point from `./run` and `./dev_container` to `./holohub`, and improve the source code patching logic using the CLI module.

wyli:
Hi @tbirdso, I don't receive any further concerns, could you help merge this? I can update with any follow-ups later if needed. thanks!

PR #921, opened 2025-06-12
Infra: Update default Python3 installation strategy

tbirdso:
Support installing `python3` in default `ubuntu:24.04` container

To address failure:
```
=> ERROR [12/17] RUN pip install meson                                                      0.4s 
------
 > [12/17] RUN pip install meson:
0.378 error: externally-managed-environment
0.378 
0.378 × This environment is externally managed
0.378 ╰─> To install Python packages system-wide, try apt install
0.378     python3-xyz, where xyz is the package you are trying to
0.378     install.
0.378     
0.378     If you wish to install a non-Debian-packaged Python package,
0.378     create a virtual environment using python3 -m venv path/to/venv.
0.378     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
0.378     sure you have python3-full installed.
0.378     
0.378     If you wish to install a non-Debian packaged Python application,
0.378     it may be easiest to use pipx install xyz, which will manage a
0.378     virtual environment for you. Make sure you have pipx installed.
0.378     
0.378     See /usr/share/doc/python3.12/README.venv for more information.
0.378 
0.378 note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
0.378 hint: See PEP 668 for the detailed specification.
```

Test with: `./holohub run-container --base_img ubuntu:24.04 --img holohub:24.04`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,28 @@ ARG DEBIAN_FRONTEND=noninteractive
 #
 
 # Install python3 if not present (needed for holohub CLI)
+ARG PYTHON_VERSION=python3.12
 RUN if ! command -v python3 >/dev/null 2>&1; then \
-        apt-get update && apt-get install -y python3 python3-pip; \
-    fi
+        apt-get update \
+        && apt-get install --no-install-recommends -y \
+            software-properties-common curl \
+        && add-apt-repository ppa:deadsnakes/ppa \
+        && apt-get update \
+        && apt-get install --no-install-recommends -y \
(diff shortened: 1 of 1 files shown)


tbirdso:
Added a minor fix for `ubuntu:22.04` base, validated with both `ubuntu:22.04` and `ubuntu:24.04`. Intends to support nightly testing on dGPU host with 24.04 base image.

PR #922, opened 2025-06-12
Adding holohub CLI for integration testing

jjomier:


tbirdso:
@jjomier  the CLI is undergoing rapid updates, could you please include the git revision from which CLI sources are forked in the commit message?

PR #924, opened 2025-06-13
CLI: run-container forward trailing commands

wyli:
Update HoloHub CLI to forward arguments following "--" to the docker run entrypoint.

example usage:
```
./holohub run-container endoscopy_tool_tracking --no-docker-build -- echo 'hello'
```


tbirdso (approved):
Looks good, agreed on brief review that `argparse` doesn't have a standard way to parse trailing arguments, I'm slightly surprised about that.

One comment on `sys.argv` override, risk seems minor, what do you think?

wyli:
> Looks good, agreed on brief review that `argparse` doesn't have a standard way to parse trailing arguments, I'm slightly surprised about that.
> 
> One comment on `sys.argv` override, risk seems minor, what do you think?

thanks, the closest I think it's 'nargs=argparse.REMAINDER' but it'll catch all the unrecognized arguments into the forwarding. (also updated to avoid modifying sys.argv directly)

PR #926, opened 2025-06-16
CLI implement vscode config

wyli:
example:
```
./holohub vscode
```


tbirdso (approved):
LGTM! Verified with `./holohub vscode endoscopy_tool_tracking`, able to build and run in the dev container and hit a breakpoint with GDB in VS Code.

cc @mocsharp  for viz

PR #927, opened 2025-06-17
CLI add --configure-args, environ var

wyli:
- add support of `--configure-args='-DCUSTOM_OPTION=ON' --configure-args='-Dtest=ON'`
- add default path before the building commands `"/usr/local/cuda/bin", "/opt/cuda/bin"`
- exclude template dir in `./holohub list`


utilities/cli/holohub.py
```diff
@@ -503,11 +521,27 @@ def build_project_locally(
...
+        for cuda_path in cuda_paths:
+            if Path(cuda_path).exists() and cuda_path not in current_path:
+                current_path = f"{current_path}:{cuda_path}"
+                if dryrun:
+                    print(
+                        holohub_cli_util.format_cmd(
+                            f"export PATH=${{PATH}}:{cuda_path}", is_dryrun=True
+                        )
+                    )
+        env["PATH"] = current_path
```
wyli: in the local build, these changes are meant to fix 
```
The CUDA compiler identification is unknown
CMake Error at operators/tool_tracking_postprocessor/CMakeLists.txt:17 (project):
  No CMAKE_CUDA_COMPILER could be found.
``` 
do we need both `-DCMAKE_PREFIX_PATH` and `PATH` to have the cuda path or there's a preference for one of them? I think the ./run script relies on nvcc being on $PATH and doesn't explicitly set cmake_prefix_path. cc @tbirdso @jjomier

tbirdso: Details on CMake's CUDA Toolkit search strategy are here: [FindCUDAToolkit](https://cmake.org/cmake/help/v3.24/module/FindCUDAToolkit.html#search-behavior)

I'd suggest that we should consider NVCC setup as an environment prerequisite and shouldn't handle it in the HoloHub CLI. The `holoscan` apt package for C++ dev depends on the `cuda-nvcc-12-6` apt package as a prerequisite. If we are building `./holohub build endoscopy_tool_tracking --local` then I would expect `nvcc` is already available in one of the documented search locations:
- `PATH`
- `/usr/local/cuda`
- `/usr/local/cuda-X.Y`

Could you please provide more details on the environment in which `No CMAKE_CUDA_COMPILER could be found.` was observed? I can follow up on the internal bug.

wyli: not sure how things were installed beforehand but some debugging logs using the main branch ./holohub (the following error was fixed by manually adding -DCMAKE_PREFIX_PATH):

```
$ echo $PATH
/home/swqa/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin

$ which nvcc
/usr/local/bin/nvcc

$ nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2024 NVIDIA Corporation
Built on Wed_Oct_30_00:08:18_PDT_2024
Cuda compilation tools, release 12.6, V12.6.85
Build cuda_12.6.r12.6/compiler.35059454_0

$ ./holohub build endoscopy_tool_tracking --language cpp --local
2025-06-17 10:53:01 $ cmake -B /home/swqa/hailong/holoscantest/results/3.4.0.0/holohub-internal/build/endoscopy_tool_tracking -S /home/swqa/hailong/holoscantest/results/3.4.0.0/holohub-internal --no-warn-unused-cli -DPython3_EXECUTABLE=/usr/bin/python3 -DPython3_ROOT_DIR=/usr -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=/opt/nvidia/holoscan/lib -DHOLOHUB_DATA_DIR:PATH=/home/swqa/hailong/holoscantest/results/3.4.0.0/holohub-internal/data -DAPP_endoscopy_tool_tracking=ON -Dholoscan_DIR=/opt/nvidia/holoscan/lib/cmake/holoscan -G Ninja
Not searching for unused variables given on the command line.
-- The CXX compiler identification is GNU 11.4.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Unable to find cuda_runtime.h in "/usr/local/include" for CUDAToolkit_INCLUDE_DIRECTORIES.
-- Unable to find cudart library.
CMake Error at /usr/share/cmake-4.0/Modules/FindPackageHandleStandardArgs.cmake:227 (message):
  Could NOT find CUDAToolkit (missing: CUDAToolkit_INCLUDE_DIRECTORIES
  CUDA_CUDART) (found version "12.6.85")
Call Stack (most recent call first):
  /usr/share/cmake-4.0/Modules/FindPackageHandleStandardArgs.cmake:591 (_FPHSA_FAILURE_MESSAGE)
  /usr/share/cmake-4.0/Modules/FindCUDAToolkit.cmake:1070 (find_package_handle_standard_args)
  /usr/share/cmake-4.0/Modules/CMakeFindDependencyMacro.cmake:78 (find_package)
  /opt/nvidia/holoscan/lib/cmake/holoscan/holoscan-dependencies.cmake:19 (find_dependency)
  /opt/nvidia/holoscan/lib/cmake/holoscan/holoscan-config.cmake:68 (include)
  applications/endoscopy_tool_tracking/cpp/CMakeLists.txt:19 (find_package)


-- Configuring incomplete, errors occurred!
Error running command: cmake -B /home/swqa/hailong/holoscantest/results/3.4.0.0/holohub-internal/build/endoscopy_tool_tracking -S /home/swqa/hailong/holoscantest/results/3.4.0.0/holohub-internal --no-warn-unused-cli -DPython3_EXECUTABLE=/usr/bin/python3 -DPython3_ROOT_DIR=/usr -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=/opt/nvidia/holoscan/lib -DHOLOHUB_DATA_DIR:PATH=/home/swqa/hailong/holoscantest/results/3.4.0.0/holohub-internal/data -DAPP_endoscopy_tool_tracking=ON -Dholoscan_DIR=/opt/nvidia/holoscan/lib/cmake/holoscan -G Ninja
Exit code: 1

$ ls /usr/local/cuda/include/cuda_runtime.h
/usr/local/cuda/include/cuda_runtime.h


```

tbirdso: Just to confirm, does the error above still appear after running `./holohub clear-cache`?

On my devkit I am able to reproduce the original `No CMAKE_CUDA_COMPILER` error when NVCC is not in PATH. We could warn and point users to [CUDA post-installation docs](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#environment-setup) for adding NVCC to PATH if not found; however, many projects may not require NVCC, so this wouldn't necessarily be a valid warning for all projects..

I'm not able to reproduce the error above on my IGX at the moment, will investigate. Could you please share your CMakeCache.txt file here or in an internal thread?

wyli: Sounds good, please followup with ticket 5346748 I only temporarily accessed the testing environment earlier

tbirdso: I think there is an environment issue here:

```
$ echo $PATH
/home/swqa/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin

$ which nvcc
/usr/local/bin/nvcc
```

Seems like a symlink for `/usr/local/cuda-12.6/bin/nvcc` was added at `/usr/local/bin/nvcc`, which impacts the FindCUDAToolkit module's ability to infer `CUDAToolkit_INCLUDE_DIRECTORIES` based on the NVCC path. Instead  [CUDA post-installation docs](https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#environment-setup)  recommends adding `/usr/local/cuda/bin` to `PATH` directly.

utilities/cli/holohub.py
```diff
@@ -503,11 +521,27 @@ def build_project_locally(
...
         benchmark: bool = False,
+        configure_args: Optional[list[str]] = None,
     ) -> tuple[Path, dict]:
         """Helper method to build a project locally"""
         project_data = self.find_project(project_name=project_name, language=language)
         project_type = project_data.get("project_type", "application")
 
+        # Set up environment to ensure CUDA is in PATH
+        env = os.environ.copy()
+        cuda_paths = ["/usr/local/cuda/bin", "/opt/cuda/bin"]
```
tbirdso: 1. Always providing `/usr/local/cuda` risks ignoring edge cases where the dev has multiple CUDA versions installed and wants to explicitly request a different version than default
2. Is `/opt/cuda/bin` a standard find path? Not sure I've encountered this one

Updated the internal ticket for tracking. I suggest that we remove custom CUDA path handling from the CLI and instead document troubleshooting edge cases in the event of building as superuser impacting CMake path search behavior.

utilities/cli/holohub.py
```diff
@@ -531,6 +565,15 @@ def build_project_locally(
         build_dir.mkdir(parents=True, exist_ok=True)
 
         proj_prefix = holohub_cli_util.determine_project_prefix(project_type)
+
+        # Build CMAKE_PREFIX_PATH with both Holoscan SDK and CUDA paths
+        cmake_prefix_paths = [HoloHubCLI.DEFAULT_SDK_DIR]
+        cuda_paths = ["/usr/local/cuda", "/opt/cuda"]
```
tbirdso: Typically not needed if `/usr/local/cuda/bin` is already added to `PATH` as expected.

Maybe we should check the PATH after `./holohub setup` and suggest (print) setup steps if the CUDA path is not set up?

utilities/cli/holohub.py
tbirdso: I'm wary of having the build environment deviate significantly from the host environment, can make debugging difficult. Preference is to recommend containers as the favored way to provide custom environment setup.

tbirdso (changes requested):

utilities/cli/holohub.py
wyli: I think we can remove all the 'guessings' of path, and add a `./holohub print_info` to print runtime environment info? what do you think?

tbirdso: Agreed, `./holohub print_info` or some similar method to inspect the environment to be used at app runtime would be useful. Previously `./dev_container build_and_run <project> --dryrun` was one way to inspect PYTHONPATH and working directory updates for a given project run.

wyli: ok, removed the path settings, we can add util command to print system info later.

PR #930, opened 2025-06-18
CLI implement env-info

wyli:
add debugging info printing:
```
./holohub env-info
./holohub run-container -- ./holohub env-info
```

### DIFF
--- a/utilities/cli/container.py
+++ b/utilities/cli/container.py
@@ -38,7 +38,7 @@ from .util import (
     run_command,
 )
 
-base_sdk_version = "3.3.0"
+base_sdk_version = "3.4.0"
 
 
 class HoloHubContainer:
--- a/utilities/cli/container.py
+++ b/utilities/cli/container.py
@@ -38,7 +38,7 @@ from .util import (
     run_command,
 )
 
-base_sdk_version = "3.4.0"
+base_sdk_version = "3.3.0"
 
 
 class HoloHubContainer:
--- a/utilities/cli/util.py
+++ b/utilities/cli/util.py
@@ -701,7 +701,7 @@ def collect_git_info(holohub_root: Path) -> None:
         print(f"  Modified: {git_status.splitlines()}")
     finally:
         try:
-            os.chdir(original_cwd)  # try torestore the original working directory
+            os.chdir(original_cwd)  # try to restore the original working directory
         except Exception:
             pass
 
--- a/utilities/cli/util.py
+++ b/utilities/cli/util.py
@@ -680,8 +680,8 @@ def collect_holohub_info(
 def collect_git_info(holohub_root: Path) -> None:
     """Collect and display Git repository information"""
     print(f"\n{Color.blue('Git Repository Information:')}")
-    if not holohub_root.exists() and holohub_root.is_dir():
-        print(f"  HoloHub root directory does not exist: {holohub_root}")
+    if not holohub_root.exists() or not holohub_root.is_dir():
+        print(f"  HoloHub root directory does not exist or is not a directory: {holohub_root}")
         return
     original_cwd = os.getcwd()
     try:
--- a/utilities/cli/util.py
+++ b/utilities/cli/util.py
@@ -16,6 +16,7 @@
 
 import grp
 import os
+import platform
 import re
 import shutil
 import subprocess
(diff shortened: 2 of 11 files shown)


utilities/cli/util.py
```diff
@@ -641,3 +650,142 @@ def get_devcontainer_config(
...
+    original_cwd = os.getcwd()
+    try:
+        os.chdir(holohub_root)
+    except Exception as e:
+        print(f"  Cannot access HoloHub directory: {e}")
+        return
+    try:
+        git_branch = run_info_command(["git", "branch", "--show-current"])
+        git_commit_full = run_info_command(["git", "rev-parse", "HEAD"])
+        git_status = run_info_command(["git", "status", "--porcelain"])
```
tbirdso, line 695: Did not know about `--porcelain`, neat

utilities/cli/util.py
```diff
@@ -641,3 +650,142 @@ def get_devcontainer_config(
...
+        print(f"  Branch: {git_branch} Commit: {git_commit}")
+        print(f"  Modified: {git_status.splitlines()}")
+    finally:
+        try:
+            os.chdir(original_cwd)  # try to restore the original working directory
+        except Exception:
+            pass
+
+
+def collect_docker_info() -> None:
```
tbirdso, line 709: Would it make sense to also report the `nvidia-container-toolkit` version here? `apt list --installed 2>/dev/null | grep nvidia-container-toolkit`

utilities/cli/util.py
```diff
@@ -641,3 +650,142 @@ def get_devcontainer_config(
...
+        "PYTHONPATH",
+        "PATH",
+        "LD_LIBRARY_PATH",
+        "CMAKE_BUILD_TYPE",
+        "DOCKER_BUILDKIT",
+        "XDG_SESSION_TYPE",
+        "XDG_RUNTIME_DIR",
+    ]
+    for var in sorted(other_env_vars):
+        print(f"  {var}: {os.environ.get(var) or '(not set)'}")
```
tbirdso, line 784: `PYTHONPATH` here is accurate for the general environment. We also update `PYTHONPATH` at run command time in case the app build has generated a `holohub` module. From `./holohub run endoscopy_tool_tracking --language python --dryrun`:
```py
2025-06-18 13:09:00 [dryrun] $ export PYTHONPATH=/workspace/holohub/benchmarks/holoscan_flow_benchmarking:/opt/nvidia/holoscan/python/lib:/workspace/holohub:/opt/nvidia/holoscan/lib/../python/lib:/workspace/holohub/build/endoscopy_tool_tracking/python/lib:/workspace/holohub
```

What do you think about adding an optional project arg to print the anticipated environment when the project is actually run, i.e. `./holohub env-info endoscopy_tool_tracking`? Does not need to block this PR.

utilities/cli/util.py
```diff
@@ -641,3 +650,142 @@ def get_devcontainer_config(
...
+def collect_environment_variables() -> None:
+    """Collect and display environment variables"""
+    print(f"\n{Color.blue('HoloHub Environment Variables:')}")
+    holohub_env_vars = [
+        "HOLOHUB_CMD_NAME",
+        "HOLOHUB_BUILD_LOCAL",
+        "HOLOHUB_BASE_IMAGE",
+        "HOLOHUB_APP_NAME",
+        "HOLOHUB_CONTAINER_BASE_NAME",
+        "HOLOHUB_IMAGE",
```
tbirdso: Is this used?

utilities/cli/util.py
wyli: yes, that's a good idea I think it could be done by another PR. we may consider providing info about the specific location of cmakefiles folder and docker inspect instructions as well.

utilities/cli/container.py
```diff
@@ -38,7 +38,7 @@
     run_command,
 )
 
-base_sdk_version = "3.3.0"
+base_sdk_version = "3.4.0"
```
tbirdso: Need fix in https://github.com/nvidia-holoscan/holohub/commit/48ac6e1c639d3cdde39b6cf31fbd801a44580e4a applied:

```suggestion
base_sdk_version = "3.3.0"
```

PR #936, opened 2025-06-18
CLI python wrapping instead of bash

wyli:
python wrapper so that we don't have a layer of bash for potentially more portability

### DIFF
--- a/holohub
+++ b/holohub
@@ -18,9 +18,7 @@
 import os
 import sys
 
-# Add the holohub root directory to Python path
 sys.path.append(os.path.dirname(os.path.abspath(__file__)))
-# Set the command name used by the current file name
 os.environ["HOLOHUB_CMD_NAME"] = f"./{os.path.basename(__file__)}"
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


holohub
```diff
@@ -15,5 +15,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-PYTHONPATH=${PYTHONPATH}:${SCRIPT_DIR} python3 ${SCRIPT_DIR}/utilities/cli/holohub.py "$@"
+import os
+import sys
+
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
```
agirault, line 21: could also just append `./utilities/cli` here and import just `holohub` below? just seems risky that this could find any python modules inside holohub nost just this one, but that is consistent with what was done with bash

wyli:
don't see a need to continue developing this, closing.

for future ref, some docker run entrypoint is currently coupled with bash, they should be modified if revisiting this PR. https://github.com/nvidia-holoscan/holohub/blob/127ef5281b40201779a0cc7492b294a4eda5abba/utilities/cli/util.py#L582-L586

PR #937, opened 2025-06-19
CLI enable custom script entrypoint

wyli:
This mainly considers `./holohub [build|run|install]` within a container:
1. if with `--docker-opts="--entrypoint=/custom/script"`
```
./holohub build endoscopy_tool_tracking --docker-opts="--entrypoint=/opt/nvidia/nvidia_entrypoint.sh"
```
the within container command will be `/opt/nvidia/nvidia_entrypoint.sh ./holohub build endoscopy_tool_tracking --local `

2. depending on the entrypoint, it automatically adds `-c` to the command string, for example when the entrypoint is `/bin/bash`.

### DIFF
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, entrypoint]
 
 permissions:
   contents: read
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2, entrypoint]
+    branches: [main, holoscan-sdk-lws2]
 
 permissions:
   contents: read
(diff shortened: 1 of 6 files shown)


utilities/cli/util.py
```diff
@@ -542,23 +544,76 @@ def docker_args_to_devcontainer_format(docker_args: List[str]) -> List[str]:
     return result
 
 
+def get_shell_command_args(
```
tbirdso: Nitpick: "shell" could refer to either host or container, suggest rename here and elsewhere

```suggestion
def get_entrypoint_command_args(
```

tbirdso (approved):
Looks good, where is a good place to document the new default behavior? Maybe the [migration guide](https://github.com/nvidia-holoscan/holohub/pull/938) would be appropriate to compare `dev_container` behavior versus new default behavior in HoloHub CLI?

.github/workflows/check_docker.yml
```diff
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, entrypoint]
```
```suggestion
    branches: [main, holoscan-sdk-lws2]
```

PR #939, opened 2025-06-19
Improve docker for colonoscopy application

jjomier:
Add separate dockerfile to improve time to build and run.
Fix CMake files to allow building python applications without a compiler

### DIFF
--- a/applications/colonoscopy_segmentation/Dockerfile
+++ b/applications/colonoscopy_segmentation/Dockerfile
@@ -1,0 +1,54 @@
+# syntax=docker/dockerfile:1
+
+# SPDX-FileCopyrightText: Copyright (c) 2022-2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+############################################################
+# Base image
+############################################################
+
+
+FROM nvcr.io/nvidia/cuda:12.9.0-base-ubuntu24.04
+
+# Install the dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        curl \
+        unzip \
+        cmake \
+        make \
+        patch \
+        ffmpeg \
+        python3 \
+        python3-venv \
+        libnpp-12-9 \
+        cuda-nvrtc-12-9 \
+        libnvinfer10 \
+        libnvinfer-plugin10 \
+        libnvonnxparsers10 \
+        && rm -rf /var/lib/apt/lists/*
+
+RUN if [ $(uname -m) = "aarch64" ]; then ARCH=arm64; else ARCH=linux; fi \
+    && curl -S -# -L -o ngccli_linux.zip https://ngc.nvidia.com/downloads/ngccli_${ARCH}.zip \
+    && unzip -q ngccli_linux.zip \
+    && rm ngccli_linux.zip \
+    && chmod u+x ngc-cli/ngc \
+    && ln -s $(pwd)/ngc-cli/ngc /usr/local/bin/
+
+RUN python3 -m venv /local-venv
+RUN /local-venv/bin/pip3 install holoscan
+RUN /local-venv/bin/pip3 install cupy-cuda12x
+
+ENV PATH="/local-venv/bin:$PATH"
--- a/applications/colonoscopy_segmentation/Dockerfile
+++ b/applications/colonoscopy_segmentation/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-# SPDX-FileCopyrightText: Copyright (c) 2022-2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,12 +14,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-
-############################################################
-# Base image
-############################################################
-
 
 FROM nvcr.io/nvidia/cuda:12.9.0-base-ubuntu24.04
 
--- a/applications/colonoscopy_segmentation/Dockerfile
+++ b/applications/colonoscopy_segmentation/Dockerfile
@@ -18,15 +18,16 @@
 FROM nvcr.io/nvidia/cuda:12.9.0-base-ubuntu24.04
 
 # Install the dependencies
+# wget, unzip and ffmpeg are needed to download NGC-CLI and generate input data
+# python packages are required for the application
+# libnpp, cuda-nvrtc and libnvinfer are required for Holoscan inference
 RUN apt-get update && apt-get install -y --no-install-recommends \
-        curl \
+        wget \
         unzip \
         cmake \
         make \
         patch \
         ffmpeg \
-        python3 \
-        python3-venv \
         libnpp-12-9 \
         cuda-nvrtc-12-9 \
         libnvinfer10 \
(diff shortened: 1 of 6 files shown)


applications/colonoscopy_segmentation/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+############################################################
+# Base image
+############################################################
+
+
+FROM nvcr.io/nvidia/cuda:12.9.0-base-ubuntu24.04
+
+# Install the dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
```
agirault, line 24: nitpick: comment why each dep is needed

applications/colonoscopy_segmentation/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+
+RUN if [ $(uname -m) = "aarch64" ]; then ARCH=arm64; else ARCH=linux; fi \
+    && curl -S -# -L -o ngccli_linux.zip https://ngc.nvidia.com/downloads/ngccli_${ARCH}.zip \
+    && unzip -q ngccli_linux.zip \
+    && rm ngccli_linux.zip \
+    && chmod u+x ngc-cli/ngc \
+    && ln -s $(pwd)/ngc-cli/ngc /usr/local/bin/
+
+RUN python3 -m venv /local-venv
+RUN /local-venv/bin/pip3 install holoscan
```
agirault: nit: pin version?

applications/colonoscopy_segmentation/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+        && rm -rf /var/lib/apt/lists/*
+
+RUN if [ $(uname -m) = "aarch64" ]; then ARCH=arm64; else ARCH=linux; fi \
+    && curl -S -# -L -o ngccli_linux.zip https://ngc.nvidia.com/downloads/ngccli_${ARCH}.zip \
+    && unzip -q ngccli_linux.zip \
+    && rm ngccli_linux.zip \
+    && chmod u+x ngc-cli/ngc \
+    && ln -s $(pwd)/ngc-cli/ngc /usr/local/bin/
+
+RUN python3 -m venv /local-venv
```
agirault: why create a venv if you're already in a container? Seems sensible to install on system due to the isolation you already get.

applications/colonoscopy_segmentation/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+############################################################
+# Base image
+############################################################
+
+
+FROM nvcr.io/nvidia/cuda:12.9.0-base-ubuntu24.04
```
agirault: nitpick: no need for that comment on top if you don't have over layers. also lots of empty lines

```suggestion

FROM nvcr.io/nvidia/cuda:12.9.0-base-ubuntu24.04
```

applications/colonoscopy_segmentation/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+        make \
+        patch \
+        ffmpeg \
+        python3 \
+        python3-venv \
+        libnpp-12-9 \
+        cuda-nvrtc-12-9 \
+        libnvinfer10 \
+        libnvinfer-plugin10 \
+        libnvonnxparsers10 \
```
agirault, line 35: no need for cudart? transitively handled?

applications/colonoscopy_segmentation/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    && unzip -q ngccli_linux.zip \
+    && rm ngccli_linux.zip \
+    && chmod u+x ngc-cli/ngc \
+    && ln -s $(pwd)/ngc-cli/ngc /usr/local/bin/
+
+RUN python3 -m venv /local-venv
+RUN /local-venv/bin/pip3 install holoscan
+RUN /local-venv/bin/pip3 install cupy-cuda12x
+
+ENV PATH="/local-venv/bin:$PATH"
```
agirault:
```suggestion
ENV PATH="/local-venv/bin:$PATH"

```

agirault (approved):
just nitpicks

applications/colonoscopy_segmentation/Dockerfile
jjomier: I think we want to use the latest available version of Holoscan.

applications/colonoscopy_segmentation/Dockerfile
jjomier: It does not work without venv on Ubuntu 24.04. It requires venv to install.
I'll add more to that. Apparently this is a new way to ensure that new packages don't break python system installation.

applications/colonoscopy_segmentation/Dockerfile
jjomier: Seems that cudart is included in the base image `nvcr.io/nvidia/cuda:12.9.0-base-ubuntu24.04`

applications/colonoscopy_segmentation/Dockerfile
tbirdso: Recommend aligning to the default Holoscan SDK / HoloHub container approach with allowing breaking packages by default in Holoscan SDK >=3.3.0 containers: https://github.com/nvidia-holoscan/holohub/blob/main/Dockerfile#L37

So far we haven't encountered issues where package updates have actually broken the environment. If you do encounter issues in practice then pursuing venv may make sense in that case.

applications/colonoscopy_segmentation/Dockerfile
```diff
@@ -0,0 +1,54 @@
+# syntax=docker/dockerfile:1
+
+# SPDX-FileCopyrightText: Copyright (c) 2022-2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tbirdso:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

applications/colonoscopy_segmentation/Dockerfile
tbirdso, line 18: What speedup are you seeing in time to build and run with smaller base image + package install versus larger base image with most packages pre-installed?

I wonder how speedup compares across different network speeds? Might be good to define a target network speed for time-to-run KPI in the future.

applications/colonoscopy_segmentation/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+        python3-venv \
+        libnpp-12-9 \
+        cuda-nvrtc-12-9 \
+        libnvinfer10 \
+        libnvinfer-plugin10 \
+        libnvonnxparsers10 \
+        && rm -rf /var/lib/apt/lists/*
+
+RUN if [ $(uname -m) = "aarch64" ]; then ARCH=arm64; else ARCH=linux; fi \
+    && curl -S -# -L -o ngccli_linux.zip https://ngc.nvidia.com/downloads/ngccli_${ARCH}.zip \
```
tbirdso: Should we pin the NGC CLI version? We've seen functional breakage in past updates, even in patch versions. https://github.com/nvidia-holoscan/holohub/commit/58162e5a7a82f25b6b527021d9c242784a0552fe

applications/colonoscopy_segmentation/colonoscopy_segmentation.py
```diff
@@ -28,8 +28,30 @@
 )
 from holoscan.resources import BlockMemoryPool, CudaStreamPool, MemoryStorageType
 
-from holohub.aja_source import AJASourceOp
-
+def lazy_import(module_name):
```
tbirdso: Should we move this to a common HoloHub utility at some point, given that it seems we are using this across Python apps with optional AJA components? Not a blocker for this PR

bhashemian: I agree, but there are a few considerations to keep in mind. A few months ago, I added it to the utilities as `import_utils.py`, but we encountered an issue when running tests because it wasn’t being installed or added to the Python path.

One argument against making it a HoloHub utility is that the application directory will depend on the HoloHub repository to function. This wouldn’t be a significant issue for applications that already rely on HoloHub operators, but for applications that solely depend on the Holoscan, it would be an undesirable limitation.

operators/deidentification/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 cmake_minimum_required(VERSION 3.18)
-project(deidentification)
+project(deidentification NONE)
```
bhashemian, line 17: good catch!

applications/colonoscopy_segmentation/colonoscopy_segmentation.py
```diff
@@ -120,7 +142,8 @@ def compose(self):
 
         is_aja = self.source.lower() == "aja"
         if is_aja:
-            source = AJASourceOp(self, name="aja", **self.kwargs("aja"))
+            aja_source = lazy_import("holohub.aja_source")
```
bhashemian: Since it is a lazy import, I think placing it at the top the file makes more sense. Otherwise, we could use `import` here instead, right?

applications/colonoscopy_segmentation/Dockerfile
agirault: and it's apples and oranges as well, because what is installed here is the minimal set (ex: no torch or ort)

applications/colonoscopy_segmentation/Dockerfile
agirault: and since URLs change:
```dockerfile
ENV PIP_BREAK_SYSTEM_PACKAGES=1
```

applications/colonoscopy_segmentation/colonoscopy_segmentation.py
jjomier: It seems more readable to put it inline, is that an issue?

bhashemian: Not an issue but it might be redundant. Functionally, you don’t need `lazy_import` if you’re importing and calling the module in the same code block because if it reaches the importing line, it will also reach the calling line.

Lazy import is mostly beneficial in situations where you want to import it at the top of the code, regardless of where the module is actually called, or you want to use the information about its existence to make a decision based on it.

Here, it seems that you could replace `lazy_import` with `import`, and the code would function in similar way.

applications/colonoscopy_segmentation/Dockerfile
jjomier: Tried to align with your suggestions but getting the following when running the app:

`Error: PermissionError: [Errno 13] Permission denied
/usr/local/lib/python3.12/dist-packages/holoscan-3.4.0.dist-info/axle.lck
Location: System-wide Python packages directory (/usr/local/lib/python3.12/dist-packages/)`

applications/colonoscopy_segmentation/colonoscopy_segmentation.py
jjomier: Oh I see - let me check

jjomier: I removed the lazy import

applications/colonoscopy_segmentation/Dockerfile
jjomier: I did run some benchmarks:
old Docker: ~5 minutes
new Docker: ~2 minutes

PR #941, opened 2025-06-24
Update Advanced Network Manager Python bindings

ronyrad:
- Introduced Python network initialization function `adv_net_init` to handle Holoscan config objects.
- Updated various functions for burst creation, packet management, and network interface operations.

## Summary by CodeRabbit

* **New Features**
  * Expanded Python bindings with new networking enums, types, and APIs.
  * Added comprehensive packet-burst management for transmit/receive, including header, payload, and memory helpers.
  * Introduced a unified network initialization API accepting multiple config formats.
  * New utilities for manager/port identification, stats, shutdown, and enhanced status codes for clearer diagnostics.


operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
         "Create a shared pointer burst params structure");
+
+  // Packet information functions
   m.def("get_segment_packet_length",
         py::overload_cast<BurstParams*, int, int>(&get_segment_packet_length),
-        "Get length of one segments of the packet");
+        "Get length of one segment of the packet",
+        "burst"_a,
+        "seg"_a,
+        "idx"_a);
```
grlee77: nitpick: I think it should work either way, but in Holoscan SDK at least, we tended to place the arguments before the docstring as in the suggestion below. (The same applies to the other methods in this file as well)
```suggestion
        "burst"_a,
        "seg"_a,
        "idx"_a,
        "Get length of one segment of the packet");
```

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+  py::class_<BurstHeader>(m, "BurstHeader")
+      .def(py::init<>())
+      .def_readwrite("hdr", &BurstHeader::hdr);
+
+  py::class_<BurstParams>(m, "BurstParams")
+      .def(py::init<>())
+      .def_readwrite("hdr", &BurstParams::hdr);
+
+  // Network initialization and configuration
+  // Note: The original adv_net_init(NetworkConfig&) is not bound since NetworkConfig
```
cliffburdick: I'm not sure what this comment means. Isn't NetworkConfig available in Python?

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+          std::string yaml_str;
+
+          // Check if it has a 'value' attribute (Holoscan Arg objects)
+          if (py::hasattr(config_obj, "value")) {
+            auto value = config_obj.attr("value");
+            yaml_str = py::str(value);
+          } else if (py::hasattr(config_obj, "as_dict")) {
+            auto dict_obj = config_obj.attr("as_dict")();
+            // Convert dict to YAML string
+            yaml_str = py::str(dict_obj);
```
grlee77: I don't think the string returned here would be a valid YAML string.

I think you would need something like the following to use Python's `yaml` module to dump the dict object as a valid YAML string
```suggestion
            py::object yaml_module = py::module_::import("yaml");
            py::object py_yaml_str = yaml_module.attr("dump")(dict_obj);
            yaml_str = py::cast<std::string>(py_yaml_str);
```

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+      .def_readwrite("port_id", &BurstHeaderParams::port_id)
+      .def_readwrite("q_id", &BurstHeaderParams::q_id)
+      .def_readwrite("num_segs", &BurstHeaderParams::num_segs)
+      .def_readwrite("nbytes", &BurstHeaderParams::nbytes)
+      .def_readwrite("first_pkt_addr", &BurstHeaderParams::first_pkt_addr)
+      .def_readwrite("max_pkt", &BurstHeaderParams::max_pkt)
+      .def_readwrite("max_pkt_size", &BurstHeaderParams::max_pkt_size)
+      .def_readwrite("gpu_pkt0_idx", &BurstHeaderParams::gpu_pkt0_idx)
+      .def_readwrite("gpu_pkt0_addr", &BurstHeaderParams::gpu_pkt0_addr)
+      .def_readwrite("burst_flags", &BurstHeaderParams::burst_flags);
```
grlee77, line 81: `burst_flags` doesn't exist for the `BurstHeaderParams` struct?

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+
+  // RX burst functions - all overloads
+  m.def(
+      "get_rx_burst",
+      [](int port, int q) {
+        BurstParams* burst_ptr = nullptr;
+        Status status = get_rx_burst(&burst_ptr, port, q);
+        return py::make_tuple(status, py::cast(burst_ptr, py::return_value_policy::reference));
+      },
+      "Get RX burst for specific port and queue",
```
grlee77: I think this is okay, but the user is responsible for calling `free_rx_burst` from Python to free the object when done with it? Since Python authors are likely not used to explicitly freeing resources, we may need to mention that in the docstring here (and in the other overloads for this function just below)

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+        py::overload_cast<BurstParams*, int, int>(&get_segment_packet_ptr),
+        "Get packet pointer for one segment",
+        "burst"_a,
+        "seg"_a,
+        "idx"_a);
+  m.def("get_packet_ptr",
+        py::overload_cast<BurstParams*, int>(&get_packet_ptr),
+        "Get packet pointer",
+        "burst"_a,
+        "idx"_a);
```
grlee77: Not sure these methods returning void* are useful from Python?

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+
+  // TX burst functions
   m.def("tx_burst_available",
```
grlee77: This is consistent with the previously existing bindings, but not sure if it should be named `is_tx_burst_available`?

cliffburdick (approved):

cliffburdick (approved):

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+      },
+      "Get RX burst from any queue on any port");
+
+  // Header setting functions
+  m.def("set_eth_header",
+        py::overload_cast<BurstParams*, int, char*>(&set_eth_header),
+        "Set Ethernet header in packet",
+        "burst"_a,
+        "idx"_a,
+        "dst_addr"_a);
```
grlee77: The `char*` here should probably be `const char*` in the underlying C++ code? (i.e. this function uses the char* as input but does not modify it?)

It may be better to use some lambda to pass a Python `str` containing the MAC adddress like this:

```suggestion
m.def("set_eth_header",
      [](BurstParams* burst, int idx, const std::string& dst_addr) -> Status {
        char mac_bytes[6];
        format_eth_addr(mac_bytes, dst_addr);  // Use existing utility function
        return set_eth_header(burst, idx, mac_bytes);
      },
      "Set Ethernet header in packet",
      "burst"_a,
      "idx"_a,
      "dst_addr"_a);
```

or if it expected that bytes would be used instead?
```python
m.def("set_eth_header",
      [](BurstParams* burst, int idx, py::bytes dst_addr) -> Status {
        std::string addr_str = py::cast<std::string>(dst_addr);
        if (addr_str.size() != 6) {
          throw std::invalid_argument("MAC address must be 6 bytes");
        }
        return set_eth_header(burst, idx, const_cast<char*>(addr_str.c_str()));
      },
      "Set Ethernet header in packet",
      "burst"_a,
      "idx"_a,
      "dst_addr"_a);
```

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+        "udp_len"_a,
+        "src_port"_a,
+        "dst_port"_a);
+  m.def("set_udp_payload",
+        py::overload_cast<BurstParams*, int, void*, int>(&set_udp_payload),
+        "Set UDP payload in packet",
+        "burst"_a,
+        "idx"_a,
+        "data"_a,
+        "len"_a);
```
grlee77: for void* pointer and size as input, the Python binding probably needs a lambda accepting a Python buffer object like this:

```suggestion
  m.def("set_udp_payload",
      [](BurstParams* burst, int idx, py::buffer data) -> Status {
        py::buffer_info buf_info = data.request();
        return set_udp_payload(burst, idx, buf_info.ptr, buf_info.size);
      },
      "Set UDP payload in packet",
      "burst"_a,
      "idx"_a,
      "data"_a);
```

then Python code could pass a byte string, `bytes` object or other object implementing the buffer protocol (e.g. a NumPy array) as the data.

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,316 @@
...
+      "free_tx_burst", py::overload_cast<BurstParams*>(&free_tx_burst), "Free TX burst", "burst"_a);
+  m.def(
+      "free_rx_burst", py::overload_cast<BurstParams*>(&free_rx_burst), "Free RX burst", "burst"_a);
+
+  // Network interface functions
+  m.def("get_mac_addr",
+        py::overload_cast<int, char*>(&get_mac_addr),
+        "Get MAC address of an interface",
+        "port"_a,
+        "mac"_a);
```
grlee77: Writes to the provided char* will not work from Python. Consider returning a tuple where the second element is the result formatted as a Python string (empty string if the status was not successful).

```suggestion
// Return formatted string (most common use case)
m.def("get_mac_addr",
      [](int port) -> py::tuple {
        char mac[6];
        Status status = get_mac_addr(port, mac);
        if (status == Status::SUCCESS) {
          char formatted[18];
          snprintf(formatted, sizeof(formatted), 
                   "%02x:%02x:%02x:%02x:%02x:%02x",
                   static_cast<unsigned char>(mac[0]),
                   static_cast<unsigned char>(mac[1]),
                   static_cast<unsigned char>(mac[2]),
                   static_cast<unsigned char>(mac[3]),
                   static_cast<unsigned char>(mac[4]),
                   static_cast<unsigned char>(mac[5]));
          return py::make_tuple(status, std::string(formatted));
        }
        return py::make_tuple(status, std::string(""));
      },
      "Get MAC address of an interface as formatted string",
      "port"_a);
```

ronyrad: Updated, according to the suggestion

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Updated, according to the suggestion

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: I removed the comment. But the idea is that python doesn't generate automatically NetworkConfig from Yaml. We also do not want duplicated implementation. Python function receives the Yaml Node and underlying CPP will do the work

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Changes made as suggested.

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Changes made as suggested.

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: burst_flags were added in a parallel PR. I plan that this PR will be merged after it

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Added docstrings

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Removed

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Function name was updated

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Changes made as suggested.

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Changes made as suggested.

operators/advanced_network/python/adv_network_common_pybind.cpp
ronyrad: Changes made as suggested.

ronyrad:
After reviewing the code, I don't think that python bindings will be used at packet or burst level because of the low performance. We should consider to stay with library management functions only

operators/advanced_network/python/adv_network_common_pybind.cpp
```diff
@@ -17,109 +17,335 @@
...
 
+  m.def(
+      "get_rx_burst",
+      []() {
+        BurstParams* burst_ptr = nullptr;
+        Status status = get_rx_burst(&burst_ptr);
+        return py::make_tuple(status, py::cast(burst_ptr, py::return_value_policy::reference));
+      },
+      "Get RX burst from any queue on any port. "
+      "Must call free_rx_burst() when done to free resources.");
```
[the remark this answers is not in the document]
grlee77: seems like a valid comment, but could be left to a future enhancement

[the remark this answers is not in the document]
bhashemian: @coderabbitai please do!

PR #942, opened 2025-06-24
Infra: Address `python3` version mismatch in HoloHub default Dockerfile

tbirdso:
Followup to https://github.com/nvidia-holoscan/holohub/commit/5db60cfc20d737c0dbd3863d981624223db5b8ab to address an issue where a custom `python3.X` installation in the default Dockerfile is overridden in later layers.

## Previous behavior

[Dockerfile](https://github.com/nvidia-holoscan/holohub/blob/main/Dockerfile#L36) currently installs Python 3.12 by default across base images, however the output of `./holohub build-container --base-img ubuntu:22.04` shows different Python versions installed simultaneously:

```
$ python3 --version
Python 3.10.12
$ python --version
Python 3.12.11
```

__Impact__: This behavior is unexpected and could lead to confusion for developers working in a container based on Ubuntu 22.04, such as any projects that pin their base container to Holoscan SDK <= v3.2.0.

On inspection there are two later layers that override the default Python 3.12 installation in Ubuntu 22.04 environments:
- `./holohub setup` checks for and uses `python3-dev` explicitly, which installs Python 3.10
- `gobject-inspection` and `libgirepository1.0-dev` benchmarking dependencies depend explicitly on Python 3.10 in Ubuntu 22.04

## Proposed changes

- Revert to use the default `python3` version string for installation, which uses system defaults (3.10 on Ubuntu 22.04, 3.12 on Ubuntu 24.04)
- Install minimal `python3` by default and let `./holohub setup` be responsible for `python3.X-dev` package installation
- Update `./holohub setup` to check for the minor version of the interpreter running the CLI when inspecting `python3-dev` or `python3.X-dev`, allowing `./holohub setup` to complete without installing additional Python versions

## Test coverage
- `./holohub run-container --base-img ubuntu:22.04`
- `./holohub run-container --base-img ubuntu:24.04`

## Additional notes

This PR does not attempt to address Python version pinning in benchmarking dependencies. We've had separate conversations in the past about making benchmarking dependencies an optional build step, which would support the custom version request here.

Requesting a custom version with `PYTHON3_VERSION=python3.11` or similar will currently fail at the benchmarking requirements installation build layer.

### DIFF
--- a/utilities/cli/util.py
+++ b/utilities/cli/util.py
@@ -174,6 +174,11 @@ def run_info_command(cmd: List[str]) -> Optional[str]:
         return None
 
 
+def parse_semantic_version(version: str) -> Tuple[int, int, int]:
+    """Parse semantic version string MAJOR.MINOR. into tuple of integers for comparison"""
+    return tuple(map(int, version.split(".")))
+
+
 def check_nvidia_ctk(min_version: str = "1.12.0", recommended_version: str = "1.14.1") -> None:
     """Check NVIDIA Container Toolkit version"""
 
@@ -187,20 +192,12 @@ def check_nvidia_ctk(min_version: str = "1.12.0", recommended_version: str = "1.
         match = re.search(r"(\d+\.\d+\.\d+)", output)
         if match:
             version = match.group(1)
-
             try:
-                from packaging import version as ver
-
-                version_check = ver.parse(version) < ver.parse(min_version)
-            except ImportError:
-
-                def parse_version(v):
-                    try:
-                        return tuple(map(int, v.split(".")))
-                    except ValueError:
-                        return (10, 0, 0)
-
-                version_check = parse_version(version) < parse_version(min_version)
+                version_check = parse_semantic_version(version) < parse_semantic_version(
+                    min_version
+                )
+            except ValueError:
+                version_check = False
 
             if version_check:
                 fatal(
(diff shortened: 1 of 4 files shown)


utilities/cli/util.py
```diff
@@ -187,20 +192,12 @@ def check_nvidia_ctk(min_version: str = "1.12.0", recommended_version: str = "1.
...
-                from packaging import version as ver
-
-                version_check = ver.parse(version) < ver.parse(min_version)
-            except ImportError:
-
-                def parse_version(v):
-                    try:
-                        return tuple(map(int, v.split(".")))
-                    except ValueError:
-                        return (10, 0, 0)
```
wyli, line 201: consider keeping this try... except ValueError, for use cases such as `list(map(int, '1.2.dev0'.split('.')))` it'll be a value error

wyli (approved):
looks good to me, there's one minor comment and I think `./holohub setup` is doing too many things, perhaps we should further define some options like `./holohub setup benchmark`

utilities/cli/util.py
tbirdso: Good suggestion, updated as follows for the general case:
- Added stricter regex check to get MAJOR.MINOR.PATCH... from [semantic version](https://semver.org/) and strip trailing postfix without error, i.e. `1.2.3+dev0` is a valid case
- Added test coverage to validate behavior

In my opinion we can consider `1.2.dev0` as a valid ValueError because it does not have a valid PATCH version as defined by semantic versioning. We know that we can get a semantic version for the two components where this check is currently applied (`python3-dev` and `nvidia-container-toolkit`), so this seems like a safe assumption for now.

PR #944, opened 2025-06-25
add Holochat test

wyli:


applications/holochat/tests/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
...
+  PROPERTIES
+  PASS_REGULAR_EXPRESSION "llm - INFO - Retrieving Holoscan embeddings vector DB.*Ran 4 tests.*OK"
+  FAIL_REGULAR_EXPRESSION "FAILED(?!.*telemetry event)|ERROR(?!.*telemetry event)|Exception")
+set_tests_properties(holochat_chatbot_test
+  PROPERTIES
+  PASS_REGULAR_EXPRESSION ".*Ran 3 tests.*OK"
+  FAIL_REGULAR_EXPRESSION "FAILED(?!.*telemetry event)|ERROR(?!.*telemetry event)|Exception")
+set_tests_properties(holochat_build_db_test
+  PROPERTIES
+  PASS_REGULAR_EXPRESSION "Ran 2 tests.*OK.*Building Holoscan Embeddings Chroma DB"
```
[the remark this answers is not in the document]
wyli: > The `PASS_REGULAR_EXPRESSION` expects a log segment `Building Holoscan Embeddings Chroma DB`, but the `main()` in `build_holoscan_db.py` does not print this phrase; update the regex or add the missing log.

there is a print command https://github.com/nvidia-holoscan/holohub/blob/fbface37fd0340f2fac51e3f1dbb5d03ff5a821d/applications/holochat/build_holoscan_db.py#L120

wyli:
I'tested locally with `./holohub test holochat --no-xvfb --verbose`

applications/holochat/tests/test_chatbot.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_ask_question_function(self):
+        """Test ask_question function"""
+        from chatbot import ask_question
+
+        message = "Hello"
+        chat_history = None
+        result_message, result_history = ask_question(message, chat_history)
+        self.assertEqual(result_message, "")
+        self.assertEqual(len(result_history), 1)
+        self.assertIn("Welcome to HoloChat", result_history[0][1])
```
NigelNelson, line 50: Where does this "Welcome to HoloChat" come from?

NigelNelson:
Looks very nice to me, I think it does a good job testing what is reasonable: Getting the proper PDF, building the db, testing the chatbot, and the NIM LLM.
It doesn't cover using the local LLM, but I understand that is too large of dependency to bring in at test time. My only question is will the test runner env have a NIM API key?

applications/holochat/tests/test_chatbot.py
wyli: that's from the initial prompt: https://github.com/nvidia-holoscan/holohub/blob/4e9bbaf2979ba50d3565060c054ab2cd7a48720c/applications/holochat/chatbot.py#L57

wyli:
>  will the test runner env have a NIM API key?

The test doesn't actually connect to the remote service, I feel that'll be depending too much on the network/service availability, may give us false positives, so it only checks basic config initializations. I think we can add NIM tests if there's issues reported.

tbirdso (approved):
Thanks @wyli @NigelNelson !

PR #949, opened 2025-06-26
CLI: Fix CUDA package installation in `handle_setup`

tbirdso:
### Background

Environment setup was initially ported from bash `run` script to Python CLI in b57201a. Inspection reveals installation gaps and potential script errors when CUDA Runtime is available in the environment due to bad handling of `dpkg` and `apt` commands.

### Previous Behavior

In dGPU environment without `libnvinfer-bin` pre-installed:
```bash
$ mkdir -p test-cli && cd test-cli && cmake .. && ctest -C Release -R test_holohub_setup
...
subprocess.CalledProcessError: Command '['dpkg', '--status', 'libnvinfer-bin']' returned non-zero exit status 1.
```

### Updated Behavior

1. Fix `cuda-cudart` version search to get CUDA major version for package installation
2. Use regex for cuDNN and TensorRT package search and selection
3. Handle subprocess failures more cleanly
4. Warn and continue without CUDA package installation if CUDA runtime is not available
    
Tested locally with CUDA 12.3 runtime package default. Supports internal dGPU test coverage.

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -31,7 +31,7 @@ from typing import Optional
 import utilities.cli.util as holohub_cli_util
 import utilities.metadata.gather_metadata as metadata_util
 from utilities.cli.container import HoloHubContainer, base_sdk_version
-from utilities.cli.util import Color, parse_semantic_version
+from utilities.cli.util import Color, PackageInstallationError, parse_semantic_version
 
 PYTHON_MIN_VERSION = "3.9.0"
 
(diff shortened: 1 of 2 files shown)


utilities/cli/util.py
```diff
@@ -309,13 +309,21 @@ def list_metadata_json_dir(*paths: Path) -> List[Tuple[str, str]]:
 
 
 def install_cuda_dependencies_package(
-    package_name: str, preferred_version: str, optional: bool = False, dry_run: bool = False
-) -> bool:
+    package_name: str,
+    preferred_version: Optional[str] = None,
```
wyli: is preferred_version=None supported?

utilities/cli/holohub.py
```diff
@@ -1466,6 +1416,72 @@ def handle_setup(self, args: argparse.Namespace) -> None:
...
+        except subprocess.CalledProcessError as e:
+            print(f"Error checking cudnn9 version, skipping: {e}")
+
+        # Install TensorRT dependencies
+        # TensorRT version example: libnvinfer-bin/unknown,now 10.12.0.36-1+cuda12.9 amd64
+        NVINFER_PATTERN = rf"\d+\.[0-9]+\.[0-9]+\.[0-9]+-[0-9]\+cuda{cuda_major_version}\.[0-9]+"
+        installed_libnvinferbin_version = holohub_cli_util.install_cuda_dependencies_package(
+            package_name="libnvinfer-bin",
+            preferred_version=NVINFER_PATTERN,
+            optional=True,
```
wyli: it's marked optional=True, but then the following `if not installed_...` doesn't allow optional

utilities/cli/holohub.py
```diff
@@ -1466,6 +1416,72 @@ def handle_setup(self, args: argparse.Namespace) -> None:
 
         print(Color.green("Setup for HoloHub is ready. Happy Holocoding!"))
 
+    def _setup_cuda_packages(self, cuda_major_version: int, dryrun: bool = False) -> None:
```
wyli: perhaps str type?
```suggestion
    def _setup_cuda_packages(self, cuda_major_version: str, dryrun: bool = False) -> None:
```

utilities/cli/holohub.py
```diff
@@ -1466,6 +1416,72 @@ def handle_setup(self, args: argparse.Namespace) -> None:
...
+                    package_name="libcudnn8",
+                    preferred_version=CUDNN_8_PATTERN,
+                    dry_run=dryrun,
+                )
+                holohub_cli_util.install_cuda_dependencies_package(
+                    package_name="libcudnn8-dev",
+                    preferred_version=re.escape(installed_cudnn8_version),
+                    dry_run=dryrun,
+                )
+        except subprocess.CalledProcessError as e:
```
wyli: I think this is already handled within `install_cuda_dependencies_package` if we set optional=True?

utilities/cli/holohub.py
```diff
@@ -1466,6 +1416,72 @@ def handle_setup(self, args: argparse.Namespace) -> None:
...
+
+        for trt_package_name in [
+            "libnvinfer-headers-dev",
+            "libnvinfer-dev",
+            "libnvinfer-plugin-dev",
+            "libnvonnxparsers-dev",
+        ]:
+            holohub_cli_util.install_cuda_dependencies_package(
+                package_name=trt_package_name,
+                preferred_version=libnvinfer_pattern,
```
wyli: I think this should be set to optional=True to install the packages as much as possible?

wyli (approved):
thanks, this is now more systematic, just a few inline comments to confirm the error handling strategy..

utilities/cli/holohub.py
tbirdso: Left over from refactoring, will remove

utilities/cli/util.py
tbirdso: No, updated to `str` with default version matching pattern

utilities/cli/holohub.py
tbirdso: Thanks for the feedback -- originally wanted to minimize API changes, but on review the initial approach was using too many strategies to signal and handle installation failure (process error,  optional return value, `optional` param)

Updated to the simplified strategy:
1. If the package:
  - Is already installed or is installed successfully: return the version
  - is not present and fails to install: throw an error
2. Use try/catch rather than `optional` field to control logical flow when install fails

Please let me know what you think

utilities/cli/holohub.py
tbirdso: No strong guidance for behavior when any individual package in a set fails installation, so long as behavior is well defined. I've updated to fail out as soon as possible on failure to save time on an incomplete installation, open to feedback.

PR #950, opened 2025-06-26
CLI: Fix CTest coverage regular expressions for pass/fail

tbirdso:
Fix CTest regular expressions to mark test output as passed or failed following initial migration in b57201a6.

Multiple `PASS_REGULAR_EXPRESSION` or `FAIL_REGULAR_EXPRESSION` definitions on a single test will override rather than compounding test conditions. Previous use of multiple expressions risked failing to detect test failures due to only matching against the last expression.

This update consolidates expressions into a single pass/fail regex expression per test to ensure proper "AND/OR" chaining of expressions.

Tested locally in x86 dGPU + CUDA 12.3 environment


wyli (approved):
Thanks!

PR #954, opened 2025-06-27
Add Documentation for Holohub CMake Helper Functions

bhashemian:
This pull request enhances the `cmake/HoloHubConfigHelpers.cmake` file by adding detailed documentation and examples for CMake helper functions used to build HoloHub components.

### DIFF
--- a/cmake/HoloHubConfigHelpers.cmake
+++ b/cmake/HoloHubConfigHelpers.cmake
@@ -52,7 +52,7 @@
 #   PKG_${NAME}: CMake option to enable/disable this package
 #
 # Example:
-#   add_holohub_package(my_package 
+#   add_holohub_package(my_package
 #     EXTENSIONS gxf_core gxf_serialization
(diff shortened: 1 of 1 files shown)


tbirdso (commented):
Thanks! Could we also add a CLI usage example of what configuration flags each function recognizes?

I.e. if `add_holohub_application(myapp)` is present then a user could pass `cmake -S . -B build -DAPP_myapp:BOOL=<ON/OFF>`, etc

bhashemian:
> Thanks! Could we also add a CLI usage example of what configuration flags each function recognizes?
> 
> I.e. if `add_holohub_application(myapp)` is present then a user could pass `cmake -S . -B build -DAPP_myapp:BOOL=<ON/OFF>`, etc

Sure, but I think it is already capturing the same notion in the "Creates" section, isn't it?
```cmake
# Creates:
#   APP_${NAME}: CMake option to enable/disable this application
```

PR #963, opened 2025-07-01
Update app and benchmark usages for holohub CLI

wyli:
(also include minor updates in the migration guide )

### DIFF
--- a/applications/psd_pipeline/metadata.json
+++ b/applications/psd_pipeline/metadata.json
@@ -20,7 +20,6 @@
         "changelog": {
             "1.0": "Initial Release"
         },
-        "dockerfile": "applications/psd_pipeline/Dockerfile",
         "holoscan_sdk": {
             "minimum_required_version": "2.5.0",
             "tested_versions": [
--- a/applications/psd_pipeline/metadata.json
+++ b/applications/psd_pipeline/metadata.json
@@ -20,7 +20,7 @@
         "changelog": {
             "1.0": "Initial Release"
         },
-        dockerfile: "applications/psd_pipeline/Dockerfile",
+        "dockerfile": "applications/psd_pipeline/Dockerfile",
         "holoscan_sdk": {
             "minimum_required_version": "2.5.0",
             "tested_versions": [
--- a/applications/psd_pipeline/metadata.json
+++ b/applications/psd_pipeline/metadata.json
@@ -20,6 +20,7 @@
         "changelog": {
             "1.0": "Initial Release"
         },
+        dockerfile: "applications/psd_pipeline/Dockerfile",
         "holoscan_sdk": {
             "minimum_required_version": "2.5.0",
             "tested_versions": [
--- a/benchmarks/release_benchmarking/metadata.json
+++ b/benchmarks/release_benchmarking/metadata.json
@@ -39,7 +39,7 @@
 		},
 		"run": {
 			"command": "bash <holohub_app_source>/run_benchmarks.sh",
-			"workdir": "<app_source_dir>"
+			"workdir": "holohub_app_source"
 		}
 	}
 }
--- a/benchmarks/release_benchmarking/metadata.json
+++ b/benchmarks/release_benchmarking/metadata.json
@@ -39,7 +39,7 @@
 		},
 		"run": {
 			"command": "bash <holohub_app_source>/run_benchmarks.sh",
-			"workdir": "holohub_app_source"
+			"workdir": "holohub_root"
 		}
 	}
 }
--- a/applications/psd_pipeline/README.md
+++ b/applications/psd_pipeline/README.md
@@ -181,7 +181,7 @@ In this example, if you wanted to use the `ens3f1np1` interface, you'd pass
    ./holohub build-container advanced_network --docker-file ./operators/advanced_network/Dockerfile
 
    # Add the psd-pipeline deps
-   ./holohub build-container psd_pipeline --docker-file ./applications/psd_pipeline/Dockerfile --base-img holohub:ngc-v3.1.0-dgpu --img holohub-psd-pipeline:ngc-v3.1.0-dgpu
+   ./holohub build-container psd_pipeline --base-img holohub:ngc-v3.1.0-dgpu --img holohub-psd-pipeline:ngc-v3.1.0-dgpu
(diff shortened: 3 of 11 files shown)


applications/psd_pipeline/metadata.json
```diff
@@ -20,6 +20,7 @@
         "changelog": {
             "1.0": "Initial Release"
         },
+        "dockerfile": "applications/psd_pipeline/Dockerfile",
```
tbirdso: Seems like the Dockerfile should be picked up automatically as being in the same folder as the app, could we confirm with `./holohub build-container psd_pipeline --dryrun` whether this change is required?

benchmarks/release_benchmarking/metadata.json
```diff
@@ -39,7 +39,7 @@
 		},
 		"run": {
 			"command": "bash <holohub_app_source>/run_benchmarks.sh",
-			"workdir": "<app_source_dir>"
+			"workdir": "holohub_root"
```
tbirdso, line 42: Could you comment on the motivation for the change in working directory?

Is `holohub_root` matched and replaced in the CLI run script without brackets (`<holohub_root>`)?

applications/psd_pipeline/metadata.json
wyli:
```suggestion
```

benchmarks/release_benchmarking/metadata.json
wyli: thanks, there was an error running the script, updating the workdir addressed it.

looking at the ./run implementation, actually anything with `<>` is never replaced with the path https://github.com/nvidia-holoscan/holohub/blob/5566915784fc829070b918e5dd6c5a38b9ed5c21/run#L1014
let me fix those in this pr.

wyli: this fixes the path https://github.com/nvidia-holoscan/holohub/pull/963/commits/7217c9c05d8a171ca24d592a48bea02ee7d6d31a

wyli: for the record, if we change the workdir to 'holohub_app_source', the error message is:

``` bash /workspace/holohub/benchmarks/release_benchmarking/run_benchmarks.sh --process benchmarks/release_benchmarking
find: ‘benchmarks/release_benchmarking/output’: No such file or directory
No logs found for pattern endoscopy_tool_tracking_3_[0-9]_1000_greedy_display_realtime
Error running command: bash /workspace/holohub/benchmarks/release_benchmarking/run_benchmarks.sh --process benchmarks/release_benchmarking
Exit code: 1
```

tbirdso: Good find, summarizing to make sure I understand findings:
1. "workdir": "<path>" has never worked in either `run` or `holohub run`
2. Apps with "workdir": "<holohub_app_bin>" succeeded because `holohub_app_bin` is already the default working directory

Wouldn't that mean that `release_benchmarking` also used `holohub_app_bin` previously, and we should keep it here? Or, could you share the error you saw when using the `holohub_app_bin` workdir as opposed to `holohub_root`?

wyli: correct, it seems `release_benchmarking` is broken, I tried `holohub_app_bin` and `holohub_app_source` both doesn't work.
```
# workdir: holohub_app_bin error:
cd /workspace/holohub/build/release_benchmarking/benchmarks/release_benchmarking
  File "/workspace/holohub/utilities/cli/holohub.py", line 1817, in <module>
    main()
  File "/workspace/holohub/utilities/cli/holohub.py", line 1813, in main
    cli.run()
  File "/workspace/holohub/utilities/cli/holohub.py", line 1805, in run
    args.func(args)
  File "/workspace/holohub/utilities/cli/holohub.py", line 828, in handle_run
    os.chdir(target_dir)
FileNotFoundError: [Errno 2] No such file or directory: '/workspace/holohub/build/release_benchmarking/benchmarks/release_benchmarking'
```

```
# workdir: holohub_app_source error:
cd /workspace/holohub/benchmarks/release_benchmarking
2025-07-01 13:38:08 $ bash /workspace/holohub/benchmarks/release_benchmarking/run_benchmarks.sh --process benchmarks/release_benchmarking
find: ‘benchmarks/release_benchmarking/output’: No such file or directory
No logs found for pattern endoscopy_tool_tracking_3_[0-9]_1000_greedy_display_realtime
Error running command: bash /workspace/holohub/benchmarks/release_benchmarking/run_benchmarks.sh --process benchmarks/release_benchmarking
Exit code: 1

```

changing workdir to `holohub_root` works
```
[command] python /workspace/holohub/benchmarks/holoscan_flow_benchmarking/generate_bar_graph.py /workspace/holohub/benchmarks/release_benchmarking/processed/multiai_ultrasound_3_[0-9]_1000_greedy_headless_offline/percentile_99.9_values.csv --app multiai_ultrasound --title multiai_ultrasound --quiet
```

tbirdso: >FileNotFoundError: [Errno 2] No such file or directory: '/workspace/holohub/build/release_benchmarking/benchmarks/release_benchmarking'

Can you confirm that you ran `./holohub build release_benchmarking` first create the build folder as documented [here](https://github.com/nvidia-holoscan/holohub/blob/7217c9c05d8a171ca24d592a48bea02ee7d6d31a/benchmarks/release_benchmarking/README.md#running-benchmarks-getting-started) before running with `holohub_app_bin` as the working dir? Or does HoloHub CLI behavior diverge in that the build folder is not necessarily created by this command?

wyli: I tried it's still having the issue. also the folder doesn't have the cmakelists.txt configs, so it shouldn't work in the bin directory.

tbirdso: Okay, on review I do now recall that the `run_benchmarks.sh` script is `pushd`-ing into relevant directories (HoloHub root, data output) before running relevant commands, so the working dir here is not a limiter. Thanks for bearing with me.

wyli: don't worry, it's a p0 feature we should make sure it's working as expected.

PR #969, opened 2025-07-02
improve CLI error message

wyli:
command such as `--build-args --test --dryrun` can be ambiguous, the previous `./dev_container` handled it differently, in the new CLI `--build-args=--test` is needed to make it clear that `--test` is a value provided to the option `build-args`.

adding more error messages when this is detected:

```
$ ./holohub build-container --build-args --test --dryrun
usage: ./holohub build-container [-h] [--base-img BASE_IMG] [--docker-file DOCKER_FILE] [--img IMG] [--no-cache] [--build-args BUILD_ARGS] [--verbose] [--dryrun] [--language {cpp,python}] [project]
./holohub build-container: error: argument --build-args: expected one argument

💡 Tip: ambiguous dash-prefixed arguments, use the equals format:
   Instead of: --build-args --test
   Use: --build-args=--test


💡 For help with the 'build-container' command:
  ./holohub build-container --help

```

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -26,7 +26,7 @@ import sys
 import tempfile
 from collections import defaultdict
 from pathlib import Path
-from typing import Optional
+from typing import List, Optional
 
 import utilities.cli.util as holohub_cli_util
 import utilities.metadata.gather_metadata as metadata_util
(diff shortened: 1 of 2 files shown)


tbirdso (approved):
Nice!

PR #960, opened 2025-06-30
Infra: Remove `dev_container` and `run` scripts

tbirdso:
Legacy scripts have been superseded by HoloHub CLI:
```
./holohub --help
```

PR #971, opened 2025-07-03
Add infrastructure to support custom holoscan DataLogger implementations in Holohub

grlee77:
The goal of this MR is to allow Holohub to support packaging custom implementations of the `holoscan::DataLogger` interface (new in Holoscan SDK v3.4: [Data Logging Docs](https://docs.nvidia.com/holoscan/sdk-user-guide/holoscan_data_logging.html)) in that same way that Operators are currently supported.

The build and CMake changes here are mostly a 1:1 copy of the logic currently used for operators so that each subfolder within the "data_loggers" folder gets built as a separate shared library. There are a few minor differences / simplifications:
- The OPTIONAL marker currently supported for operators is not supported for data loggers
- We also do not need to support GXF Extensions as dependencies for data loggers
- When writing loggers, the namespace `holoscan::data_loggers` should be used instead of `holoscan::ops`
- The CMake targets are prefixed with `holoscan::data_loggers` instead of `holoscan::ops` (e.g. `holoscan::data_loggers::message_logger`).

The example `MessageLogger` class bundled here is not intended as a fully featured logger. It is mainly just to demonstrate the required folder structure and CMake syntax to be used. We can potentially remove it in the future once we have some other example logger available on Holoscan.

I originally started this using the `dev_container` and `run` scripts, but also updated the new CLI accordingly.

I updated one existing application (`cvcuda_basic`) to use this logger (from both C++ and Python). This shows how DATA_LOGGERS should be listed in the `applications/CMakeLists.txt` for the application and how the `holoscan::data_loggers::message_logger` link target should be in the `CMakeLists.txt` for the C++ version of the application.

The following would build and run the application:
```bash
./holohub run cvcuda_basic
```

and due to the added `MessageLogger`, each frame will result on output like the following at the console
```cpp
[info] [message_logger.cpp:82] [timestamp: 1751549449056957] log_tensormap_data called for output port with unique id replayer.output
[info] [message_logger.cpp:82] [timestamp: 1751549449057201] log_tensormap_data called for input port with unique id image_processing.input_tensor
[info] [message_logger.cpp:82] [timestamp: 1751549449058125] log_tensormap_data called for output port with unique id image_processing.output_tensor
[info] [message_logger.cpp:82] [timestamp: 1751549449058314] log_tensormap_data called for input port with unique id holoviz.receivers:0
```

To see actual data contents one could use the `holoscan::data_loggers::BasicConsoleLogger` that ships with the core SDK instead, but I think it is useful to have a more minimal example fully contained within Holohub as a test case for the build script changes that were made.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing to HoloHub
 
-Welcome to HoloHub! We're excited that you're interested in contributing to the NVIDIA Holoscan developer community. This guide will help you understand how to make meaningful contributions to our collection of applications, operators, workflows, and tutorials.
+Welcome to HoloHub! We're excited that you're interested in contributing to the NVIDIA Holoscan developer community. This guide will help you understand how to make meaningful contributions to our collection of applications, operators, data loggers, workflows, and tutorials.
 
 ## Quick Start
 
@@ -26,7 +26,7 @@ New to HoloHub? Follow these steps:
 - [Code Quality and Standards](#code-quality-and-standards)
   - [Linting](#linting-and-code-quality)
   - [Testing](#testing)
-  - [Unit Testing Python Operators](#unit-testing-python-operators)
+  - [Unit Testing Python Operators and Data Loggers](#unit-testing-python-operators-and-data-loggers)
 - [Development Tools](#development-tools)
   - [Debugging](#debugging-and-performance)
   - [Performance](#performance)
@@ -36,7 +36,7 @@ New to HoloHub? Follow these steps:
 
 ## Introduction
 
-HoloHub is a collaborative ecosystem for the NVIDIA Holoscan SDK, featuring community-contributed applications, reusable operators, end-to-end workflows, and educational tutorials. Your contributions help expand the capabilities available to developers working on real-time AI applications across healthcare, industrial inspection, and other domains.
+HoloHub is a collaborative ecosystem for the NVIDIA Holoscan SDK, featuring community-contributed applications, reusable operators and data loggers, end-to-end workflows, and educational tutorials. Your contributions help expand the capabilities available to developers working on real-time AI applications across healthcare, industrial inspection, and other domains.
 
 Whether you're fixing a bug, adding a new feature, or sharing a complete application, this guide will help you contribute effectively to the HoloHub community.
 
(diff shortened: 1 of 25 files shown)


CONTRIBUTING.md
```diff
@@ -281,6 +282,17 @@ holohub/applications/your_app_name/
 └── CMakeLists.txt                  # For build system integration
 ```
 
+#### Data Loggers
+
+```text
+holohub/data_loggers/your_data_logger_name/
```
tbirdso, line 288: @wyli  could we support data logger scaffolding with `./holohub create` as a followup task?

data_loggers/message_logger/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  )
+
+add_library(holoscan::data_loggers::message_logger ALIAS message_logger)
+
+target_link_libraries(message_logger
+  PRIVATE
+    holoscan::core
+  )
+
+target_include_directories(message_logger INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
```
tbirdso, line 33: For portable installation:
```suggestion
target_include_directories(message_logger INTERFACE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
```

CONTRIBUTING.md
tbirdso: Also `./run list`

grlee77:
Hi @tbirdso

I reverted changes to `run` and `dev_container` as those are in the process of being removed (#960)

There is breaking API change to the `DataLogger::log_data` function signature in the upcoming v3.5 release (e.g. `std::any data` -> `const std::any& data`), so we can either:
1. wait to merge this for Holoscan v3.5 after updating to the new signature
2. merge now, and later make a corresponding breaking change here (or use `#ifdef` to support both APIs)

The actual implementation of the `log_data` method will not need change, there is just a minor change to const/reference types in the signature.

(The exact API signature for the `DataLogger` interface and methods is currently marked as experimental as we are still gathering real-world experience/feedback with it, so this is not entirely unexpected)

tbirdso:
> There is breaking API change to the DataLogger::log_data function signature in the upcoming v3.5 release

Thanks @grlee77 , we'll follow your lead here, would you prefer to merge now or wait for v3.5.0?

tbirdso:
Hi @grlee77 , could you please provide an update on this PR?

bhashemian:
Per discussion with @grlee77, I close this PR and we can revisit it later if we decide to add a concrete DataLogger on Holohub.

PR #972, opened 2025-07-03
CLI feature preset modes: ./holohub run <app> [mode]

wyli:
This is an initial implementation of preset mode:

## Scope
- a viable CLI implementation of mode presets `./holohub run|build|install <app> [mode]`
- As an example, the following apps' metadata are updated for various use cases:
    - `applications/holochat/metadata.json`
    - `applications/isaac_holoscan_bridge/metadata.json`
- a new schema specification for the 'modes'
- (existing metadata.json continue to work)

## New commands
```
  # List available modes for an app
  ./holohub modes <app>

  # Build with mode
  ./holohub build <app> [mode]

  # Run with mode  
  ./holohub run <app> [mode]

```

## Dev experience benefits

- simplify the current complex command of building and running the apps, for example
`
  ./holohub run --docker-opts="-u root -e ACCEPT_EULA=Y -e PRIVACY_CONSENT=Y -v ${HOME}/docker/isaac-sim/cache/kit:/isaac-sim/kit/cache:rw ..." isaac_holoscan_bridge
`
is simplified to `
  ./holohub run isaac_holoscan_bridge interactive
`
- better discoverability: ./holohub modes <app> shows what's available about an app
- readable and maintainable: configurations/recipes are standardized in metadata.json

## Example
```
$ ./holohub modes body_pose_estimation

Available modes for body_pose_estimation:

  default - Real-time pose estimation using V4L2 camera (/dev/video0) for live demos, real-time processing, and webcam testing
    Requirements: camera, model

  replayer - Offline pose estimation using pre-recorded video file for testing without camera, batch processing, and reproducible demos
    Requirements: model, video

  dds - Distributed pose estimation via DDS video streaming
    Requirements: dds, model

  custom_camera - Pose estimation with user-specified camera device for multiple cameras, non-standard devices, and custom hardware
    Requirements: camera, model
```

## TODOS
- ~discussion about the metadata format~
- ~cross check with more use cases~
- ~parameter overriding~
- ~unit tests after discussions~

### DIFF
--- a/applications/isaac_holoscan_bridge/metadata.json
+++ b/applications/isaac_holoscan_bridge/metadata.json
@@ -29,10 +29,40 @@
 			"Robotics"
 		],
 		"ranking": 3,
-		"requirements": {},
-		"run": {
-			"command": "python3 <holohub_app_source>/app.py",
-			"workdir": "holohub_bin"
+		"requirements": {
+			"gpu": {
+				"type": "hardware",
+				"name": "gpu",
+				"description": "NVIDIA GPU with at least 8GB memory required for Isaac Sim"
+			},
+			"isaac_sim": {
+				"type": "software",
+				"name": "Isaac Sim",
+				"description": "NVIDIA Isaac Sim simulation platform"
+			}
+		},
+		"modes": {
+			"interactive": {
+				"description": "Full Isaac Sim environment with interactive GUI and visualization for development, interactive debugging, and visual monitoring",
+				"requirements": ["gpu", "isaac_sim"],
+				"run": {
+					"command": "python3 <holohub_app_source>/app.py",
+					"workdir": "holohub_bin",
+					"docker_run_args": [
+						"-u root",
+						"-e ACCEPT_EULA=Y",
+						"-e PRIVACY_CONSENT=Y",
+						"-v ${HOME}/docker/isaac-sim/cache/kit:/isaac-sim/kit/cache:rw",
+						"-v ${HOME}/docker/isaac-sim/cache/ov:/root/.cache/ov:rw",
+						"-v ${HOME}/docker/isaac-sim/cache/pip:/root/.cache/pip:rw",
+						"-v ${HOME}/docker/isaac-sim/cache/glcache:/root/.cache/nvidia/GLCache:rw",
+						"-v ${HOME}/docker/isaac-sim/cache/computecache:/root/.nv/ComputeCache:rw",
+						"-v ${HOME}/docker/isaac-sim/logs:/root/.nvidia-omniverse/logs:rw",
+						"-v ${HOME}/docker/isaac-sim/data:/root/.local/share/ov/data:rw",
+						"-v ${HOME}/docker/isaac-sim/documents:/root/Documents:rw"
+					]
+				}
+			}
 		}
 	}
 }
--- a/utilities/cli/README.md
+++ b/utilities/cli/README.md
@@ -135,6 +135,204 @@ The new CLI introduces development lifecycle features:
 ./holohub env-info
 ```
 
+### Application Modes
+HoloHub applications with a 'modes' field in their `metadata.json` now support multiple **modes** - pre-configured setups for different use cases, hardware configurations, or deployment scenarios. This eliminates the need to remember complex command-line arguments for different use cases.
+
+#### **Understanding Modes**
+Modes potentially provide typical configurations for:
+- **Different hardware**
+- **Deployment scenarios** (e.g. development vs. production vs. cloud inference)
+- **Data sources** (e.g. live video vs. recorded files vs. distributed streaming)
+
+#### **Using Modes**
+
+Examples:
+
+```bash
+# Discover available modes for an application
+./holohub modes body_pose_estimation
+
+# Run an application in a specific mode
+./holohub run body_pose_estimation dds                    # DDS video streaming mode
+./holohub run endoscopy_tool_tracking aja                 # AJA capture card mode
+./holohub run holochat standalone                         # Local LLM inference mode
+
+# Build for a specific mode (if mode has special build requirements)
+./holohub build endoscopy_tool_tracking aja_overlay
+
+# Default behavior (no mode specified uses default_mode)
+./holohub run body_pose_estimation                        # Uses 'default_mode' if specified
+```
+
+#### **Mode vs. CLI Arguments**
+**Important**: When a mode defines specific options, you cannot override them with CLI arguments.
+This ensures mode integrity and prevents conflicting configurations.
+
+**Conflict Rules:**
+- If mode defines `run.docker_run_args` → Cannot use `--docker-opts`
+- If mode defines `build.depends` → Cannot use `--build-with`
+- If mode defines `build.docker_build_args` → Cannot use `--build-args`
+- If mode defines `build.cmake_options` → Cannot use `--configure-args`
+
+#### **For Application Developers**
+Applications can define modes in their `metadata.json`. Here's the complete field reference:
+
+##### **Mode Structure**
+Each mode is defined as a named object under the `modes` key:
+
+```json
+{
+  "metadata": {
+    "default_mode": "mode_name",
+    "modes": {
+      "mode_name": {
+        // Mode configuration fields
+      }
+    }
+  }
+}
+```
+
+##### **Application-Level Fields**
+**Optional Fields:**
+- **`default_mode`** *(string)*: Specifies which mode to use when no mode is explicitly provided. Only needed when there are multiple modes (2 or more).
+
+##### **Supported Fields for Each Mode**
+
+**Required Fields:**
+- **`description`** *(string)*: Human-readable description of what this mode does
+- **`run`** *(object)*: Run configuration (see run command fields below)
+
+**Optional Fields:**
+- **`requirements`** *(array of strings)*: List of dependency IDs required for this mode
+- **`build`** *(object)*: Build configuration (see build configuration fields below)
+
+##### **Build Configuration Fields (`build` object):**
+- **`depends`** *(array of strings)*: List of operators/dependencies to build with this mode
+- **`docker_build_args`** *(string or array)*: Docker **build** arguments (equivalent to CLI `--build-args`)
+  - Can be a single string: `"--build-arg CUSTOM=value"`
+  - Or an array: `["--build-arg", "CUSTOM=value"]`
+- **`cmake_options`** *(array of strings)*: Additional CMake configure arguments to pass to the local build step
+
+##### **Run Command Fields (`run` object):**
+- **`command`** *(string)*: Complete command to execute including all arguments
+- **`workdir`** *(string)*: Working directory for command execution
+- **`docker_run_args`** *(string or array)*: Docker **run** arguments used for both build and application containers (equivalent to CLI `--docker-opts`)
+  - Can be a single string: `"--privileged --net=host"`
+  - Or an array: `["--privileged", "--net=host"]`
+  - **Note**: These arguments apply to both the container that builds your application and the container that runs it
+  - **Common use cases**: GPU access (`--gpus=all`), device access (`--device=/dev/video0`), privileged operations (`--privileged`)
+
+##### **Complete Example**
+
+```json
+{
+  "metadata": {
+    "default_mode": "standard",
+    "modes": {
+
+      "standard": {
+        "description": "Standard camera input for development and testing",
+        "requirements": ["camera", "model"],
+        "run": {
+          "command": "python3 <holohub_app_source>/app.py --source camera",
+          "workdir": "holohub_bin"
+        }
+      },
+
+      "production": {
+        "description": "High-performance mode with GPU acceleration",
+        "requirements": ["gpu", "model", "tensorrt"],
+        "build": {
+          "depends": ["tensorrt_backend", "gpu_ops"],
+          "docker_build_args": ["--build-arg", "TENSORRT_VERSION=8.6", "--network=host"],
+          "cmake_options": ["-DUSE_TENSORRT=ON", "-DCUDA_ARCH=sm_86"]
+        },
+        "run": {
+          "command": "python3 <holohub_app_source>/app.py --backend tensorrt --optimization-level 3",
+          "workdir": "holohub_bin",
+          "docker_run_args": ["--gpus=all", "--shm-size=1g"]
+        }
+      },
+
+      "aja_hardware": {
+        "description": "AJA capture card with hardware overlay",
+        "requirements": ["aja_card_with_overlay", "model"],
+        "build": {
+          "depends": ["aja_source"]
+        },
+        "run": {
+          "command": "python3 <holohub_app_source>/app.py --source=aja --config=overlay.yaml",
+          "workdir": "holohub_bin",
+          "docker_run_args": [
+          "--privileged",
+          "--device=/dev/ajantv2",
+          "-v", "/dev:/dev"
+        ]
+        }
+      }
+    }
+  }
+}
+```
+
+##### **Advanced Pattern: Separate Build and Run Modes**
+
+For cases where build and run containers need different Docker configurations, you can create separate modes that share the same Docker image:
+
+```json
+{
+  "modes": {
+    "default_mode": "production_build",
+
+    "production_build": {
+      "description": "Build production image with network access",
+      "build": {
+        "depends": ["tensorrt_backend"],
+        "docker_build_args": ["--build-arg", "REGISTRY_TOKEN=xyz"]
+      },
+      "run": {
+        "docker_run_args": ["--network=host"],
+        "command": "echo 'Build complete'"
+      }
+    },
+
+    "production_run": {
+      "description": "Run production app with GPU access",
+      "run": {
+        "docker_run_args": ["--gpus=all", "--shm-size=1g"],
+        "command": "python3 <holohub_app_source>/app.py --gpu"
+      }
+    }
+  }
+}
+```
+
+**Usage:**
+```bash
+# Phase 1: Build with network access
+./holohub build myapp production_build
+
+# Phase 2: Run with GPU access (reuses image from phase 1)
+./holohub run myapp production_run --no-docker-build
+```
+
+Both modes automatically share the same Docker image name (`holohub:myapp`), so the run mode can use the image built by the build mode.
+
+##### **Key Points for Mode Development**
+- **`default_mode`** is only needed when you have multiple modes (2 or more). With a single mode, it will be used automatically.
+- **Mode names** must match pattern `^[a-zA-Z_][a-zA-Z0-9_]*$` (alphanumeric + underscore, can't start with number)
+- **Docker arguments** can be specified in two places for different purposes:
+  - `build.docker_build_args`: Docker **build** arguments for container image building (equivalent to CLI `--build-args`)
+  - `run.docker_run_args`: Docker **run** arguments for both build and application containers (equivalent to CLI `--docker-opts`)
+  - `run.docker_run_args` apply to both build containers (during compilation) and application containers (during execution)
+  - if `no-docker-build` is specified, `build.docker_build_args` is ignored
+- **Path placeholders** like `<holohub_app_source>`, `<holohub_data_dir>` are supported in commands
+- **CLI arguments cannot override** mode-specific configurations - modes are self-contained
+- **CLI arguments are allowed** only for options not defined by the selected mode
+- **Requirements** reference dependency IDs defined elsewhere in the metadata
+- **Modes provide complete control** over both build and runtime behavior for different deployment scenarios
+
 
 ### **Granular Build Control**
 The new architecture provides precise control over your development workflow.
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -166,6 +166,7 @@ class HoloHubCLI:
         )
         self.subparsers["build"] = build
         build.add_argument("project", help="Project to build")
+        build.add_argument("mode", nargs="?", help="Mode to build (optional)")
         build.add_argument(
             "--local", action="store_true", help="Build locally instead of in container"
         )
@@ -216,6 +217,7 @@ class HoloHubCLI:
         )
         self.subparsers["run"] = run
         run.add_argument("project", help="Project to run")
+        run.add_argument("mode", nargs="?", help="Mode to run (optional)")
         run.add_argument("--local", action="store_true", help="Run locally instead of in container")
         run.add_argument("--verbose", action="store_true", help="Print extra output")
         run.add_argument(
@@ -234,6 +236,7 @@ class HoloHubCLI:
             help="Additional arguments to pass to the application executable, "
             "example: --run-args=--flag or --run-args '-c config/file'",
         )
+
         run.add_argument(
             "--build-with",
             dest="with_operators",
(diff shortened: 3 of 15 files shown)


applications/body_pose_estimation/metadata.json
```diff
@@ -24,24 +24,87 @@
...
 		"tags": ["Computer Vision and Perception", "Distributed", "Tensor Optimization", "Human Body Pose Estimation", "DDS", "RTI Connext", "Video", "Visualization"],
 		"ranking": 2,
-		"dependencies": {
-			"data": [
-				{
-					"model": "https://docs.ultralytics.com/tasks/pose/",
-					"video": "https://www.pexels.com/video/a-woman-showing-her-ballet-skill-in-turning-one-footed-5385885/"
+		"dependencies": [
+			{
+				"id": "model",
```
agirault: why not have `id` be a key instead, like the key for each mode, to ensure uniqueness?

utilities/metadata/project.schema.json
```diff
@@ -141,6 +141,48 @@
         "description"
       ]
     },
+    "dependency": {
```
agirault: we now have `dependency`, `application_dependencies`, `dependencies`, and `depends`. I think we need to improve naming and make sure we need all of these

applications/body_pose_estimation/metadata.json
```diff
@@ -24,24 +24,87 @@
...
+					"video_device": {
+						"description": "Path to V4L2 video device (examples: /dev/video1, /dev/video2)",
+						"type": "string",
+						"default": "/dev/video0"
+					}
+				},
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"run_args": ["--video_device ${PARAM:video_device}"],
```
agirault: bit scared of how we make sure not to mix PARAM flags with regular holohub cli flags

applications/endoscopy_tool_tracking/python/metadata.json
```diff
@@ -28,19 +28,94 @@
...
-			"data": [
-					{
-							"name": "Holoscan Sample App Data for AI-based Endoscopy Tool Tracking",
-							"description": "This resource contains the convolutional LSTM model for tool tracking in laparoscopic videos by Nwoye et. al [1], and a sample surgical video.",
-							"url": "https://catalog.ngc.nvidia.com/orgs/nvidia/teams/clara-holoscan/resources/holoscan_endoscopy_sample_data",
-							"version": "20230222"
-					}
-			]
-		},
-		"run": {
```
agirault: you mentioned keeping support for this for backward compat (great). 

We should have schema validation that fails if there is both "modes" and "run"

applications/isaac_holoscan_bridge/metadata.json
```diff
@@ -29,10 +29,119 @@
...
+					],
+					"requirements": ["gpu", "isaac_sim"]
+				}
+			},
+			"headless": {
+				"description": "Headless Isaac Sim for automated workflows and CI/CD including automated testing, CI/CD pipelines, and batch processing",
+				"run": {
+					"command": "python3 <holohub_app_source>/app.py",
+					"workdir": "holohub_bin",
+					"run_args": ["--headless"],
```
agirault, line 73: Can we merge `command` and `run_args` when under `modes` (keep them separate for backward compatibility under the top level `run`) ?

utilities/metadata/project.schema.json
```diff
@@ -158,19 +200,152 @@
...
+        },
+        "run_args": {
+          "type": "array",
+          "items": {
+            "type": "string",
+            "description": "Runtime arguments. Supports parameter substitution using ${param_name}, ${env:VAR_NAME}, and ${holohub:placeholder}"
+          },
+          "description": "Array of runtime arguments to pass to the command"
+        },
+        "docker_args": {
```
agirault: we discussed calling this `docker_run_flags` to leave space for the `docker_build_args` in the future.

utilities/metadata/project.schema.json
```diff
@@ -158,19 +200,152 @@
...
         "command": {
-          "type": "string"
+          "type": "string",
+          "description": "Command to execute. Supports parameter substitution using ${param_name}, ${env:VAR_NAME}, and ${holohub:placeholder}"
         },
         "workdir": {
+          "type": "string",
+          "description": "Working directory for command execution"
+        },
+        "run_args": {
```
agirault: Can we merge `command` and `run_args` when under `modes` (keep them separate for backward compatibility under the top level `run` for now) ?

agirault: UNLESS we anticipate being able to override the `run_args` only and not the command, like we do for `docker_args`, instead of the `--param` support you worked on... at some point we'll have to decide what granularity we want to provide through the CLI vs just letting users with custom parameters run the command themselves.

agirault: cc @tbirdso @jjomier

utilities/metadata/project.schema.json
```diff
@@ -158,19 +200,152 @@
...
+          },
+          "description": "Array of Docker run arguments"
+        },
+        "requirements": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "List of dependency IDs required during run phase"
+        }
```
agirault: Per discussion, maybe worth discussing in future work once we have a use case?

utilities/metadata/project.schema.json
```diff
@@ -158,19 +200,152 @@
...
+      },
+      "description": "Run command configuration with parameter substitution support. Available substitution patterns: ${PARAM:param_name} for mode parameters, ${VAR_NAME} for environment variables"
+    },
+    "parameter_definition": {
```
agirault: per discussion, maybe worth discussing design later on.

applications/body_pose_estimation/metadata.json
```diff
@@ -24,24 +24,87 @@
...
+				"build": {
+					"depends": ["dds_video_subscriber", "dds_video_publisher"],
+					"docker_args": ["--ssh default"],
+					"requirements": ["dds"]
+				},
+				"run": {
+					"command": "python3 <holohub_app_source>/body_pose_estimation.py",
+					"workdir": "holohub_bin",
+					"run_args": ["--source dds"],
+					"docker_args": [
```
agirault: Ideally these would support single strings with all args or arrays of strings like now, for cases with short list of args.

agirault: but not a blocker, fine to start as only array and update later

applications/body_pose_estimation/metadata.json
agirault: assuming we keep it

applications/body_pose_estimation/metadata.json
wyli: removing the parameter implementation in this version

applications/body_pose_estimation/metadata.json
wyli: refactoring it to dictionaries instead of list of dictionaries

utilities/metadata/project.schema.json
wyli: merging modes.build.requirements and modes.run.requirements into modes.requirements for now

utilities/metadata/project.schema.json
wyli: simplifying this and keep the necessary ones for backward comp

utilities/metadata/project.schema.json
wyli: merging `command` `run_args` into `command` under `modes.run`

utilities/metadata/project.schema.json
wyli: so on this topic of new keys, I want to highlight that the current definition is
```
'build': {'docker_args': ...},
'run': {'docker_args': ...},
```

the comment suggests:
```
'build': {'docker_build_args': ...},
'run': {'docker_run_args': ...}
```

any preference? @tbirdso @jjomier @bhashemian

tbirdso: +1 for the current definition:
```
'build': {'docker_args': ...},
'run': {'docker_args': ...},
```

utilities/metadata/project.schema.json
agirault: can you clarify what are the fields now and how they differ?

utilities/metadata/project.schema.json
agirault: hold on a minute. that's not the suggestion.


This is to accomodate this future design we discussed in our previous design review:

```
build: {
  depends: { <operator_dict> },
  docker_env: <build_env>,
  docker_build_args: { <explicit_build_args>, … },
  docker_run_flags: { <explicit_flags>, … },
  docker_run_capabilities: { <preset_capabilities>, … },
  cmake_flags: { … }
},
run: {
  docker_env: <run_env>,
  docker_build_args: { <explicit_build_args>, … },
  docker_run_flags: { <explicit_flags>, … }
  docker_run_capabilities: { <preset_capabilities>, … }
  cmd: <command>,
  workdir: <dir>
}
```

there is a difference between `docker_build_args` (passed to `docker build`) and `docker_run_flags` (passed to `docker run`), vs `build` (build your app/operator) and `run` (run your app).

~~This is to accommodate a different run environment (dockerfile) than build environment.~~ Edit: wrong sorry, I'm mixing it myself. This is to differentiate between `docker build` and `docker run` flags.

utilities/metadata/project.schema.json
agirault: I thought we had agreed in the call to not add this extra "dependency/requirements" fields until we have a clear usage of them (and better naming)

wyli: sorry I misunderstood the suggestion. We do have a planned usage to skip app tests based on `modes.requirements` (e.g. when the required hardware is not available) in the near future, but we don't have a clear use case of separating `build.requirements` and `run.requirements`

utilities/metadata/project.schema.json
wyli: thanks, I missed some changes, now the only new is `depends` which corresponds to `--with operator1;operator2` in the CLI commands

utilities/metadata/project.schema.json
wyli: this is a good comment, not sure if we want to fully separate the concepts of 'dev container' and 'runtime container', do/will we have a use case of building the app with one set of docker build/run flags, and then running the app with another set of docker build/run flags? to me it sounds a bit too complicated for an app but I'm open to any suggestions. 

The current `./holohub run` doesn't fully support this.. the `run` will build by default but doesn't support additional docker flags. For the advanced use cases using the current CLI will take two steps:

```
# first step builds in a persistent directory
./holohub build --build-args="docker build arg" --docker-opts="docker run args" --configure-args='cmake args"
# second use the builds for the persistent directory and run the app
./holohub run --build-args=... --docker-opts=... --run-args="app run args"
```

utilities/metadata/project.schema.json
agirault: can we rename the "dependency" field as "requirements" ?

wyli: > can we rename the "dependency" field as "requirements" ?

do you mean renaming `dependencies` to `requirements` in the schema and everywhere? currently many apps have the `dependencies` section in metadata.json

utilities/cli/README.md
```diff
@@ -135,6 +135,196 @@ The new CLI introduces development lifecycle features:
...
+
+```json
+{
+  "modes": {
+
+    "production_build": {
+      "description": "Build production image with network access",
+      "build": {
+        "depends": ["tensorrt_backend"],
+        "docker_args": ["--build-arg", "REGISTRY_TOKEN=xyz"]
```
tbirdso: Did we land on only supporting a common container for building and running applications? Or did we still want to consider `docker_build_args` and `docker_run_args` for the application build, which can be distinct from `docker_build_args` and `docker_run_args` for the application run? @agirault

utilities/cli/README.md
```diff
@@ -135,6 +135,196 @@ The new CLI introduces development lifecycle features:
...
+  }
+}
+```
+
+**Usage:**
+```bash
+# Phase 1: Build with network access
+./holohub build myapp production_build
+
+# Phase 2: Run with GPU access (reuses image from phase 1)
```
tbirdso, line 316: Can we use a different image? Or not supported?

utilities/cli/README.md
```diff
@@ -135,6 +135,196 @@ The new CLI introduces development lifecycle features:
...
+##### **Complete Example**
+
+```json
+{
+  "metadata": {
+    "modes": {
+
+      "default": {
+        "description": "Standard camera input for development and testing",
+        "requirements": ["camera", "model"],
```
tbirdso, line 236: Remind me, are requirements listed under this field simply for documentation purposes right now?

utilities/cli/holohub.py
```diff
@@ -762,12 +888,22 @@ def handle_run(self, args: argparse.Namespace) -> None:
         skip_docker_build, skip_local_build = holohub_cli_util.check_skip_builds(args)
         is_local_mode = args.local or os.environ.get("HOLOHUB_BUILD_LOCAL")
 
-        if is_local_mode:
-            if args.docker_opts:
-                holohub_cli_util.fatal(
```
tbirdso: Should we keep this check?

utilities/cli/holohub.py
```diff
@@ -916,13 +1057,19 @@ def handle_run(self, args: argparse.Namespace) -> None:
...
             )
 
-            run_cmd = f"{self.script_name} run {args.project} --language {language} --local"
+            run_cmd = f"{self.script_name} run {args.project}"
+            if mode_name:
+                run_cmd += f" {mode_name}"
+            run_cmd += f" --language {language}"
+            # Only add --local if we're not in a nested container scenario
+            if not (mode_config and "run" in mode_config and "docker_args" in mode_config["run"]):
+                run_cmd += " --local"
```
tbirdso: Not sure I follow, could you please clarify the "nested container scenario" here? Are we trying to support docker-in-docker?

utilities/cli/README.md
wyli: yes we can use a different image, that'll be a simpler use case: two modes sharing some host folders by using the same mounting parameter, the rest will be independent for each mode.

utilities/cli/README.md
wyli: yes, it's been changed to list here, I plan to use it to include/exclude in application tests.

utilities/cli/README.md
wyli: imo we can keep keep the simple use case for now, each mode corresponds to one container build/run seems to cover all the apps we have.

utilities/cli/README.md
tbirdso: Cool, cc @agirault if useful for ANO

utilities/cli/README.md
tbirdso: cc @agirault for comment about current vs future reqs here

utilities/cli/holohub.py
wyli: I couldn't remember the use case here... removed the logic and all tests work ok, so I just remove them.

utilities/cli/holohub.py
wyli: added it back as I don't have a preference here... just to note that there are other container only args such as `ssh-x11`, `--add-volume`. Also I feel maintaining a list of all the container-only args here is tedious.

tbirdso (commented):
Looks good to me, would be good to have review and approval from @agirault as well before merge

utilities/metadata/project.schema.json
agirault: yes. The names don't match between the list, and their reference in the presets. if we need to change it, we should change it earlier than later. It also causes confusions with the addition of the `depends` field now, per discussion here: https://github.com/nvidia-holoscan/holohub/pull/972#discussion_r2216398969

utilities/metadata/project.schema.json
agirault: The point was to prepare for the design we had discussed to eventually support separate build and run containers, and not having to change the interface in the future once more when that happens.

If we deviate from this design we should do a proper review of the design first.

utilities/cli/README.md
agirault: thank you. duplicate of this thread: https://github.com/nvidia-holoscan/holohub/pull/972#discussion_r2246086282

utilities/cli/README.md
```diff
@@ -135,6 +135,196 @@ The new CLI introduces development lifecycle features:
 ./holohub env-info
 ```
 
+### **Application Modes**
```
agirault: nitpick: no need to bold the titles in this page, they're bold already

utilities/cli/README.md
```diff
@@ -135,6 +135,196 @@ The new CLI introduces development lifecycle features:
...
+**Optional Fields:**
+- **`requirements`** *(array of strings)*: List of dependency IDs required for this mode
+- **`build`** *(object)*: Build configuration (see build configuration fields below)
+
+##### **Build Configuration Fields (`build` object):**
+- **`depends`** *(array of strings)*: List of operators/dependencies to build with this mode
+- **`docker_args`** *(string or array)*: Docker **build** arguments (equivalent to CLI `--build-args`)
+  - Can be a single string: `"--build-arg CUSTOM=value"`
+  - Or an array: `["--build-arg", "CUSTOM=value"]`
+- **`configure_args`** *(array of strings)*: Additional configure arguments (such as cmake flags) to pass to the local build step
```
agirault: should we be explicit and name this `cmake_configure_args` ? cc @jjomier

utilities/cli/README.md
```diff
@@ -135,6 +135,196 @@ The new CLI introduces development lifecycle features:
...
+}
+```
+
+**Usage:**
+```bash
+# Phase 1: Build with network access
+./holohub build myapp production_build
+
+# Phase 2: Run with GPU access (reuses image from phase 1)
+./holohub run myapp production_run --no-docker-build
```
agirault: Is this new? was this design discussed?

Sorry but I see some big issues with this design:
- `run` instruction in `build` mode doesn't actually run an app in a certain mode, (here it does just echo)
- I can't tell what the behavior of the `build` instruction in `run` mode would look like, if it's even defined, but the concern is the same as above, and the example here requiring to add `--no-docker-build` illustrates that problem
- the main point for build vs run image won't be the flags, but the docker image itself. I don't believe this MR touches (nor should it) on docker environment selection (as in, it's still either the Dockerfile in the same folder, or the top one)

run/build are instructions, I think conflating them with modes is a mistake. The designed we reviewed internally plans for future support of a separate run vs build environments attached to instructions, not to modes.

utilities/metadata/project.schema.json
agirault: See comment above actually, I adjusted the last paragraph where I mixed things up.

Other attempt to clarify. I think we need to clearly dissociate both types of args/flags which `docker_args` does not do:

1. `docker_build_args`: what you pass to `docker build` ([docs](https://docs.docker.com/reference/cli/docker/buildx/build/))
   - passed with `--build-args` in the holohub cli - I think should be renamed to `--docker-build-args` to avoid confusions with cmake build 
2. `docker_run_flags`: what you pass to `docker run` ([docs](https://docs.docker.com/reference/cli/docker/container/run/))
   - passed with `--docker-opts` in the cli - I think should be renamed to `--docker-run-flags`

utilities/metadata/project.schema.json
wyli: `dependencies` is being used for all the metadata.json, do we have enough motivation to introduce the change? I feel we can do that in another iteration and keep this PR focusing on the new feature cc @jjomier @tbirdso

utilities/cli/README.md
wyli:
```suggestion
### Application Modes
```

utilities/cli/README.md
wyli: this is originally from the `run` script https://github.com/nvidia-holoscan/holohub/blob/0d64dd6022aae855d19448043c0c4e5083603e2c/run#L666
and is used in the current `holohub` script. `cmake_configure_args` or `cmake_config` is also good, I don't have preference here. please let me know your suggestions @tbirdso @jjomier

utilities/cli/README.md
wyli: this doesn't introduce new design, this is to show the simple API we've been discussing can achieve more complicated complex use case if needed...

wyli: I'll change `build.docker_args` to `build.docker_build_args` and `run.docker_args` to `run.docker_run_args`

wyli: renamed in https://github.com/nvidia-holoscan/holohub/pull/972/commits/3a3975cd0132f0e45a2b890fc09aed37f6ce2418

utilities/metadata/project.schema.json
agirault: yes and no. In my opinion, since we introduced `depends` and `requirement` keys in this MR, aligning with `requirement` and not `depends` is important. Also, holohub configs are still all self-contained in this repo, so better do breaking changes to the schemas while we can since it should not impact anyone directly.

applications/isaac_holoscan_bridge/metadata.json
```diff
@@ -29,10 +29,40 @@
...
+				"description": "NVIDIA GPU with at least 8GB memory required for Isaac Sim"
+			},
+			"isaac_sim": {
+				"type": "software",
+				"name": "Isaac Sim",
+				"description": "NVIDIA Isaac Sim simulation platform"
+			}
+		},
+		"modes": {
+			"interactive": {
```
jjomier, line 45: We should enforce a default mode, why is this not a default mode?

utilities/metadata/project.schema.json
wyli: Thanks, this is done in https://github.com/nvidia-holoscan/holohub/pull/1035

utilities/cli/README.md
wyli: thanks, this is updated to `cmake_config` from `configure_args` in the preset modes.

applications/isaac_holoscan_bridge/metadata.json
wyli: thanks, the default mode logic is now updated:
- in the metadata.json validation, if there's more than 1 mode, `'default_mode': <name>` must be specified
- in the CLI implementation, it's an error only if the user doesn't specify a mode in the command and there are multiple modes without defining a 'default_mode'. 3f5982ba782658ed913783b8b354bd662bd3082f

.github/workflows/check_docker.yml
```diff
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, mode-preset-2]
```
```suggestion
    branches: [main, holoscan-sdk-lws2]
```

wyli:
fyi @AndreasHeumann @NigelNelson I've updated the holochat and isaac_holoscan_bridge app to demonstrate the new feature of defining CLI parameters in the metadata. Please let me know any concerns (plan to merge this soon).

NigelNelson:
> fyi @AndreasHeumann @NigelNelson I've updated the holochat and isaac_holoscan_bridge app to demonstrate the new feature of defining CLI parameters in the metadata. Please let me know any concerns (plan to merge this soon).

Looks good to me!

ghost:
> fyi @AndreasHeumann @NigelNelson I've updated the holochat and isaac_holoscan_bridge app to demonstrate the new feature of defining CLI parameters in the metadata. Please let me know any concerns (plan to merge this soon).

Looks good to me as well, thanks!

PR #973, opened 2025-07-04
CLI consolidate pkg installl and sudo usage

wyli:
Reviewed CLI `handle_setup`
- only require `sudo` when it's necessary 
- when sudo is used, show command and reason when prompting for input
- for environment without `sudo` just run command as it is

e.g.

```
$ ./holohub setup --dryrun
INFO: Package wget is already installed
INFO: Package xvfb is already installed
INFO: Package git is already installed
INFO: Package unzip is already installed
INFO: Package ffmpeg is already installed
INFO: Package ninja-build is already installed
INFO: Package libv4l-dev is already installed
INFO: Installing libcudnn9-cuda-12=9.10.2.21-1
[SUDO REQUIRED] Package management requires root privileges
2025-07-04 10:37:35 [dryrun] $ sudo apt-get update
[SUDO REQUIRED] Package management requires root privileges
2025-07-04 10:37:35 [dryrun] $ sudo apt install --no-install-recommends -y libcudnn9-cuda-12=9.10.2.21-1
INFO: Installing libcudnn9-dev-cuda-12=9.10.2.21-1
[SUDO REQUIRED] Package management requires root privileges
...
```

### DIFF
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2, sudo-warning]
+    branches: [main, holoscan-sdk-lws2]
 
 permissions:
   contents: read
--- a/utilities/cli/util.py
+++ b/utilities/cli/util.py
@@ -304,7 +304,7 @@ def get_host_gpu() -> str:
         )
         if not output or "Orin (nvgpu)" in output.decode():
             return "igpu"
-    except Exception:
+    except subprocess.CalledProcessError:
         return "dgpu"
 
     return "dgpu"


.github/workflows/check_docker.yml
```diff
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, sudo-warning]
```
wyli: remove before merging

```suggestion
    branches: [main, holoscan-sdk-lws2]
```

applications/ehr_query_llm/fhir/README.md
```diff
@@ -29,14 +29,14 @@ There are several ways to build and run this application and package it as a Hol
 
 It is further expected that you have read the [HoloHub README](../../../README.md), have cloned the HoloHub repository to your local system, and the current working directory is the HoloHub root, `holohub`.
 
-**_Note_**:
```
bhashemian: Thanks @wyli for this PR. Just a quick note: the tailing double spaces at the end of a line in markdown is similar to <br>. Here and elsewhere, could please bring them back? Thanks

applications/ehr_query_llm/fhir/README.md
```diff
@@ -29,14 +29,14 @@ There are several ways to build and run this application and package it as a Hol
 
 It is further expected that you have read the [HoloHub README](../../../README.md), have cloned the HoloHub repository to your local system, and the current working directory is the HoloHub root, `holohub`.
 
-**_Note_**:  
+**_Note_**:
```
wyli:
```suggestion
**_Note_**:  
```

applications/ehr_query_llm/fhir/README.md
```diff
@@ -29,14 +29,14 @@ There are several ways to build and run this application and package it as a Hol
...
 The application listens for request messages to start retrieving resources from the server and then publishes the results, so another application is needed to drive this workflow (e.g., the LLM application). To help with simple testing, a Python script is provided as part of this application, and its usage is described below in this [section](#test-the-running-application).
 
 ### Quick Start Using Holohub Container
 
 This is the simplest and fastest way to start the application in a Holohub dev container and get it ready to listen to request messages.
 
-**_Note_**:  
+**_Note_**:
```
wyli:
```suggestion
**_Note_**:  
```

applications/ehr_query_llm/fhir/README.md
wyli: thanks, I didn't know this syntax

utilities/cli/util.py
```diff
@@ -144,22 +148,90 @@ def fatal(message: str) -> None:
...
 
 
+def _get_maybe_sudo() -> str:
+    """Get sudo command if available, with caching to avoid repeated subprocess calls"""
+    global _sudo_available
+
+    if _sudo_available is not None:
+        return _sudo_available
+    try:
+        subprocess.run(["sudo", "--version"], capture_output=True, check=True, timeout=5)
```
tbirdso: At a glance it looks like this is adding a one time `sudo --version` check every time a HoloHub CLI command is called that invokes `run_command`.

A quick local check on IGX implies this adds maybe 30-50 ms to command execution, so probably not worth spending effort on. If we do find a need to avoid this extra overhead, could consider checking whether the command needs `sudo` first before polling for `sudo` on the system.

utilities/cli/util.py
```diff
@@ -319,6 +391,61 @@ def __init__(self, package_name: str, version_pattern: str, message: str = None)
...
+        result = subprocess.run(
+            ["dpkg", "-l", package_name], capture_output=True, text=True, check=False
+        )
+        return result.returncode == 0 and "ii" in result.stdout
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        return False
+
+
+def ensure_apt_updated(dry_run: bool = False) -> None:
+    """Ensure apt package list is updated, but only once per session"""
```
tbirdso, line 431: The risk of missing some update in some long-running session might outweigh the benefit of enforcing "only once" here.

Could we run `apt update` once as the first command in sequence, and document an implicit assumption in subsequent package installs that the apt cache must already be updated? Note that `apt install` already makes the same assumption.

utilities/cli/util.py
```diff
@@ -382,6 +508,7 @@ def install_cuda_dependencies_package(
 
         # Install the package
         info(f"Installing {package_name}={matching_version}")
+        ensure_apt_updated(dry_run=dry_run)
```
tbirdso: Suggest assuming apt cache is up to date

utilities/cli/util.py
```diff
@@ -319,6 +391,61 @@ def __init__(self, package_name: str, version_pattern: str, message: str = None)
...
+            info(f"Package {package_name} will be checked/installed with version specification")
+        elif not is_package_installed(package_name):
+            missing_packages.append(package)
+        else:
+            info(f"Package {package_name} is already installed")
+
+    if missing_packages:
+        ensure_apt_updated(dry_run=dry_run)
+        install_cmd = ["apt", "install"] + apt_options + missing_packages
+        run_command(install_cmd, dry_run=dry_run)
```
tbirdso, line 472: What is the behavior if any of the packages are not found in the apt cache?

utilities/cli/util.py
```diff
@@ -382,6 +508,7 @@ def install_cuda_dependencies_package(
 
         # Install the package
         info(f"Installing {package_name}={matching_version}")
+        ensure_apt_updated(dry_run=dry_run)
         run_command(
```
tbirdso: Should we update to use `install_packages_in_missing` in CUDA setup?

tbirdso (commented):
Thanks @wyli ! Can you please provide a non-dryrun output as well to compare? Does only the first `sudo` call block when running `./holohub setup` in a clean environment such as the [CUDA base container](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda)?

wyli:
> Thanks @wyli ! Can you please provide a non-dryrun output as well to compare? Does only the first `sudo` call block when running `./holohub setup` in a clean environment such as the [CUDA base container](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda)?

yes, it'll block the first sudo call, details depend on the os sudo password timeout config, it's usually 15min. some logs without dryrun:

<details>

with root user:
```
./holohub setup
[SUDO REQUIRED] Package management requires root privileges
2025-07-08 09:20:31 $ apt-get update
Hit:1 http://archive.ubuntu.com/ubuntu noble InRelease
Hit:2 http://archive.ubuntu.com/ubuntu noble-updates InRelease
Hit:3 http://archive.ubuntu.com/ubuntu noble-backports InRelease
Hit:4 http://security.ubuntu.com/ubuntu noble-security InRelease
Reading package lists... Done                               
[SUDO REQUIRED] Package management requires root privileges
2025-07-08 09:20:32 $ apt install --no-install-recommends -y wget xvfb git unzip ffmpeg ninja-build libv4l-dev
Reading package lists... Done
Building dependency tree... Done

```

with my local environment:
```
./holohub setup
INFO: Package wget is already installed
INFO: Package xvfb is already installed
INFO: Package git is already installed
INFO: Package unzip is already installed
INFO: Package ffmpeg is already installed
INFO: Package ninja-build is already installed
INFO: Package libv4l-dev is already installed
INFO: Package libcudnn9-cuda-12 found with version 9.10.2.21-1
INFO: Installing libcudnn9-dev-cuda-12=9.10.2.21-1
[SUDO REQUIRED] Package management requires root privileges
2025-07-08 09:23:53 $ sudo apt-get update
[sudo] password for wenqil: 
Get:1 file:/var/holoscan-local-repo-ubuntu2204-3.3.0  InRelease [1,572 B]
Get:1 file:/var/holoscan-local-repo-ubuntu2204-3.3.0  InRelease [1,572 B]
Hit:2 http://gb.archive.ubuntu.com/ubuntu jammy InRelease
Hit:3 http://gb.archive.ubuntu.com/ubuntu jammy-updates InRelease
Hit:4 http://gb.archive.ubuntu.com/ubuntu jammy-backports InRelease
Get:5 https://nvidia.github.io/libnvidia-container/stable/deb/amd64  InRelease [1,477 B]
Get:6 https://cli.github.com/packages stable InRelease [3,917 B]
Hit:7 https://linux.teamviewer.com/deb stable InRelease        
Hit:8 https://download.docker.com/linux/ubuntu jammy InRelease                                                                                                                                                                           
Hit:9 https://packages.microsoft.com/repos/code stable InRelease                                                                                                                                                                         
Hit:10 http://download.opensuse.org/repositories/home:/bluca:/openconnect/Ubuntu_22.04  InRelease                                                                                                                                        
Hit:11 https://deb.nodesource.com/node_16.x nodistro InRelease                                                                                                                                      
Hit:12 http://security.ubuntu.com/ubuntu jammy-security InRelease                                                                                                             
Hit:13 https://dl.google.com/linux/chrome/deb stable InRelease                                                      
Hit:14 https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64  InRelease  
Hit:15 https://apt.kitware.com/ubuntu jammy InRelease                  
Fetched 5,394 B in 2s (2,607 B/s)
Reading package lists... Done
[SUDO REQUIRED] Package management requires root privileges
2025-07-08 09:23:59 $ sudo apt install --no-install-recommends -y libcudnn9-dev-cuda-12=9.10.2.21-1
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  libcudnn9-headers-cuda-12
The following NEW packages will be installed
  libcudnn9-dev-cuda-12 libcudnn9-headers-cuda-12
0 to upgrade, 2 to newly install, 0 to remove and 25 not to upgrade.
Need to get 47.2 kB of archives.
After this operation, 284 kB of additional disk space will be used.
Get:1 https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64  libcudnn9-headers-cuda-12 9.10.2.21-1 [34.3 kB]
Get:2 https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64  libcudnn9-dev-cuda-12 9.10.2.21-1 [12.9 kB]
Fetched 47.2 kB in 2s (25.2 kB/s)                 
Selecting previously unselected package libcudnn9-headers-cuda-12.
(Reading database ... 399842 files and directories currently installed.)
Preparing to unpack .../libcudnn9-headers-cuda-12_9.10.2.21-1_amd64.deb ...
Unpacking libcudnn9-headers-cuda-12 (9.10.2.21-1) ...
Selecting previously unselected package libcudnn9-dev-cuda-12.
Preparing to unpack .../libcudnn9-dev-cuda-12_9.10.2.21-1_amd64.deb ...
+ exit 0
Unpacking libcudnn9-dev-cuda-12 (9.10.2.21-1) ...
Setting up libcudnn9-headers-cuda-12 (9.10.2.21-1) ...
Setting up libcudnn9-dev-cuda-12 (9.10.2.21-1) ...
^CError running command: sudo apt install --no-install-recommends -y libcudnn9-dev-cuda-12=9.10.2.21-1

```

</details>

utilities/cli/util.py
wyli: thanks for checking. the `_sudo_available` variable will cache the output, so the overhead is only for the very first run_command for any `./holohub` process. I've updated it to use `shutil.which("sudo")` it should be faster. Now it doesn't check if `sudo` is actually functional but a case that `sudo` path exists but not working shouldn't be critical this CLI should handle..

utilities/cli/util.py
wyli: in that case the entire process will exit after the `apt install`. I've also updated the `handle_setup` to make it clear that there are these stages:
```
        holohub_cli_util.install_packages_if_missing(
            ["wget", "xvfb", "git", "unzip", "ffmpeg", "ninja-build", "libv4l-dev"],
            dry_run=args.dryrun,
        )
        holohub_cli_util.setup_cmake(dry_run=args.dryrun)
        holohub_cli_util.setup_python_dev(dry_run=args.dryrun)
        holohub_cli_util.setup_ngc_cli(dry_run=args.dryrun)
        holohub_cli_util.setup_cuda_dependencies(dry_run=args.dryrun)
```
any error within the stage will exit the process.

didn't add any retry logic here as it'll be more intuitive to manually run `./holohub setup` again if needed

utilities/cli/holohub.py
```diff
@@ -1245,282 +1241,32 @@ def _install_template_deps(self, dry_run: bool = False) -> None:
...
-        holohub_cli_util.run_command(
-            ["apt", "install", "--no-install-recommends", "-y", "ffmpeg"],
-            dry_run=args.dryrun,
-        )
-
-        # Install libv4l-dev
-        holohub_cli_util.run_command(
-            ["apt-get", "install", "--no-install-recommends", "-y", "libv4l-dev"],
+        holohub_cli_util.install_packages_if_missing(
+            ["wget", "xvfb", "git", "unzip", "ffmpeg", "ninja-build", "libv4l-dev"],
```
tbirdso, line 1290: Rather than running `apt update` in `install_packages_if_missing`, could we add a single cache update at the start of setup (line 1243 here) and then assume the cache is up to date for the remainder of apt operations?

wyli: that will require `sudo`, even when all the deps are already installed properly...I hope to avoid sudo in this case. or do you see any requirements of keeping the apt cache up to date early?

tbirdso: Main concern is whether having `apt update` in each package upgrade call adds unreasonable overhead on first call, but maybe it is a reasonable tradeoff for first time vs subsequent runtimes.

Could you please compare the runtime for first time installation in a clean environment under the current impl, with the runtime for first time installation with only one `apt update` at the start, and report on the magnitude of difference? Maybe there are few enough install steps that the difference is trivial.

wyli: If the concern is about multiple `apt update` calls, this implementation of global caching variable would address it...

https://github.com/nvidia-holoscan/holohub/blob/6e9186c7f8e5fdb6b5da5d90f60442a4d3781f2a/utilities/cli/util.py#L405-L410

The first call (triggered by any missing package) will set `global _apt_updated` to True it'll stay true during the `./holohub setup`, the actual `apt update` won't be called again. So it'll be the same overhead as the main current branch. Maybe we can go for this version? otherwise I can take more time to think about it, it's a nice to have feature, the current main doesn't have any major issue in handling setup...

tbirdso: Right, so three competing designs:
1. Current (worst) where `apt update` runs eagerly each time we want to install a new package set
2. The design above (reverted in https://github.com/nvidia-holoscan/holohub/pull/973/commits/f036ce55da4fab4102367de0f9bd74c74582279d) where a global var tracks whether `apt update` has been run at all in this CLI session. Minor risk of going out of date if remote changes occur after the first `apt update` (rare)
3. The design I've suggested in this thread where `apt update` runs eagerly exactly once at the start of `setup`. Ensures `apt` cache is up to date before running any checks, but same risk of going out of date if remote changes occur. Risks unnecessary cache update as the initial package check with `dpkg` does not depend on the apt cache.

On comparison of above I've changed my mind and I'm in agreement on reverting to your original design in (2), thanks.

wyli: thanks for looking into these, agree with the analysis. in practice 2 and 3 don't have major difference from a user's perspective, mainly this PR add some logic to determine `sudo` based on the running environment.

utilities/cli/util.py
```diff
@@ -304,7 +304,7 @@ def get_host_gpu() -> str:
         )
         if not output or "Orin (nvgpu)" in output.decode():
             return "igpu"
-    except subprocess.CalledProcessError:
+    except Exception:
```
tbirdso: Seems like we could check for `SubprocessError`? https://docs.python.org/3/library/subprocess.html#exceptions

Not a blocker for merge

utilities/cli/util.py
```diff
@@ -232,7 +304,7 @@ def get_host_gpu() -> str:
...
```
wyli:
```suggestion
    except subprocess.CalledProcessError:
```

utilities/cli/util.py
```diff
@@ -87,10 +87,14 @@ def color_method(text: str, bold: bool = False) -> str:
...
 
 
+_sudo_available = None  # Cache for sudo availability check
+_apt_updated = False  # track whether apt update has been called
+
+
 # Utility Functions
 def get_timestamp() -> str:
     """Get current timestamp in the format used by the bash script"""
-    return datetime.utcnow().strftime("%Y-%m-%d %H:%M:%S")
```
wyli, line 93: for the record this is to address warning message:

```
datetime.utcnow()
<stdin>:1: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
datetime.datetime(2025, 7, 10, 11, 54, 1, 445500)
```

PR #977, opened 2025-07-07
Update Dockerfile to Avoid Unnecessary Warning

### LINKED ISSUE #784: Warning when building rivermax docker in advanced network operators
Building Rivermax docker image in operaters/advanced_network following the guide shows the following warning:
 1 warning found (use docker --debug to expand):
 - InvalidDefaultArgInFrom: Default value for ARG ${BASE_IMAGE} results in empty or invalid base image name (line 20)

bhashemian:
Fixes #784 

This PR set a reasonable default value for BASE_IMAGE in Dockerfile to avoid the following warning:
```
InvalidDefaultArgInFrom: Default value for ARG ${BASE_IMAGE} results in empty or invalid base image name
```

It also makes it easier for the user to build the Dockerfile without the Holohub CLI by providing the SDK version and GPU type while maintaining a backward compatibility.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,9 +20,9 @@
 # Base image
 ############################################################
 
-ARG BASE_IMAGE
 ARG GPU_TYPE
-
+ARG BASE_SDK_VERSION
+ARG BASE_IMAGE=nvcr.io/nvidia/clara-holoscan/holoscan:v${BASE_SDK_VERSION}-${GPU_TYPE}
 FROM ${BASE_IMAGE} AS base
 
 ARG DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -20,9 +20,9 @@
 # Base image
 ############################################################
 
-ARG BASE_IMAGE
 ARG GPU_TYPE
-
+ARG BASE_SDK_VERSION
```
tbirdso, line 24: Should we update the default container build call to pass a default version for `BASE_SDK_VERSION` in addition to the default value for `BASE_SDK_VERSION`? https://github.com/nvidia-holoscan/holohub/blob/main/utilities/cli/container.py#L366

bhashemian: @tbirdso do you mean providing `BASE_SDK_VERSION` in addition to `BASE_IMAGE`?
It doesn't affect any current Dockerfile since none of them are using `BASE_SDK_VERSION` when it is called from CLI but I can definitely add that.

bhashemian: The CLI now provides `BASE_SDK_VERSION` as a build argument.

PR #980, opened 2025-07-07
Build and run instructions for the high_speed_endoscopy application

sohamm17:


applications/high_speed_endoscopy/README.md
```diff
@@ -29,6 +29,39 @@ recommended hardware, GSYNC and RDMA enabled on exclusive display mode is 10ms o
...
 
+## Building and Running the Application
+
+This application cannot be compiled within the Holohub container. It needs to be compiled on the 
+host machine.
+
+### Set up the environment
+
+```bash
+export LD_LIBRARY_PATH=/usr/lib/aarch64-linux-gnu:$LD_LIBRARY_PATH
```
tbirdso: Typically handled by `/etc/ld.so.conf.d/aarch64-linux-gnu.conf` (at least it is on my IGX), can you confirm this is needed?

Can check:
```
ldconfig -p | grep /usr/lib/aarch64-linux-gnu   # should see at least one library listed
```

applications/high_speed_endoscopy/README.md
```diff
@@ -29,6 +29,39 @@ recommended hardware, GSYNC and RDMA enabled on exclusive display mode is 10ms o
...
+export LD_LIBRARY_PATH=/opt/nvidia/holoscan/lib:$LD_LIBRARY_PATH
```
tbirdso: Typically handled by RPATH in the output build, do you see the same?

For instance, I see:
```
holohub/build$ readelf -d ./endoscopy_tool_tracking/applications/endoscopy_tool_tracking/cpp/endoscopy_tool_tracking | grep PATH
 0x000000000000001d (RUNPATH)            Library runpath: [/opt/nvidia/holoscan/lib]
```

applications/high_speed_endoscopy/README.md
```diff
@@ -29,6 +29,39 @@ recommended hardware, GSYNC and RDMA enabled on exclusive display mode is 10ms o
...
+export LD_PRELOAD=/usr/lib/aarch64-linux-gnu/nvidia/libnvjpeg.so
```
tbirdso: Is there a known issue requiring this pre-load?

applications/high_speed_endoscopy/README.md
```diff
@@ -29,6 +29,39 @@ recommended hardware, GSYNC and RDMA enabled on exclusive display mode is 10ms o
...
+export LD_LIBRARY_PATH=/opt/EVT/eSDK:$LD_LIBRARY_PATH
+```
+
+### Compile the application
```
tbirdso, line 43: `./holohub build high_speed_endoscopy --local` ?

applications/high_speed_endoscopy/README.md
```diff
@@ -29,6 +29,39 @@ recommended hardware, GSYNC and RDMA enabled on exclusive display mode is 10ms o
...
+cd build/high_speed_endoscopy
+cmake ../.. -DAPP_high_speed_endoscopy=ON
+make -j
+```
+
+### Run the application
+
+```bash
+cd <HOLOHUB_SOURCE_DIR>
+sudo ./run launch high_speed_endoscopy
```
tbirdso:
```suggestion
sudo ./holohub run high_speed_endoscopy
```

applications/high_speed_endoscopy/README.md
anishag14: Yes. we have seen this multiple times over different applications that are run without the container.

applications/high_speed_endoscopy/README.md
anishag14: This command has not been tried. We are providing the commands that did work without the container.

applications/high_speed_endoscopy/README.md
anishag14: My system it was needed and that was the only way I could build emergent application.

applications/high_speed_endoscopy/README.md
anishag14: as mentioned below

sohamm17: I tried removing those paths from LD_LIBRARY_PATH, and it still works. SO, I'll push an update after removing this since it might have been fixed in some later releases.

applications/high_speed_endoscopy/README.md
sohamm17: Same as above - it looks like it could be removed

applications/high_speed_endoscopy/README.md
sohamm17: Looks like this could be removed as well.

applications/high_speed_endoscopy/README.md
sohamm17: I tried it now, and it works.

applications/high_speed_endoscopy/README.md
sohamm17: This command did not work. I think we need a way to run commands locally from the host machine.

applications/high_speed_endoscopy/README.md
tbirdso: >that was the only way I could build emergent application

To confirm, needed to build or to run? Would typically expect LD_LIBRARY_PATH is only needed when building

applications/high_speed_endoscopy/README.md
tbirdso: Thanks, what is the error seen without it? Is this tracked as a bug?

applications/high_speed_endoscopy/README.md
tbirdso: If this is required on your IGX, it suggest the IGX might have an issue with its setup and needs reflash..

applications/high_speed_endoscopy/README.md
tbirdso: Would it be possible to verify the updated command? This is a first-party app and developer experience of one versus six commands to build matters

tbirdso: `./holohub build ...` is doing effectively the same as the above under the hood, but a local Emergent SDK installation is required, so we need your help to validate the build command 🙏

sohamm17: @tbirdso, yes, the updated command works. However, I also kept the original `cmake` and `make` commands for documentation and future reference.

applications/high_speed_endoscopy/README.md
tbirdso: Ah, try adding the `--local` flag:

```bash
sudo ./holohub run high_speed_endoscopy --local
```

applications/high_speed_endoscopy/README.md
tbirdso: Thanks @sohamm17  for confirming!

applications/high_speed_endoscopy/README.md
sohamm17: Yes, I tried with `--local` but it still did not work.

tbirdso: Could you please share the error or behavior you're seeing with `--local`?

applications/high_speed_endoscopy/README.md
```diff
@@ -29,14 +29,72 @@ recommended hardware, GSYNC and RDMA enabled on exclusive display mode is 10ms o
...
 
 3. **Problem:** The application fails to connect to the EVT camera with error message “GVCP ack error”.
     - It could be an issue with the HR12 power connection to the camera. Disconnect the HR12 power connector from the camera and try reconnecting it.
+
+4. **Problem:** The applications fails with a segmentation fault at runtime, in the HoloViz visualization operator.
+    - It could be an issue with the configured BAR size on the GPU. Use the [display mode selector tool](https://developer.nvidia.com/displaymodeselector) to increase the BAR size.
+    - If you are using a lower resolution monitor, then you might need to decrease the resolution of the vistualization in the `high_speed_endoscopy.yaml` file.
+
+
+## Known Issues
```
tbirdso, line 97: These are concerning, could you please file internal bugs for investigation (if you haven't already)?

sohamm17:
> Thanks @sohamm17 for resolving the Python app issue. Please sign commits with `git commit --amend --no-edit -s` to resolve the DCO check failure.

Fixed it @tbirdso . Why is `lint-check` of an unrelated network operator showing up here: https://github.com/nvidia-holoscan/holohub/actions/runs/16197749456/job/45728580223

tbirdso:
> Why is lint-check of an unrelated network operator showing up here

Link-check* runs for all applications in the repo, you're correct that the failure here is unrelated and not a blocker for merging this contribution.

I'll open an issue to see about running link-check only on updated files in PRs.

Issue #979, opened 2025-07-07
entrypoint issue ./holohub run colonoscopy_segmentation


wyli:
customized base image may not have an entrypoint but have a `CMD` for example:
https://github.com/nvidia-holoscan/holohub/blob/fce33b3635cd3436f7536802b35e7c2a7442d383/applications/colonoscopy_segmentation/Dockerfile#L18

this causes issues when running application, e.g. `./holohub run colonoscopy_segmentation`

error:
```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: exec: "./holohub run colonoscopy_segmentation --language python --local": stat ./holohub run colonoscopy_segmentation --language python --local: no such file or directory: unknown
```

workaround (to override the entrypoint):
```
./holohub run colonoscopy_segmentation --docker-opts='--entrypoint=/bin/bash'
```

PR #982, opened 2025-07-08
enhances ./holohub run entrypoint

wyli:
- fixes https://github.com/nvidia-holoscan/holohub/issues/979
- reviewed the use cases:
  - custom entrypoint `./holohub run-container --docker-opt='--entrypoint=/bin/bash' -- echo test`
  - dockerfile entrypoint `./holohub run-container -- echo test`
  - tested various base image settings of entrypoint

### DIFF
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2, fixes-entrypoint]
+    branches: [main, holoscan-sdk-lws2]
 
 permissions:
   contents: read


.github/workflows/check_docker.yml
```diff
@@ -2,11 +2,19 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, fixes-entrypoint]
```
```suggestion
    branches: [main, holoscan-sdk-lws2]
```

.github/workflows/check_docker.yml
```diff
@@ -26,15 +34,42 @@ jobs:
...
+  test-entrypoint-images:
+    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - image: "alpine:latest"  # CMD ["/bin/sh"]
+          - image: "busybox:latest" # CMD ["sh"]
+          - image: "jenkins/jenkins:latest"    # ENTRYPOINT ["/usr/local/bin/jenkins.sh"]
+          - image: "nginx:latest"   # ENTRYPOINT ["/docker-entrypoint.sh"] CMD ["nginx", "-g", "daemon off;"]
```
tbirdso: Slight concern about relying on these for test coverage, we don't typically use these images and maintenance effort could be disproportionate if one of them decides to shift their entrypoint strategy upstream, even if unlikely.

Would it be difficult to write our own minimal Dockerfiles to cover these cases? Could either define in shell steps here or check in as test files.

utilities/cli/util.py
```diff
@@ -569,21 +569,31 @@ def get_entrypoint_command_args(
...
         except ValueError:
-            return "", [command]
-    entrypoint = get_container_entrypoint(img, dry_run=dry_run)
-    if not entrypoint:  # image has no entrypoint, docker uses default "/bin/sh -c" for command
-        return "", [command]
+            pass
+    entrypoint = [entrypoint] if entrypoint else get_container_entrypoint(img, dry_run=dry_run)
+
+    if not entrypoint:  # image has no entrypoint, use default "/bin/sh -c"
+        return "", ["/bin/sh", "-c", command]
```
tbirdso: Is there a functional difference between `--entrypoint="/bin/sh" -c <command>` and `/bin/sh -c <command>`? In both cases `/bin/sh` is the first script to run with args `-c <command>`

Why `/bin/sh` over `/bin/bash`?

wyli: thanks, I'll update to `/bin/bash` to be consistent

.github/workflows/check_docker.yml
wyli: updated to use basic dockerfile to simulate the use cases

PR #984, opened 2025-07-09
./holohub lint in a container

wyli:
this adds support `./holohub lint` in a container
- include `.local/bin` path for the pip installed bins during linting
- change RUFF_CACHE_DIR, BLACK_CACHE_DIR to avoid permission issue when in a container
- use pip version of clang-format instead of the apt install to avoid the `sudo` requirement

### DIFF
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, lint-docker]
 
 permissions:
   contents: read
@@ -42,6 +42,9 @@ jobs:
           ./holohub run-container --no-docker-build --add-volume "/tmp" | grep -q "/tmp" || { echo "add-volume test failed"; exit 1; }
           ./holohub run-container --no-docker-build -- echo hello > /tmp/trailing-args.log 2>&1
           grep -q "hello" /tmp/trailing-args.log || { echo "trailing args test failed"; cat /tmp/trailing-args.log; exit 1; }
+
+          # test linting in docker
+          ./holohub run-container --base-img ubuntu:${ubuntu_version} --no-docker-build -- "./holohub lint --install-dep; ./holohub lint"
 
   test-entrypoint:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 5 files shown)


utilities/cli/holohub.py
```diff
@@ -1073,14 +1092,20 @@ def handle_lint(self, args: argparse.Namespace) -> None:
...
+                    "--exclude",
+                    ".ruff_cache",
+                    "--exclude",
+                    ".local",
                     "--recursive",
                     args.path,
                 ]
                 if args.dryrun:
-                    holohub_cli_util.run_command(cmd, dry_run=True)
+                    holohub_cli_util.run_command(cmd, dry_run=True, env=env)
```
wyli:
```suggestion
                    holohub_cli_util.run_command(cmd, dry_run=True)
```

.github/workflows/check_docker.yml
```diff
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, lint-docker]
```
```suggestion
    branches: [main, holoscan-sdk-lws2]
```

utilities/cli/holohub.py
```diff
@@ -1024,6 +1024,19 @@ def handle_lint(self, args: argparse.Namespace) -> None:
 
         exit_code = 0
 
+        # Add ~/.local/bin to PATH for pip-installed executables
```
tbirdso, line 1027: Concerned that this might produce unexpected behavior depending on what the user has in their host `holohub/.local` folder at runtime? What do you think?

.github/workflows/check_docker.yml
```diff
@@ -43,6 +43,9 @@ jobs:
           ./holohub run-container --no-docker-build -- echo hello > /tmp/trailing-args.log 2>&1
           grep -q "hello" /tmp/trailing-args.log || { echo "trailing args test failed"; cat /tmp/trailing-args.log; exit 1; }
 
+          # test linting in docker
+          ./holohub run-container --base-img ubuntu:${ubuntu_version} --no-docker-build -- "./holohub lint --install-dep; ./holohub lint --fix; ./holohub lint"
```
tbirdso: Preference on keeping the full arg name versus shortened?

```suggestion
          ./holohub run-container --base-img ubuntu:${ubuntu_version} --no-docker-build -- "./holohub lint --install-dependencies; ./holohub lint --fix; ./holohub lint"
```

utilities/cli/holohub.py
wyli: this is a temporary variable as context for `run_command` during `handle_lint`, it doesn't change the system env, I think it should be ok

tbirdso: Good point!

tbirdso (approved):
LGTM, thank you!

PR #1002, opened 2025-07-11
premerge ci add copyright header/nvstaging url checks

wyli:


.github/workflows/scripts/check_copyright.py
```diff
@@ -0,0 +1,308 @@
+"""
```
wyli, line 1: script adapted from holoscan-sdk gitlab scripts

.github/workflows/check_compliance.yml
```diff
@@ -0,0 +1,35 @@
+name: Check Compliance
+
+on:
+  push:
+    branches: [main, temp-test]
```
wyli:
```suggestion
    branches: [main]
```

.github/workflows/check_compliance.yml
```diff
@@ -0,0 +1,35 @@
...
+        fetch-depth: 0
+
+    - name: Check copyright headers (PR - changed files only)
+      if: github.event_name == 'pull_request'
+      run: |
+        echo "Checking copyright headers in changed files..."
+        python3 .github/workflows/scripts/check_copyright.py --git-modified-only origin/${{ github.base_ref }} \
+          --exclude-config .github/workflows/scripts/copyright_excludes.txt
+
+    - name: Check copyright headers (all files)
```
tbirdso: Can you confirm whether we are checking copyright headers for all files in Holoscan SDK or just modified files? Not sure we should claim copyright year updates in unmodified files. Also, this might get noisy each January 1st.

wyli: confirm checking
- premerge is to check the modified
(https://github.com/nvidia-holoscan/holohub/actions/runs/16223099226/job/45808338177)
- all files on main branch when pushing to main (https://github.com/wyli/holohub/actions/runs/16223098089/job/45808333721)

looks like the second is not needed, it's only useful for this PR (update: removed the second regular run as it's not needed).

the script has an `--update-current-year` option can easily update the header automatically

wyli:
is it correct to use `nvstaging` resource here? https://github.com/nvidia-holoscan/holohub/blob/fbde0077cda97bea6eb02c60b723e94d9e7cc68a/applications/polyp_detection/README.md?plain=1#L26

tbirdso:
> is it correct to use nvstaging resource here?

No, should be updated to public `nvidia` URL

wyli:
> is it correct to use `nvstaging` resource here?
> 
> https://github.com/nvidia-holoscan/holohub/blob/fbde0077cda97bea6eb02c60b723e94d9e7cc68a/applications/polyp_detection/README.md?plain=1#L26

Hi @yiheng-wang-nv this nvstaging URL looks temporary, is there an update on this?

tbirdso:
> Hi @yiheng-wang-nv this nvstaging URL looks temporary, is there an update on this?

This is probably the public version? Published 2025.06.12 https://catalog.ngc.nvidia.com/orgs/nvidia/teams/clara-holoscan/models/polyp_detection_rt_detr_model

wyli:
> 
> This is probably the public version? Published 2025.06.12 https://catalog.ngc.nvidia.com/orgs/nvidia/teams/clara-holoscan/models/polyp_detection_rt_detr_model

thanks, updated to use the public version  in `applications/polyp_detection/README.md` instead of nvstaging cc @yiheng-wang-nv

.devcontainer/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-# SPDX-FileCopyrightText: Copyright (c) 2024 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
tbirdso: I don't think we should update copyright headers unless we are making material changes to the file in the same PR. Can we remove these from this PR?

.github/workflows/check_compliance.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Check copyright headers (PR - changed files only)
+      if: github.event_name == 'pull_request'
+      run: |
+        echo "Checking copyright headers in changed files..."
+        python3 .github/workflows/scripts/check_copyright.py --git-modified-only origin/${{ github.base_ref }} \
```
tbirdso, line 20: Do you think it would be valuable to add HoloHub CLI support for these checks, either as part of `./holohub lint` or as a standalone `./holohub check_copyright` command? Not a blocker

wyli: for now I think this is a repo-specific check. we can make it more generic (e.g. differet license types, formatting style) and available in the CLI if there are user needs in the future

.devcontainer/Dockerfile
wyli: thanks, this makes sense, I added an option `--ignore-year-mismatch`. not sure why this option was not available in the original script and it always enforced the current year.

operators/advanced_network/advanced_network/managers/rivermax/rmax_service/ipo_receiver_io_node.cpp
```diff
@@ -1,13 +1,18 @@
 /*
- * Copyright (c) 2017-2024 NVIDIA CORPORATION & AFFILIATES. ALL RIGHTS RESERVED.
```
wyli, line 2: the main branch's license looks like a typo `ipo_receiver_io_node.cpp` comparing other files in the same folder

tbirdso: Seems likely, but we should confirm with @agirault @cliffburdick  before changing ANO copyrights

tbirdso: Could move to a separate PR if you'd rather avoid blocking these changes on ANO feedback

applications/dds/dds_video/bootstrap.sh
```diff
@@ -1,4 +1,18 @@
 #!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights
```
tbirdso, line 2: Thanks for adding missing copyright headers, looks great. For understanding, did the script catch these missing headers, and were they fixed automatically or manually?

.github/workflows/scripts/copyright_excludes.txt
```diff
@@ -0,0 +1,28 @@
+# Copyright check exclude patterns
```
tbirdso, line 1: Could you please comment on why we are excluding these folders? Are these mostly third party folders with non-Apache 2.0 licenses?

tbirdso (changes requested):
Mostly looks good, but blocked on approval to change ANO file license

operators/advanced_network/advanced_network/managers/rivermax/rmax_service/ipo_receiver_io_node.cpp
ronyrad: This file will be removed anyway when Rivermax PR will be merged. I believe you can change it meanwhile as above

.github/workflows/scripts/copyright_excludes.txt
wyli: Yes, 3rd party copyright headers in these folders

applications/dds/dds_video/bootstrap.sh
wyli: Yes, the script will try to detect missing ones, it can only automatically fix the minor outdated year info. Overall format still need manual/GenAI inputs

tbirdso: Sounds good!

operators/advanced_network/advanced_network/managers/rivermax/rmax_service/ipo_receiver_io_node.cpp
tbirdso: Thanks for confirming @ronyrad

PR #1012, opened 2025-07-16
review vscode tasks

wyli:


.vscode/tasks.json
```diff
@@ -272,32 +272,6 @@
...
-            "presentation": {
-                "echo": true,
-                "reveal": "silent",
-                "focus": true,
-                "panel": "dedicated",
-                "showReuseMessage": false,
-                "clear": true,
-                "group": "grpc_endoscopy_tool_tracking"
-            }
-        },
```
bhashemian, line 300: May I know what is the reason to remove `grpc_endoscopy_tool_tracking`?

bhashemian (commented):
Thanks @wyli for reviewing and updating vscode tasks. Overall looks good to me but I am wondering if we should also update the the two benchmarking tasks to use the new CLI `./holohub build <app> --benchmark`. Thanks!

.vscode/tasks.json
```diff
@@ -41,11 +41,11 @@
         {
             "type": "shell",
             "label": "Build basic_networking_ping",
-            "command": "./run",
+            "command": "./holohub",
             "args": [
                 "build",
```
tbirdso, line 46: Just to confirm, does the dev container know it is in a HoloHub container environment or will this try to build the app container first?

.vscode/tasks.json
```diff
@@ -28,7 +28,7 @@
         {
             "type": "shell",
             "label": "Clear Build Cache",
-            "command": "./run",
+            "command": "./holohub",
             "args": [
                 "clear_cache"
```
tbirdso:
```suggestion
                "clear-cache"
```

.vscode/tasks.json
wyli: there's a duplicated item around line 370

.vscode/tasks.json
wyli: I don't start a container when I open vscode, so this command will start a container first. is this expected?

wyli:
> update the the two benchmarking tasks to use the new CLI `./holohub build <app> --benchmark`. Thanks!

Hi @bhashemian  do you mean these lines? https://github.com/nvidia-holoscan/holohub/blob/86ea79e9ac46dc2d8fcf1923edb2ca76eba29595/.vscode/tasks.json#L538-L567, I think these are up to date in this PR

.vscode/tasks.json
tbirdso: `--local` would be consistent with the `run` script and ensure that running tasks in the dev container works, but @mocsharp might be the best reference to validate VS Code integration here

ghost: Yes, we don't want to start a container for building, using `--local` is required for all builds.

wyli: thanks for confirming!

tbirdso: Confirming my understanding -- are the tasks here always expected to run inside a VS Code dev container based on the given app's build Dockerfile, so that all of the dependencies are already available?

If the user's VS Code environment is the host rather than dev container, can we always assume that the user is intending to build on bare metal (host)? Or is there some scenario where we do want to support containerized build from tasks.json, whether now or as a future enhancement?

If we are not targeting building on the host machine, would it make sense to set `HOLOHUB_BUILD_LOCAL` in the VS Code dev container environment in place of the `--local` flag?

Not a blocker for this PR, just want to understand better how tasks.json is being used.

wyli: We can add tasks for both local and docker if that's easier to use..

ghost: @tbirdso The build tasks are launched as part of a debug profile. They are not intended to launched independently.

That means the builds always have to use `--local` no matter if VSCode is using the dev container or running bare metal. Because the built code is intended to be called by the debugger.

tbirdso: Great, thanks @AndreasHeumann for the explanation. No further concerns 👍

PR #1016, opened 2025-07-16
fixes ./holohub test

wyli:
- ./holohub test `--img` was not used in `container.run`
- the ctest exit code is 0 (should be non-zero) when the test cases fail


tbirdso (approved):
Good catch!

PR #1017, opened 2025-07-17
Enhanced Slang shader support

AndreasHeumann:
Could not launch kernels from kernels with slang so implemented two
kernels. Nevertheless, performance is actually even slightly better
than the non-Slang implementation.

Add required features to SlangShaderOp and add tests.

### DIFF
--- a/applications/endoscopy_tool_tracking/cpp/main.cpp
+++ b/applications/endoscopy_tool_tracking/cpp/main.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2024 NVIDIA CORPORATION & AFFILIATES. All rights
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2025 NVIDIA CORPORATION & AFFILIATES. All rights
  * reserved. SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -23,6 +23,8 @@
 #include <holoscan/operators/video_stream_recorder/video_stream_recorder.hpp>
 #include <holoscan/operators/video_stream_replayer/video_stream_replayer.hpp>
 #include <lstm_tensor_rt_inference.hpp>
+#include <slang_shader_op.hpp>
+#include <string>
 #include <tool_tracking_postprocessor.hpp>
 #ifdef VTK_RENDERER
 #include <vtk_renderer.hpp>
@@ -48,7 +50,9 @@
 
 class App : public holoscan::Application {
  public:
+  void set_path(const std::string& path) { app_path_ = path; }
   void set_source(const std::string& source) { source_ = source; }
+  void set_postprocessor(const std::string& postprocessor) { this->postprocessor_ = postprocessor; }
   void set_visualizer_name(const std::string& visualizer_name) {
     this->visualizer_name = visualizer_name;
   }
@@ -94,7 +98,8 @@ class App : public holoscan::Application {
       source = make_operator<ops::AJASourceOp>(
           "aja", from_config("aja"), from_config("external_source"));
 #else
-throw std::runtime_error("AJA is requested but not available. Please enable AJA at build time.");
+      throw std::runtime_error(
+          "AJA is requested but not available. Please enable AJA at build time.");
 #endif
       source_block_size = width * height * 4 * 4;
       source_num_blocks = use_rdma ? 3 : 4;
@@ -178,21 +183,31 @@ throw std::runtime_error("AJA is requested but not available. Please enable AJA 
     const uint64_t tool_tracking_postprocessor_block_size =
         std::max(107 * 60 * 7 * 4 * sizeof(float), 7 * 3 * sizeof(float));
     const uint64_t tool_tracking_postprocessor_num_blocks = 2 * 2;
-    auto tool_tracking_postprocessor = make_operator<ops::ToolTrackingPostprocessorOp>(
-        "tool_tracking_postprocessor",
-        Arg("device_allocator") =
-            make_resource<BlockMemoryPool>("device_allocator",
-                                           1,
-                                           tool_tracking_postprocessor_block_size,
-                                           tool_tracking_postprocessor_num_blocks));
-
+    std::shared_ptr<Operator> tool_tracking_postprocessor;
+    std::shared_ptr<BlockMemoryPool> postprocessor_allocator =
+        make_resource<BlockMemoryPool>("device_allocator",
+                                       1,
+                                       tool_tracking_postprocessor_block_size,
+                                       tool_tracking_postprocessor_num_blocks);
+    if (postprocessor_ == "slang_shader") {
+      tool_tracking_postprocessor = make_operator<ops::SlangShaderOp>(
+          "slang_postprocessor",
+          Arg("shader_source_file", app_path_ + "/postprocessor.slang"),
+          Arg("allocator") = postprocessor_allocator);
+    } else if (postprocessor_ == "tool_tracking_postprocessor") {
+      tool_tracking_postprocessor = make_operator<ops::ToolTrackingPostprocessorOp>(
+          "tool_tracking_postprocessor",
+          Arg("device_allocator") = postprocessor_allocator);
+    } else {
+      throw std::runtime_error("Invalid postprocessor: " + postprocessor_);
+    }
     if (this->visualizer_name == "holoviz") {
       std::shared_ptr<BlockMemoryPool> visualizer_allocator;
       if (((record_type_ == Record::VISUALIZER) && (source_ == "replayer"))
 #ifdef DELTACAST_VIDEOMASTER
           || overlay_enabled
 #endif
-    ) {
+      ) {
         visualizer_allocator =
             make_resource<BlockMemoryPool>("allocator", 1, source_block_size, source_num_blocks);
       }
@@ -297,7 +312,9 @@ throw std::runtime_error("AJA is requested but not available. Please enable AJA 
   }
 
  private:
+  std::string app_path_ = "";
   std::string source_ = "replayer";
+  std::string postprocessor_ = "tool_tracking_postprocessor";
   std::string visualizer_name = "holoviz";
   Record record_type_ = Record::NONE;
   std::string datapath = "";
@@ -309,7 +326,8 @@ bool parse_arguments(int argc, char** argv, std::string& data_path, std::string&
       {"data", required_argument, 0, 'd'}, {"config", required_argument, 0, 'c'}, {0, 0, 0, 0}};
 
   while (int c = getopt_long(argc, argv, "d:c:", long_options, NULL)) {
-    if (c == -1 || c == '?') break;
+    if (c == -1 || c == '?')
+      break;
 
     switch (c) {
       case 'c':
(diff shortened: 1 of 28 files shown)


applications/endoscopy_tool_tracking/cpp/main.cpp
```diff
@@ -178,21 +183,34 @@ throw std::runtime_error("AJA is requested but not available. Please enable AJA
...
+      tool_tracking_postprocessor = make_operator<ops::ToolTrackingPostprocessorOp>(
+          "tool_tracking_postprocessor",
+          Arg("device_allocator") =
+              make_resource<BlockMemoryPool>("device_allocator",
+                                             1,
+                                             tool_tracking_postprocessor_block_size,
+                                             tool_tracking_postprocessor_num_blocks));
+    } else {
+      throw std::runtime_error("Invalid postprocessor: " + postprocessor_);
+    }
```
grlee77: can define the same allocator outside of the if/else

```suggestion
    auto postprocessor_allocator =
        make_resource<BlockMemoryPool>("device_allocator",
                                       1,
                                       tool_tracking_postprocessor_block_size,
                                       tool_tracking_postprocessor_num_blocks);
    if (postprocessor_ == "slang_shader") {
      tool_tracking_postprocessor = make_operator<ops::SlangShaderOp>(
          "slang_postprocessor",
          Arg("shader_source_file", app_path_ + "/postprocessor.slang"),
          Arg("allocator") = postprocessor_allocator);
    } else if (postprocessor_ == "tool_tracking_postprocessor") {
      tool_tracking_postprocessor = make_operator<ops::ToolTrackingPostprocessorOp>(
          "tool_tracking_postprocessor", Arg("device_allocator") = postprocessor_allocator);
    } else {
      throw std::runtime_error("Invalid postprocessor: " + postprocessor_);
    }
```

applications/endoscopy_tool_tracking/python/endoscopy_tool_tracking.py
```diff
@@ -239,17 +244,33 @@ def compose(self):
             107 * 60 * 7 * 4 * bytes_per_float32, 7 * 3 * bytes_per_float32
         )
         tool_tracking_postprocessor_num_blocks = 2 * 2
-        tool_tracking_postprocessor = ToolTrackingPostprocessorOp(
-            self,
-            name="tool_tracking_postprocessor",
-            device_allocator=BlockMemoryPool(
+        if self.postprocessor == "slang_shader":
```
grlee77, line 254: can also reuse the allocator in both code paths here (couldn't create the suggestion via the UI in this case)

```py
        post_processor_allocator = BlockMemoryPool(
            self,
            name="device_allocator",
            storage_type=MemoryStorageType.DEVICE,
            block_size=tool_tracking_postprocessor_block_size,
            num_blocks=tool_tracking_postprocessor_num_blocks,
        ),

        if self.postprocessor == "slang_shader":
            tool_tracking_postprocessor = SlangShaderOp(
                self,
                name="slang_postprocessor",
                shader_source_file=os.path.join(os.path.dirname(__file__), "postprocessor.slang"),
                allocator=post_processor_allocator,
            )
        elif self.postprocessor == "tool_tracking_postprocessor":
            tool_tracking_postprocessor = ToolTrackingPostprocessorOp(
                self,
                name="tool_tracking_postprocessor",
                device_allocator=post_processor_allocator,
            )
        else:
            raise ValueError(f"Invalid postprocessor: {self.postprocessor}")
```

applications/endoscopy_tool_tracking/python/postprocessor.slang
grlee77, line 1: This appears to be identical to the one in the cpp folder. Can we use a symlink or a copy at build time to avoid duplicate source files?

operators/slang_shader/command.cpp
```diff
@@ -19,96 +19,407 @@
...
+
+      const uint64_t element_size = nvidia::gxf::PrimitiveTypeSize(maybe_primitive_type.value());
+
+      shape =
+          nvidia::gxf::Shape({static_cast<int32_t>(video_frame_info.height),
+                              static_cast<int32_t>(video_frame_info.width),
+                              static_cast<int32_t>(color_plane.bytes_per_pixel / element_size)});
+
+      strides.resize(shape.rank());
+      strides[0] = static_cast<uint64_t>(color_plane.bytes_per_pixel) * element_size;
```
grlee77: Can you double check this stride?

Isn't `color_plane.bytes_per_pixel` defined as the number of channels (e.g. 4 for RGBA) times the element size for a single channel? Why would you also multiply again by `element_size`?

When setting the tensor shape you divided `color_plane.bytes_per_pixel / element_size` to get the number of channels. In that case, it seems the stride should just be for a single element (`element_size`)?


```suggestion
      strides[0] = element_size;
```

operators/slang_shader/command.cpp
```diff
@@ -19,96 +19,407 @@
...
+
+/**
+ * Swizzles a shape based on a string. The string is a sequence of characters that specify the
+ * dimensions of the shape. The characters are:
+ * - 'c': the component count
+ * - 'x': the first dimension
+ * - 'y': the second dimension
+ * - 'z': the third dimension
+ * - 'w': the fourth dimension
+ * - '0-9': the dimension number
```
grlee77, line 62: Was the channel layout/naming convention documented anywhere in the README for the operators and/or the examples?

operators/slang_shader/command.cpp
```diff
@@ -119,101 +430,156 @@ void CommandParameter::execute(CommandWorkspace& workspace) {
...
-  param_data[2] = shape.dimension(2);
+  memcpy(workspace.shader_parameters_.data() + parameter_offset_, &strides, sizeof(ulonglong3));
 }
 
 void CommandReceiveCudaStream::execute(CommandWorkspace& workspace) {
-  workspace.cuda_stream_ = workspace.op_input_.receive_cuda_stream();
+  // If there are inputs, we receive the CUDA stream from the input port, else we allocate a new
+  // stream
+  if (workspace.op_input_.inputs().size() > 0) {
+    workspace.cuda_stream_ = workspace.op_input_.receive_cuda_stream();
```
grlee77: This code path will only work if `inputs().size() == 1`, otherwise because no port name argument was passed to `receive_cuda_stream` it will raise an exception if there is more than one input port.

It seems this function needs to throw an error if inputs().size() > 1 or pass a port name to `receive_cuda_stream`


Separately, the comment above indicates that this returns the stream from the input port, but that is not typically the actual behavior. The design (as with the older `CudaStreamHandler` you had implemented) is to synchronize any input stream with the operator's internal stream and then return that internal stream. Although, I think if there is no `CudaStreamPool` for the operator then it does return the stream found on the input.

As of Holoscan v3.5 the only case where the operator will not have a `CudaStreamPool` present is if it had an explicit "cuda_stream_pool" parameter (`Parameter<std::shared_ptr<CudaStreamPool>> cuda_stream_pool_;`), but no value was set for the parameter. I am wondering if we should reconsider that decision, though, as this potentially confusing / hard to explain. It may be simpler if we update the SDK to just **always** have a stream pool available by default even in the case an explicit parameter is present, but not sure if that would be a breaking change for any existing apps?

operators/slang_shader/test/test_slang_shader_op.cpp
```diff
@@ -0,0 +1,756 @@
...
+        printf("%d %d %d\n", invocations.x, invocations.y, invocations.z);
+      }
+    }
+  )r";
+
+  auto application = holoscan::make_application<SlangShaderApp>(shader_source);
+  application->add_video_buffer_source_op();
+  application->add_sink_op();
+
+  // 757869354 is the value of the first pixel in the input video buffer (42, 43, 44, 45)
```
grlee77, line 817: I don't follow the comment. Where does 757869354 come from? (It seems to be a small video buffer where the initial value is 42)

operators/slang_shader/test/test_slang_shader_op.cpp
grlee77, line 1: Great to see this test coverage for the operator code here!

applications/endoscopy_tool_tracking/cpp/main.cpp
ghost: Good suggestion, done.

applications/endoscopy_tool_tracking/python/endoscopy_tool_tracking.py
ghost: Done

applications/endoscopy_tool_tracking/python/postprocessor.slang
ghost: Yes, it's identical, will add a symlink.

operators/slang_shader/command.cpp
ghost: Not yet, it was here and in the holoscan.slang file. I added a section to the README.

operators/slang_shader/command.cpp
ghost: You are right, this is wrong. But I think it should be just color_plane.bytes_per_pixel so that stride[0] gets you to the next item in x direction. Similar as stride[1] gets you to the next line in y direction.
I think my test is also wrong.

operators/slang_shader/command.cpp
ghost: You are right, somehow I thought that when passing no port name it will get the streams of all ports, synchronize with the allocated stream and returns it. This is clearly not the case. I'll change that code and add a test with multiple inputs.
Thanks for catching this.

Regarding the `cuda_stream_pool` parameter. I think it's a good suggestion to always have a stream pool by default. I'm also not sure if it will break existing apps, but I don't think so.

operators/slang_shader/test/test_slang_shader_op.cpp
ghost: This is the pixel containing [42,43,44,45] converted to a int (since SinkOp prints a int): '(45 << 24) + (44 << 16) + (43 << 8) + 42'

AndreasHeumann (commented):
Thanks for the review @grlee77 !

operators/slang_shader/command.cpp
ghost: Actually it's a bit different. Since the shape is in HWC I decided to organize the stride array in a similar way. So stride[2] is the component stride, stride[1] the line stride and stride[2] the image stride.

operators/slang_shader/README.md
```diff
@@ -71,19 +71,54 @@ The SlangShaderOp uses special attributes to define how shader parameters intera
...
+- **Allowed characters**: `"x"`, `"y"`, `"z"`, `"c"`, `"0"` - `"9"`
+  - `"x"`, `"y"`, `"z"`: Select specific dimensions
+  - `"c"`: Component count
+  - `"0"` - `"9"`: Static values
+
+#### Examples
+```slang
+[holoscan::input("input_data")]           // Binds to a resource named "input_data"
+[holoscan::output("model:output")]        // Binds to the "output" tensor in the "model" tensor map
+[holoscan::alloc::size_of("input_tensor.cx")]  // Allocates based on the x dimension of "input_tensor"
```
grlee77:
```suggestion
[holoscan::alloc::size_of("input_tensor.cx")]  // Allocates based on the x dimension and component count of "input_tensor"
```

operators/slang_shader/test/test_slang_shader_op.cpp
grlee77: I see now, thanks for clarifying

operators/slang_shader/command.cpp
grlee77: Looks good now, thanks

grlee77 (approved):
Thanks for making the suggested updates. Approved, with one comment regarding the README

operators/slang_shader/README.md
ghost: Thanks, changed.

PR #1019, opened 2025-07-17
Arbitrary byte matching with flex parser

cliffburdick:
Allows users to match any arbitrary 32b word after a UDP header up to 8 bytes away. Both a mask and value are used to allow any bits/bytes to be matched in any order. A UDP header is only required for this PR, but in the future this can be more flexible. An example config in the RX section is as follows:

```
        flex_items:
        - name: "flex_item_0"
          id: 0
          offset: 16
          udp_dst_port: 4096
        flows:
        - name: "flow_0"
          id: 0
          action:
            type: queue
            id: 0
          match:
            flex_item_id: 0
            val: 0x00000000
            mask: 0x03000000
```

The `flex_items` section specifies a new flexible parsing match based on a UDP destination port and byte offset after the UDP header. The matches in each flow section allow a 32b value and mask to be computed, and if a match hits, the action for the flow is taken.


agirault (approved):
LGTM

bhashemian:
Hi @cliffburdick, would you be able to resolve the conflicts for this PR and update it with the latest changes on `main`? Thanks

PR #1020, opened 2025-07-18
App: Update H.264 script for GXF 5.0 / GXF MM 4.1.1 extensions

tbirdso:
Update to fetch GXF MM dependencies from NGC:
https://catalog.ngc.nvidia.com/orgs/nvidia/teams/graph-composer/resources/videodecoderextension/version

GXF MM 4.1.1 supports GXF 5.0 / Ubuntu >= 24.04 (glibc >= 2.39).

Changes:
- Update to use latest Holoscan SDK >= 3.3.0 dev container (based on Ubuntu 24.04)
- Update to fetch latest GXF MM extensions
- Script updates to 1) exit container build with failure if GXF MM installation fails, 2) log installation for easier debugging

**Needs attention**: h264_video_decode Python test fails with default run command:
```
[warning] [type_registry.cpp:57] Unknown type: nvidia::gxf::VideoReadBitStream
[error] [gxf_codelet.cpp:44] Unable to find the GXF type name ('nvidia::gxf::VideoReadBitStream') for the codelet
```
However, the symbol is defined in `/opt/nvidia/holoscan/lib/libgxf_core.so` and running the test with LD_PRELOAD succeeds:
```
LD_PRELOAD=/opt/nvidia/holoscan/lib/libgxf_core.so xvfb-run -a ctest -C Release --output-on-failure -R python -VV
...
The following tests passed:
        h264_video_decode_python_test

100% tests passed, 0 tests failed out of 1
```

### DIFF
--- a/applications/h264/install_dependencies.sh
+++ b/applications/h264/install_dependencies.sh
@@ -17,6 +17,7 @@
 
 # This scripts installs gxf multimedia libraries and corresponding dependencies
 # required for H264 Encode / Decode applications.
+set -e
 
 # URL of the DeepStream dependencies required for gxf-mm extensions above
 DS_DEPS_URL="https://api.ngc.nvidia.com/v2/resources/org/nvidia/gxf_and_gc/4.0.0/files?redirect=true&path=nvv4l2_x86_ds-7.0.deb"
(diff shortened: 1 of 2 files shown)


applications/h264/install_dependencies.sh
```diff
@@ -17,6 +17,7 @@
 
 # This scripts installs gxf multimedia libraries and corresponding dependencies
 # required for H264 Encode / Decode applications.
+set -e
 
 # URL of the DeepStream dependencies required for gxf-mm extensions above
 DS_DEPS_URL="https://api.ngc.nvidia.com/v2/resources/org/nvidia/gxf_and_gc/4.0.0/files?redirect=true&path=nvv4l2_x86_ds-7.0.deb"
```
apatole, line 23: @tbirdso Thanks for the patch! Just confirming, do we need to update this (DS_DEPS_URL="https://api.ngc.nvidia.com/v2/resources/org/nvidia/gxf_and_gc/4.0.0/files?redirect=true&path=nvv4l2_x86_ds-7.0.deb"
)  as well?

tbirdso: I did not see any issues that seemed specific to the DeepStream dependencies, so it seems likely they are compatible

Update would not hurt but probably not a blocker for investigating the reported symbol issue

tbirdso: GXF MM team confirms not a blocker, let's unblock for now and update when DS 8.0 dependencies are available

PR #1021, opened 2025-07-22
CLI add local-sdk-root path

wyli:
the build directory of local sdk by default would be something like `/workspace/holoscan-sdk/build-x86_64/` according to https://github.com/nvidia-holoscan/holoscan-sdk/blob/9c5b3c3d4831f2e65ebda6b79ae9b1c5517c6a7c/run#L226-L228

this PR revise the hard-coded `HOLOSCAN_LIB_PATH=/workspace/holoscan-sdk/build/lib` to allow for the build folder name variants.

### DIFF
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2, fixes-sdk-path]
+    branches: [main, holoscan-sdk-lws2]
 
 permissions:
   contents: read
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, fixes-sdk-path]
 
 permissions:
   contents: read
--- a/doc/developer.md
+++ b/doc/developer.md
@@ -127,7 +127,7 @@ $ ./holohub run-container --img holohub:local-sdk-latest --local-sdk-root <path_
 where `<path_to_holoscan_sdk>` is the path to the Holoscan SDK root directory containing the build directory.
 Please refer to the [Holoscan SDK Developer Guide](https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/DEVELOP.md) for more details on how to build the Holoscan SDK from source.
 
-In the container, to verify the build directory is mounted correctly, run the following command:
+In the container, to verify the build directory (with a python build) is mounted correctly, run the following command:
 ```bash
 $ python -c "import holoscan; print(holoscan.__file__)"
 ```
@@ -135,6 +135,9 @@ The output should be something like:
 ```bash
 /workspace/holoscan-sdk/build-x86_64/python/lib/holoscan/__init__.py
 ```
+
+If Python supported is not enabled, `/workspace/holoscan-sdk` can be manually inspected to confirm the mount.
+The directory should contain a non-empty `build-<arch>-<gpu_type>` or `install-<arch>-<gpu_type>` directory.
 
 ### Launch a Named HoloHub Container
 
--- a/doc/developer.md
+++ b/doc/developer.md
@@ -127,6 +127,15 @@ $ ./holohub run-container --img holohub:local-sdk-latest --local-sdk-root <path_
 where `<path_to_holoscan_sdk>` is the path to the Holoscan SDK root directory containing the build directory.
 Please refer to the [Holoscan SDK Developer Guide](https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/DEVELOP.md) for more details on how to build the Holoscan SDK from source.
 
+In the container, to verify the build directory is mounted correctly, run the following command:
+```bash
+$ python -c "import holoscan; print(holoscan.__file__)"
+```
+The output should be something like:
+```bash
+/workspace/holoscan-sdk/build-x86_64/python/lib/holoscan/__init__.py
+```
+
 ### Launch a Named HoloHub Container
 
 To launch custom HoloHub container with fully qualified name, e.g. "holohub:ngc-sdk-sample-app"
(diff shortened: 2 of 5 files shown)


utilities/cli/util.py
```diff
@@ -310,6 +310,53 @@ def get_host_gpu() -> str:
     return "dgpu"
 
 
+def get_host_arch() -> str:
+    """Get host architecture"""
+    machine = platform.machine()
+    if machine in ["x86_64", "amd64"]:
+        return "x86_64"
+    if machine in ["aarch64", "arm64"]:
```
tbirdso: Should we control for casing?
```suggestion
    machine = platform.machine().lower()
    if machine in ["x86_64", "amd64"]:
        return "x86_64"
    if machine in ["aarch64", "arm64"]:
```

utilities/cli/util.py
```diff
@@ -310,6 +310,53 @@ def get_host_gpu() -> str:
...
+def get_arch_gpu_str() -> str:
+    """Get architecture+GPU string like bash get_arch+gpu_str()"""
+    arch = get_host_arch()
+    if arch == "aarch64":
+        gpu = get_host_gpu()
+        return f"{arch}-{gpu}"
+    return arch
+
+
+def find_build_dir(local_sdk_root: Optional[Union[str, Path]] = None) -> str:
```
tbirdso: Suggest clarifying SDK versus HoloHub dir
```suggestion
def find_sdk_build_dir(local_sdk_root: Optional[Union[str, Path]] = None) -> str:
```

utilities/cli/util.py
```diff
@@ -310,6 +310,53 @@ def get_host_gpu() -> str:
...
+    """
+    find a suitable build directory in the SDK root
+    https://github.com/nvidia-holoscan/holoscan-sdk/blob/9c5b3c3d4831f2e65ebda6b79ae9b1c5517c6a7c/run#L226-L228
```
tbirdso: Thinking about enhancements for the search strategy and target directories here:
- Do we really want to target the build directory or is it better to target the SDK installation directory (`holoscan-sdk/install-<arch>-<gpu>`)? Maybe either is acceptable but install is preferred?
- Could we allow the user to pass just a build/install directory without the parent source dir? What about if the user has a local Holoscan SDK installation directory that does not match the Holoscan SDK source naming convention for some reason, such as if the user extracted a tarball into a custom folder?
- Are there expected contents we could rely on inside the folder to verify we have a valid SDK installation, rather than going solely based on naming schemes?

CMake uses [find_package](https://cmake.org/cmake/help/latest/command/find_package.html) to find and verify that some package is available and complete, typically with some `<project>-config.cmake` or `Find<project>.cmake` file. Maybe we don't want to be quite that extensive in our check, but could still change our search strategy to verify that some `holoscan-config.cmake` file is available in whatever custom SDK root we find?

IIRC the strategy here is at parity with the previous `run` script, so updates to the search strategy could be considered enhancements for followup and are not necessarily blocking.

utilities/cli/container.py
```diff
@@ -593,7 +594,8 @@ def get_pythonpath_options(
         """Get PYTHONPATH configuration"""
         benchmarking_path = "/workspace/holohub/benchmarks/holoscan_flow_benchmarking"
         if local_sdk_root:
-            sdk_paths = f"/workspace/holoscan-sdk/build/python/lib:{benchmarking_path}"
+            build_dir = find_build_dir(local_sdk_root)
+            sdk_paths = f"/workspace/holoscan-sdk/{build_dir}/python/lib:{benchmarking_path}"
```
tbirdso: Is the build path guaranteed to be relative to `/workspace/holoscan-sdk`?

doc/developer.md
```diff
@@ -124,6 +124,18 @@ To use a HoloHub container image built with a local Holoscan SDK container:
...
+Please refer to the [Holoscan SDK Developer Guide](https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/DEVELOP.md) for more details on how to build the Holoscan SDK from source.
+
+In the container, to verify the build directory is mounted correctly, run the following command:
+```bash
+$ python -c "import holoscan; print(holoscan.__file__)"
+```
+The output should be something like:
+```bash
+/workspace/holoscan-sdk/build-x86_64/python/lib/holoscan/__init__.py
+```
```
tbirdso: Note that a custom Holoscan SDK installation built with `-DHOLOSCAN_BUILD_PYTHON:BOOL=OFF` would not have Python bindings, so this check isn't guaranteed to succeed. Maybe we could add a note that `/workspace/holoscan-sdk` can be manually inspected for verification if Python supported is disabled?

utilities/cli/container.py
wyli: good point, I'll rename it to `find_hsdk_build_rel_dir` for clarity

utilities/cli/util.py
wyli: combining the other comment I'll update it to find_hsdk_build_rel_dir

utilities/cli/util.py
wyli: thanks, all good ideas, I've updated the search logic to prefer install-* and allow for directory without parent dir, validation of the dir based on some subdirs

.github/workflows/check_docker.yml
```diff
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, fixes-sdk-path]
```
```suggestion
    branches: [main, holoscan-sdk-lws2]
```

utilities/cli/util.py
```diff
@@ -329,32 +329,78 @@ def get_arch_gpu_str() -> str:
...
+    """
+    Validate if a directory contains a valid Holoscan SDK installation.
+    """
+    path = Path(path) if isinstance(path, str) else path
+    if not path.exists() or not path.is_dir() or not (path / "lib").exists():
+        return False
+    # Check for at least one of these to confirm it's a Holoscan SDK
+    return (path / "include" / "holoscan").exists() or (
+        path / "lib" / "cmake" / "holoscan"
+    ).exists()
```
tbirdso: Existence of `holoscan-config.cmake` or `HoloscanConfig.cmake` might be a more concrete check than the `include/holoscan` or `lib/cmake/holoscan` path. See [CMake search strategy](https://cmake.org/cmake/help/v3.26/command/find_package.html#id6) for reference

tbirdso (commented):
The update looks good, one suggestion to align to CMake search strategy

utilities/cli/util.py
wyli: thanks, updated

PR #1023, opened 2025-07-24
holohub cli:  Fix for --configure-args

wendell-hom:
When running `./holohub build --configure-args="..." endoscopy_tool_tracking`, it looks like the holohub cli parses the command but the resulting build string is incorrect.

It's running this

```
docker run ... -c "./holohub build endoscopy_tool_tracking --local --pkg-generator DEB \
    --configure-args "-DAPP_colonoscopy_segmentation=1""
```
This fix updates the code so that it runs this command instead
```
docker run ... -c "./holohub build endoscopy_tool_tracking --local --pkg-generator DEB \
    --configure-args='-DAPP_colonoscopy_segmentation=1'"
```


wyli (approved):
thanks, looks good to me -- root cause same as https://github.com/nvidia-holoscan/holohub/pull/931

utilities/cli/holohub.py
```diff
@@ -732,7 +732,7 @@ def handle_build(self, args: argparse.Namespace) -> None:
                 build_cmd += " --benchmark"
             if getattr(args, "configure_args", None):
                 for configure_arg in args.configure_args:
-                    build_cmd += f' --configure-args "{configure_arg}"'
+                    build_cmd += f" --configure-args='{configure_arg}'"
```
wyli: suggest 'shlex.quote' to be consistent with the run_arg usage https://github.com/nvidia-holoscan/holohub/pull/931
```suggestion
                    build_cmd += f" --configure-args={shlex.quote(args. configure_arg)}"
```

could you please also help update the other cases
https://github.com/nvidia-holoscan/holohub/blob/d77ef04925a93ab41a8430593d04a247ea2183e4/utilities/cli/holohub.py#L944
https://github.com/nvidia-holoscan/holohub/blob/d77ef04925a93ab41a8430593d04a247ea2183e4/utilities/cli/holohub.py#L1384

wendell-hom: Updated, thanks!

wendell-hom:
@tbirdso @wyli thanks for the review, ready to merge if everything looks ok

PR #1028, opened 2025-08-01
CI: update copyright check

wyli:
- allow for external affiliation in the copyright header
- more permissive matching pattern

### DIFF
--- a/applications/volume_rendering_xr/.devcontainer/library-scripts/common-debian.sh
+++ b/applications/volume_rendering_xr/.devcontainer/library-scripts/common-debian.sh
@@ -3,7 +3,7 @@
 #   https://github.com/microsoft/vscode-dev-containers/blob/v0.241.1/containers/ubuntu/.devcontainer/library-scripts/common-debian.sh
 
 #-------------------------------------------------------------------------------------------------------------
-# Copyright (c) 2022-2025 Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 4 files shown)


applications/volume_rendering_xr/.devcontainer/library-scripts/common-debian.sh
```diff
@@ -3,7 +3,7 @@
 #   https://github.com/microsoft/vscode-dev-containers/blob/v0.241.1/containers/ubuntu/.devcontainer/library-scripts/common-debian.sh
 
 #-------------------------------------------------------------------------------------------------------------
-# Copyright (c) Microsoft Corporation. All rights reserved.
+# Copyright (c) 2022-2025 Microsoft Corporation. All rights reserved.
```
tbirdso: Not sure about this one, we typically don't want to update external copyrights but also it should indeed have a date. Seems OK to update here for now but we might rethink it if it pops up again in the future? Not a blocker

wyli:
```suggestion
# Copyright (c) 2022 Microsoft Corporation. All rights reserved.
```

wyli: updated to just add 2022 (the ci will fail as expected and allowed because we modify a file without updating it to the current year)

PR #1030, opened 2025-08-01
Bump Holoscan SDK Version to 3.5.0

bhashemian:


README.md
```diff
@@ -9,7 +9,8 @@ Visit [https://nvidia-holoscan.github.io/holohub](https://nvidia-holoscan.github
 
 This is a central repository for the NVIDIA Holoscan AI sensor processing community to share reference applications, operators, tutorials and benchmarks. We invite users and developers of the Holoscan platform to reuse and contribute to this repository.
 
-# Table of Contents
+## Table of Contents
```
jjomier, line 12: Why is this changed?

bhashemian: It is simply to ensure that the README is conforming to Markdown conventions. Only one top-level heading (#) should exist but please let me know if there is any reason to keep it that way.

jjomier (approved):

PR #1035, opened 2025-08-04
update 'dependencies' to 'requirements' in component metadata

wyli:
changes discussed while developing new feature preset mode https://github.com/nvidia-holoscan/holohub/pull/972


agirault (approved):
ty! change only makes sense after your other MR but it's fine if merged before, either way.

PR #1047, opened 2025-08-06
Add CUDA::cuda_driver to target link libraries

wendell-hom:
Need to link against CUDA::cuda_drivers to avoid undefined references for some envs


tbirdso (changes requested):
My impression is that we are only observing the issue in question in internal builds, can you please confirm whether you're seeing the link issue in the public case?

If not, let's pursue an internal fix.

agirault:
Can we close this?

PR #1048, opened 2025-08-06
add python 310 check

wyli:


pyproject.toml
```diff
@@ -32,6 +32,7 @@ extend-exclude = '''
   | ^install*/*  # exclude install folders
   | cmake/pybind11/*  # exclude the files under cmake/pybind11/
   | .local/*  # exclude the files under .local/
+  | .ipynb  # mute warning 'Skipping .ipynb files as Jupyter dependencies are not installed.'
```
tbirdso, line 35: We could consider adding notebook linting in the future but I think we only have one or two right now, not a blocker

PR #1050, opened 2025-08-06
Handle List of Languages for Holohub Container

bhashemian:
When defined languages in `metadata.json` is a list, like `["C++", "Python"]`, which is the case when having a Python binding. The CLI failed to find the project, and also the container failed to check if there is any language specific container. This PR, addressed those issues.


bhashemian:
Making it a draft PR to check on the cli test cases before reviewing.

utilities/cli/util.py
```diff
@@ -424,15 +424,31 @@ def get_group_id(group: str) -> Optional[int]:
...
+
+    # Normalize language name
+    language = language.lower()
+    if language in ["cpp", "c++"]:
         return "cpp"
-    elif language.lower() == "python" or language.lower() == "py":
+    if language in ["python", "py"]:
         return "python"
-    return None
+    raise ValueError(f"Invalid language: {language}")
```
wyli, line 444: this might be restrictive for apps where 'language' is an optional field

bhashemian: Thanks, that's a good point. The first check that this function does is to ensure is not empty (`""`, `[]`, `None`), otherwise, it returns `""`. The supported languages are currently only python and c++, so it might not be too restrictive in that sense, right? On the other hand, this check will let the user know if they have misspelled the language like `pythn`. Does it make sense?

wyli (approved):

PR #1054, opened 2025-08-07
Documentation: Add README for Prohawk Video Processing Operator

bhashemian:


bhashemian:
@nvidia-holoscan/holohub I couldn’t find the author of this operator, as mentioned in the metadata (“Tim Wooldridge”), to request a review. Does anyone know him? 

PS: the commits are signed by @jjomier.

operators/prohawk_video_processing/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+```bash
+# Debug build
+cmake .. -DCMAKE_BUILD_TYPE=Debug
+make -j$(nproc)
+
+# Release build
+cmake .. -DCMAKE_BUILD_TYPE=Release
+make -j$(nproc)
+```
```
tbirdso: Please update to HoloHub CLI

operators/prohawk_video_processing/README.md
```diff
@@ -0,0 +1,252 @@
...
+- **Processing**: Can be chained with other video processing operators
+
+### With Existing Video Pipelines
+
+The operator can be easily integrated into existing video processing pipelines by:
+
+1. Adding the operator to your Holoscan fragment
+2. Connecting it to your video source
+3. Connecting the output to your desired sink
+4. Configuring the appropriate filter for your use case
```
tbirdso: Isn't this the case for all Holoscan SDK operators in general?

operators/prohawk_video_processing/README.md
```diff
@@ -0,0 +1,252 @@
...
+make -j$(nproc)
+
+# Release build
+cmake .. -DCMAKE_BUILD_TYPE=Release
+make -j$(nproc)
+```
+
+### Testing
+
+The operator includes interactive testing capabilities through the built-in display window. You can test different filters and configurations in real-time.
```
tbirdso: Have we confirmed this?

operators/prohawk_video_processing/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Extending the Operator
+
+To add new filters or modify existing ones:
+
+1. Edit `prohawkop.cpp` to add new filter cases
+2. Update the interactive controls in the compute method
+3. Rebuild the operator
+4. Test with your video stream
```
tbirdso: I don't think we should recommend editing the `.cpp` file

operators/prohawk_video_processing/README.md
bhashemian: Thanks, removed it.

operators/prohawk_video_processing/README.md
bhashemian: Thanks, removed it.

operators/prohawk_video_processing/README.md
bhashemian: Thanks for your comment. The CLI wouldn’t work because it doesn’t provide a Dockerfile, so the dependencies wouldn’t be met. Thus, I removed it.

tbirdso: HoloHub CLI is still intended to streamline build commands in the local build environment though, regardless of whether it's containerized or not

Generally:
```
./holohub build --local <project>
```

I do see a Dockerfile in the demo app providing dependencies:
```
./holohub build <project> --docker-file applications/prohawk_video_replayer/Dockerfile
```

bhashemian: Thanks, @tbirdso! I definitely can add a Dockerfile for this operator, but is that what we want to do for all the operators? Currently, most operators lack building instructions in their README files and rely on the related application to provide the dependencies. Am I missing something?

tbirdso: No, what I meant was that if you wanted to keep build instructions in the README, we should provide command strings using the HoloHub CLI rather than `cmake` commands

However I see build instructions are removed from the latest README version, so no issue 👍

operators/prohawk_video_processing/README.md
```diff
@@ -0,0 +1,137 @@
...
+## Usage
+
+Please refer to [Prohawk Video Processing](../../applications/prohawk_video_replayer/README.md) for requirements and usage examples.
+
+### Basic Usage
+
+The operator can be used in Holoscan applications to process video streams:
+
+```python
+import holoscan as holo
```
tbirdso: I don't think I've seen this before, are we recommending this naming scheme?

operators/prohawk_video_processing/README.md
```diff
@@ -0,0 +1,137 @@
...
+    // Constructor
+    ProhawkOp();
+    
+    // Setup method
+    void setup(OperatorSpec& spec) override;
+    
+    // Compute method
+    void compute(InputContext& op_input, OutputContext& op_output, ExecutionContext&) override;
+};
+```
```
tbirdso: All operators have these required methods, is this adding anything?

bhashemian: I agree, removed it. Thanks

operators/prohawk_video_processing/README.md
bhashemian: Not really, removed it. Thanks

PR #1066, opened 2025-08-11
App: Update `colonoscopy_segmentation` README

tbirdso:
Update `colonoscopy_segmentation` README in response to internal feedback.

Also add mode presets to streamline running with recommended replayer and AJA settings.

### DIFF
--- a/applications/colonoscopy_segmentation/README.md
+++ b/applications/colonoscopy_segmentation/README.md
@@ -2,63 +2,76 @@
 
 Full workflow including a generic visualization of segmentation results from a polyp segmentation models.
 
-### Requirements
+## Quick Start
+
+```bash
+./holohub run colonoscopy_segmentation
+```
+
+## Requirements
 
 - Python 3.8+
 - The provided applications are configured to either use the AJA capture card for input stream, or a pre-recorded video of the colonoscopy data (replayer). Follow the [setup instructions from the user guide](https://docs.nvidia.com/holoscan/sdk-user-guide/aja_setup.html) to use the AJA capture card.
 
-### Data
+## Data
 
 [📦️ (NGC) Sample App Data for AI Colonoscopy Segmentation of Polyps](https://catalog.ngc.nvidia.com/orgs/nvidia/teams/clara-holoscan/resources/holoscan_colonoscopy_sample_data)
 
 The data is automatically downloaded and converted to the correct format when building the application.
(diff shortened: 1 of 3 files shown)


applications/colonoscopy_segmentation/README.md
```diff
@@ -2,63 +2,76 @@
...
-    python3 colonoscopy_segmentation.py --source=aja --no-contours
-    ```
-Note that segmentation contours can be shown (instead of segmentation masks) by changing `--no-contours` to `--contours`.
+### Application Configurations
 
-### Holoscan SDK version
+To pass runtime arguments to the application:
+```bash
+./holohub run colonoscopy_segmentation --run-args="--contours --source=\"replayer\""
+```
```
tbirdso, line 54: Hi @wyli , my goal in this PR is to use modes such that:
1. Users can use the default replayer mode or the aja mode with minimal CLI arguments
2. Users seeking more advanced behavior can specify run arguments such as enabling contours or setting the input source without using a mode preset

However, observed behavior is such that setting any modes at all in the metadata.json prevents us from using `--run-args` at all, since we always assume we are using the default mode

Could you please confirm behavior and whether there is a suggested approach to accomplish both (1) and (2) above?

wyli: correct, it's a critical known issue from the first PR, the current CLI only supports either (1) or (2) but not both. A possible solution is that we enable all args overrides only in the default mode. but I'm open to any other suggestions about mode + command line overrides cc @agirault @jjomier

tbirdso: What about an implicit default mode with no trailing arguments that always accepts `--run-args`? I.e. let `./holohub run colonoscopy_segmentation` run `python3 colonoscopy_segmentation.py` without any trailing arguments and accept `./holohub run colonoscopy_segmentation [--run-args ...]`

Seems like this could be achieved via your proposal of accepting override args in the default mode

This seems like a major blocker to modes adoption, can't reliably add modes to colonoscopy_segmentation without this granularity

wyli: sure, I'll implement a version for discussions

wyli: with the latest holohub main branch the overriding works now when mode is not specified (default mode) `./holohub run colonoscopy_segmentation --run-args="--contours --source=replayer"`

tbirdso: Nice, thank you for adding that! Let me update

PR #1073, opened 2025-08-14
CLI: add CLI args overriding for default mode

wyli:
Enable CLI parameter overrides for default modes when an app has multiple modes and the command implicitly calls the default mode.

**Implicit Default Mode Overrides:**
When a mode is auto-selected (no mode name specified), CLI parameters are allowed and will override mode settings with warnings:

```bash
# These work - mode auto-selected, CLI parameters override with warnings
./holohub run holochat --run-args="--debug"
./holohub run myapp --build-with="ops" --docker-opts="--net=host"
```

**Explicit Mode Configuration:**
When a mode is explicitly specified, the CLI ensures consistent configuration by using the mode's predefined settings.
CLI parameters override mode settings is not supported:

```bash
# with the mode specified, CLI parameters are not supported
./holohub build myapp standard --build-with="ops"      # not supported mode + CLI args if 'standard' mode has 'build.depends'
./holohub run holochat cloud --run-args="--test=case"  # not supported mode + CLI args
```

In this case, users could either (1) modify the mode settings in `metadata.json` to match the CLI parameters, or (2) use the default mode with additional CLI parameters.

- modified `body_pose_estimation` app as an example

### DIFF
--- a/.github/workflows/check_docker.yml
+++ b/.github/workflows/check_docker.yml
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2, default-overriding]
+    branches: [main, holoscan-sdk-lws2]
 
 permissions:
   contents: read


.github/workflows/check_docker.yml
```diff
@@ -2,7 +2,7 @@ name: Check Docker
 
 on:
   push:
-    branches: [main, holoscan-sdk-lws2]
+    branches: [main, holoscan-sdk-lws2, default-overriding]
```
```suggestion
    branches: [main, holoscan-sdk-lws2]
```

PR #1075, opened 2025-08-14
Update pydoc.hpp

varya3:
Changed include directive for project-local header macros.hpp to double quotes for local search


tbirdso (commented):
Hi, I do not see a local `macros.hpp` file, could you please comment on the expected search path? Expected to see the file in https://github.com/nvidia-holoscan/holohub/tree/main/operators/orsi/orsi_segmentation_postprocessor/python

Please review the pipeline failure as well

tbirdso:
Closing due to inactivity, please reopen with responses to the questions above if this is a point of concern.

PR #1076, opened 2025-08-14
Quality of life updates for basic network operators

dylan-eustice:
- RX: Parameterize a max burst interval
  - This is an optional parameter that dictates the maximum amount of time (ms) that can pass between bursts when packets are being held. This is an optional parameter and the default value leaves the functionality of the operator unchanged.
- TX: Parameterize whether pointer memory is deleted
  - If 'delete_payload = False', the memory passed to the network TX operator is not deleted. The default behavior of deleting the memory will be functionally identical to before. This allows memory in upstream operators to be re-used in subsequent TX calls.


cliffburdick (approved):

bhashemian:
@dylan-eustice could you please address the DCO check issue?

You should be able to fix it by 
- running `git rebase HEAD~10 --signoff` in your local repo, and 
- then force push those changes: `git push --force-with-lease origin deustice/bno-updates`

bhashemian:
@agirault please let us know if you have any comments on this PR. Otherwise, I will merge it once DCO issue is fixed. Thanks

dylan-eustice:
@bhashemian fixed DCO issue

tbirdso:
From link-check error:

>No links were found. This usually indicates a configuration error.
If this was expected, set 'failIfEmpty: false' in the args.

I'm not sure we should be failing if PR-altered files do not contain URLs to check. Maybe @wyli  could help review the pipeline approach? Probably not a PR blocker.

bhashemian:
> From link-check error:
> 
> > No links were found. This usually indicates a configuration error.
> > If this was expected, set 'failIfEmpty: false' in the args.
> 
> I'm not sure we should be failing if PR-altered files do not contain URLs to check. Maybe @wyli could help review the pipeline approach? Probably not a PR blocker.

@tbirdso, it’s happening for some other PRs as well. @wyli already added this suggested argument to the CI two weeks ago, so I’m not sure why it’s complaining. Maybe it’s not the right place. @wyli, I’d appreciate it if you could take a look. 

https://github.com/nvidia-holoscan/holohub/blob/c453c05d2e311fbc50807b57feab7c8b99944642/.github/workflows/check_urls.yml#L66

Anyways, not a blocker. Will merge it after @agirault's confirmation.

PR #1079, opened 2025-08-15
Enhance fetch_holohub_operator to support optional Python bindings

jjomier:
- Added DISABLE_PYTHON argument to fetch_holohub_operator function in FetchHolohubOperator.cmake to control the building of Python bindings.
- Updated README.md to reflect the new DISABLE_PYTHON parameter and provide examples for using Python bindings.
- Introduced main.py as an example of using Holohub operators in Python applications.

### DIFF
--- a/tutorials/holohub_operators_external_applications/README.md
+++ b/tutorials/holohub_operators_external_applications/README.md
@@ -307,7 +307,7 @@ After building your project, Python modules are typically installed in the `pyth
 cd build
 
 # Set the PYTHONPATH to include the built Python modules
-export PYTHONPATH=$PYTHONPATH:./python/lib
+export PYTHONPATH=$PYTHONPATH:$(pwd)/python/lib
 
 # Run your Python application
 python3 ../main.py


cmake/FetchHolohubOperator.cmake
```diff
@@ -57,12 +63,25 @@ function(fetch_holohub_operator OPERATOR_NAME)
...
+        # Add code to create python module output directory if building python
+        && echo "if(HOLOHUB_BUILD_PYTHON)" >> "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src/CMakeLists.txt"
+        && echo "  if(NOT CMAKE_INSTALL_LIBDIR)" >> "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src/CMakeLists.txt"
+        && echo "    set(CMAKE_INSTALL_LIBDIR lib)" >> "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src/CMakeLists.txt"
+        && echo "  endif()" >> "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src/CMakeLists.txt"
+        && echo "  set(HOLOHUB_PYTHON_MODULE_OUT_DIR \\$\\{CMAKE_BINARY_DIR\\}/python/\\$\\{CMAKE_INSTALL_LIBDIR\\}/holohub)" >> "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src/CMakeLists.txt"
+        && echo "  file(MAKE_DIRECTORY \\$\\{HOLOHUB_PYTHON_MODULE_OUT_DIR\\})" >> "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src/CMakeLists.txt"
+        && echo "endif()" >> "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src/CMakeLists.txt"
+        && echo "add_subdirectory(operators/${ARGS_PATH})" >> "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src/CMakeLists.txt"
```
tbirdso: Can we compose this into a single "echo" command rather than splitting over multiple calls, in case there is an interruption during setup?

tbirdso (commented):
Cool!

cmake/FetchHolohubOperator.cmake
```diff
@@ -27,15 +27,17 @@
...
 # Example:
 #   fetch_holohub_operator(realsense_camera)
 #   fetch_holohub_operator(dds_operator_base PATH dds/base)
 #   fetch_holohub_operator(custom_operator REPO_URL "https://github.com/custom/holohub.git")
 #   fetch_holohub_operator(custom_operator BRANCH "dev")
+#   fetch_holohub_operator(custom_operator DISABLE_PYTHON)
 function(fetch_holohub_operator OPERATOR_NAME)
 
-  cmake_parse_arguments(ARGS "" "PATH;REPO_URL;BRANCH" "" ${ARGN})
+  cmake_parse_arguments(ARGS "DISABLE_PYTHON" "PATH;REPO_URL;BRANCH" "" ${ARGN})
```
tbirdso, line 40: Thoughts on making this opt-in rather than opt-out?

```suggestion
  cmake_parse_arguments(ARGS "WITH_PYTHON" "PATH;REPO_URL;BRANCH" "" ${ARGN})
```

cmake/FetchHolohubOperator.cmake
```diff
@@ -49,6 +51,10 @@ function(fetch_holohub_operator OPERATOR_NAME)
     set(ARGS_BRANCH "main")
   endif()
 
+  if(NOT ARGS_DISABLE_PYTHON)
+    set(HOLOHUB_BUILD_PYTHON ON)
+  endif()
```
tbirdso, line 56: I think this is set here to populate the initial variable in the checkout cache. Could you please add a comment to clarify?

cmake/FetchHolohubOperator.cmake
jjomier: We want to build python by default.

cmake/FetchHolohubOperator.cmake
jjomier: This is documented as part of the function.

tutorials/holohub_operators_external_applications/README.md
```diff
@@ -277,6 +285,74 @@ cmake ..
...
+#### Running Python Applications
+
+After building your project, Python modules are typically installed in the `python/lib` directory within your build directory. To run Python applications that use the Holohub operators:
+
+```bash
+# Navigate to your build directory
+cd build
+
+# Set the PYTHONPATH to include the built Python modules
+export PYTHONPATH=$PYTHONPATH:./python/lib
```
tbirdso: Guard against changes to the working directory

```suggestion
export PYTHONPATH=$PYTHONPATH:$(pwd)/python/lib
```

PR #1082, opened 2025-08-18
Improve README for AprilTag Detection Operator

bhashemian:


operators/apriltag_detector/python/README.md
```diff
@@ -1,17 +1,59 @@
...
+
+- **`input`**: Video stream input (typically RGBA or RGB format)
+
+#### Output Ports
+
+- **`output`**: Detected AprilTag information including:
+  - Tag ID
+  - Corner coordinates
+  - Center position
+  - Detection confidence
```
tbirdso: If we're aiming to add API information for operators in their READMEs, it would help to either add more details here on the format of output data, or to at least add a link to inline definitions where developers can find the specs in order to use the output struct

PR #1085, opened 2025-08-19
CLI: add build-name-suffix to ./holohub test

wyli:
Added `--build-name-suffix` parameter to the `./holohub test` command

the existing `-DTAG` option is hard-coded to `base_img.split(":")[-1]` which is not flexible enough for cases such as `no-docker-build`

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -380,6 +380,10 @@ class HoloHubCLI:
         test.add_argument(
             "--no-docker-build", action="store_true", help="Skip building the container"
         )
+        test.add_argument(
+            "--build-name-suffix",
+            help="Suffix to use for ctest build name (defaulting to the image tag)",
+        )
         test.set_defaults(func=self.handle_test)
 
         # Add clear-cache command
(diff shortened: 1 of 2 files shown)


utilities/cli/holohub.py
```diff
@@ -380,6 +380,10 @@ def _create_parser(self) -> argparse.ArgumentParser:
         test.add_argument(
             "--no-docker-build", action="store_true", help="Skip building the container"
         )
+        test.add_argument(
+            "--build-name-suffix",
+            help="Suffix to use for ctest build name (defaulting to the image tag)",
```
tbirdso, line 385: Why "suffix" rather than `--build-name` or `--ctest-tag`?

wyli: This is mainly used in the ctest config as part of the build name: https://github.com/nvidia-holoscan/holohub/blob/7e1cb9c7b06b1e3e6f6ea033911a76bf8e9d4f3d/utilities/testing/holohub.container.ctest#L35 although the variable name is 'TAG' it could actually be a branch name or commit hash, we might use it to track the GPU type as well...

tbirdso: Ah got it, thanks!

PR #1087, opened 2025-08-19
Impact of Async Lock-free Buffer with SCHED_DEADLINE

sohamm17:


applications/async_buffer_deadline/README.md
```diff
@@ -0,0 +1,68 @@
...
+```bash
+# Basic run with default settings (100 messages, no async buffer)
+./holohub run async_buffer_deadline
+
+# Run with async buffer enabled
+./holohub run async_buffer_deadline -- --async-buffer
+
+# Run with custom message count and periods
+./holohub run async_buffer_deadline -- --messages 50 --tx1-period 15 --tx2-period 25 --async-buffer
+```
```
tbirdso: @wyli  potential candidate for HoloHub CLI mode presets support

tutorials/async_buffer_deadline/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Conclusion and Guidance
+
+The asynchronous lock-free buffer connection between the operators enables
+connected operators to run independently. When they are configured with 
+`SCHED_DEADLINE` Linux scheduling policy, they can run independently with 
+their respective runtime and periods. When the default buffer is used,
+operators may depend on other operator's runtime and periods, even when
+`SCHED_DEADLINE` policy is associated with them. Therefore, developers must
+take precaution when using `SCHED_DEADLINE` policy with the default buffer.
```
tbirdso: Not sure I 100% follow the impact here, could you please clarify the behavior of operators when using the default buffer + `SCHED_DEADLINE`? Is the impact that operators will continue normal operation without performance benefit from `SCHED_DEADLINE`, or does this mean that there is risk of bad behavior where operators run independently when they are expected to rely on other operators' runtime+periods for correct performance of the default buffer?

tbirdso (commented):
Looks good! One clarifying question on conclusions and risks.

tutorials/async_buffer_deadline/README.md
sohamm17: @tbirdso your first point is correct. When default buffers are used with `SCHED_DEADLINE`, operators will continue normal operation. However, they may not see any benefit of using `SCHED_DEADLINE` since they are dependent on other operator's behavior. It is also true that they may see degraded performance since operators have to satisfy multiple constraints when both default buffer and `SCHED_DEADLINE` are used. Let me clarify this in the README.

applications/async_buffer_deadline/main.cpp
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+  void compute(InputContext&, OutputContext& op_output, ExecutionContext&) override {
+    PingMessage msg;
+    msg.index = index_;
+    msg.timestamp = std::chrono::duration_cast<std::chrono::microseconds>(
+                        std::chrono::high_resolution_clock::now().time_since_epoch())
+                        .count();
+
+    // Busy-loop for 5ms before emitting
```
tbirdso:
```suggestion
    // Busy-loop for the requested processing interval before emitting
```

applications/async_buffer_deadline/main.cpp
```diff
@@ -0,0 +1,359 @@
...
+        HOLOSCAN_LOG_INFO(
+            "Ping RX2: Received index {} (latency: {:.3f} ms)", msg.index, latency_ms);
+      }
+    }
+
+    // Busy-loop for 10ms after receiving
+    auto start_time = std::chrono::high_resolution_clock::now();
+    while (std::chrono::high_resolution_clock::now() - start_time < std::chrono::milliseconds(1)) {
+      // Busy wait
+    }
```
tbirdso: 1ms or 10ms?

tutorials/async_buffer_deadline/README.md
tbirdso: Thanks for clarifying, the updated README is clear 👍

applications/async_buffer_deadline/main.cpp
sohamm17: Thanks for pointing out! Fixed both.

PR #1089, opened 2025-08-20
SlangShaderOp: Use CUDA driver API instead of runtime API

AndreasHeumann:
Runtime API for CUDA library management needs at least CTK 12.8 but we want to support earlier versions also.


tbirdso (approved):
Thank you, looks good! I see the cuda driver APIs here are available in >= 12.0, no concerns.

PR #1090, opened 2025-08-20
CLI: ./holohub test update ctest options

wyli:
~change the ctest config to unquoted `${configure_options}` to allow for multiple `-Doptions`~

existing `./holohub test --cmake-options` will pass the -D options to the app level.
there's still a need for `--ctest-options` with additional config directly set for `ctest ...`

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -376,13 +376,15 @@ class HoloHubCLI:
         test.add_argument("--platform-name", help="Platform name")
         test.add_argument(
             "--cmake-options",
-            help="CMake options (semicolon-separated for multiple). "
-            "Example: --cmake-options='-DCUSTOM_OPTION=ON;-DDEBUG_MODE=1'",
+            action="append",
+            help="CMake options"
+            "example: --cmake-options='-DCUSTOM_OPTION=ON' --cmake-options='-DDEBUG_MODE=1'",
         )
         test.add_argument(
             "--ctest-options",
-            help="CTest options (semicolon-separated for multiple). "
-            "Example: --ctest-options='-DGPU_TYPE=rtx4090;-DDEBUG_MODE=ON'",
+            action="append",
+            help="CTest options"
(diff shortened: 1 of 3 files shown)


utilities/cli/holohub.py
```diff
@@ -374,7 +374,11 @@ def _create_parser(self) -> argparse.ArgumentParser:
         test.add_argument("--site-name", help="Site name")
         test.add_argument("--cdash-url", help="CDash URL")
         test.add_argument("--platform-name", help="Platform name")
-        test.add_argument("--cmake-options", help="CMake options")
+        test.add_argument(
+            "--cmake-options",
+            help="CMake configuration options (semicolon-separated for multiple). "
```
wyli:
```suggestion
            help="CMake options (semicolon-separated for multiple). "
```

utilities/cli/holohub.py
```diff
@@ -374,7 +374,16 @@ def _create_parser(self) -> argparse.ArgumentParser:
...
+        test.add_argument(
+            "--cmake-options",
+            help="CMake options (semicolon-separated for multiple). "
+            "Example: --cmake-options='-DCUSTOM_OPTION=ON;-DDEBUG_MODE=1'",
+        )
+        test.add_argument(
+            "--ctest-options",
+            help="CTest options (semicolon-separated for multiple). "
+            "Example: --ctest-options='-DGPU_TYPE=rtx4090;-DDEBUG_MODE=ON'",
+        )
```
tbirdso: Wondering about delimiters, I recently observed behavior in `./holohub build` requiring multiple configure arg parameters rather than space-delimited, can we align to one or the other?

I.e.
```
./holohub build --configure-args="-DMYARG1:BOOL=ON" --configure-args="-DMYARG2:BOOL=OFF"
```

Versus
```
./holohub test --cmake-options="-DMYARGS1:BOOL=ON;-DMYARGS2:BOOL=OFF"
```

wyli: the current `--configure-args` is easier to write and parse. I'll update to be consistent with that style, that'll be breaking change to the current `--cmake-options` though.

tbirdso (approved):
Thank you!

PR #1098, opened 2025-08-22
Do not check links of deleted files.

cosmith-nvidia:
The link-check workflow will currently check links for the result of `git diff --name-only`, which will return added, modified, and deleted files. Applying `--diff-filter=d` will exclude deleted files so that those links are not checked.


tbirdso (approved):
Thanks Connor, seems like somehow this is the first time the check has encountered a deleted file since it was added. Fix looks good.

Please sign the commit to unblock merge: `git commit --amend --no-edit -s`

PR #1100, opened 2025-08-25
Enable dynamic packet size in basic network TX

dylan-eustice:
Optional field, behavior is unchanged if not set

### DIFF
--- a/operators/basic_network/basic_network_operator_rx.cpp
+++ b/operators/basic_network/basic_network_operator_rx.cpp
@@ -169,7 +169,8 @@ void BasicNetworkOpRx::compute([[maybe_unused]] InputContext&, OutputContext& op
 
   // Only emit if we have packets to send
   if (pkts_in_batch_ > 0) {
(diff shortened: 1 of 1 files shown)


dylan-eustice:
@cliffburdick @agirault I've been making these updates to BNO backward compatible and as minimal as possible, so that older applications using the operator don't break. Because of that, this PR feels like it makes the logic more opaque and brittle. Happy to take suggestions or rework how it's implemented if you're ok with bigger changes to the overall operator flow.

cliffburdick:
> @cliffburdick @agirault I've been making these updates to BNO backward compatible and as minimal as possible, so that older applications using the operator don't break. Because of that, this PR feels like it makes the logic more opaque and brittle. Happy to take suggestions or rework how it's implemented if you're ok with bigger changes to the overall operator flow.

Right now I don't know of anyone using BNO for anything other than test applications, so I think it's perfectly fine to possibly break backwards compatibility. @awthomp do you know anyone?

cliffburdick (approved):

dylan-eustice:
Thanks @cliffburdick. Made a couple of changes, feeling better about it now. It should still be backward compatible though.

PR #1103, opened 2025-08-26
App: Fix H.264 data download at build time

tbirdso:
Address error where `LD_PRELOAD` fix for runtime H.264 issue blocks NGC CLI from properly downloading Endoscopy Tool Tracking data at build time.

Provide README instructions to distinguish required build and runtime environments for Python apps.

### DIFF
--- a/applications/h264/h264_endoscopy_tool_tracking/README.md
+++ b/applications/h264/h264_endoscopy_tool_tracking/README.md
@@ -38,7 +38,6 @@ The data is automatically downloaded when building the application.
 Separate build and run commands are required to address the known [symbol loading issue](../README.md#symbol-error-at-load).
 
 ```bash
-# C++ version
 ./holohub build h264_endoscopy_tool_tracking --language python
 
 # Python version
--- a/applications/h264/h264_endoscopy_tool_tracking/README.md
+++ b/applications/h264/h264_endoscopy_tool_tracking/README.md
@@ -39,7 +39,7 @@ Separate build and run commands are required to address the known [symbol loadin
 
 ```bash
 # C++ version
-./holohub build h264_endoscopy_tool_tracking --language cpp
+./holohub build h264_endoscopy_tool_tracking --language python
 
 # Python version
 # Note: LD_PRELOAD required to address symbol issue
--- a/applications/h264/h264_endoscopy_tool_tracking/README.md
+++ b/applications/h264/h264_endoscopy_tool_tracking/README.md
@@ -25,16 +25,26 @@ can be specified in the 'h264_endoscopy_tool_tracking.yaml' file.
 
 The data is automatically downloaded when building the application.
 
-## Building and Running H.264 Endoscopy Tool Tracking Application
+## Build and Run the H.264 Endoscopy Tool Tracking Application
 
-* Building and running the application from the top level Holohub directory:
+### C++
+
+```bash
+./holohub run h264_endoscopy_tool_tracking --language cpp
+```
+
+### Python
+
+Separate build and run commands are required to address the known [symbol loading issue](../README.md#symbol-error-at-load).
 
 ```bash
 # C++ version
-./holohub run h264_endoscopy_tool_tracking --language cpp
+./holohub build h264_endoscopy_tool_tracking --language cpp
 
 # Python version
-./holohub run h264_endoscopy_tool_tracking --language python
+# Note: LD_PRELOAD required to address symbol issue
+./holohub run h264_endoscopy_tool_tracking --language python \
+    --docker-opts="-e LD_PRELOAD=/opt/nvidia/holoscan/lib/libgxf_core.so"
 ```
 
 Important: on aarch64, applications also need tegra folder mounted inside the container and
--- a/applications/h264/h264_endoscopy_tool_tracking/python/metadata.json
+++ b/applications/h264/h264_endoscopy_tool_tracking/python/metadata.json
@@ -48,9 +48,19 @@
         }
       ]
     },
-    "run": {
-      "command": "python3 <holohub_app_source>/h264_endoscopy_tool_tracking.py --data <holohub_data_dir>/endoscopy",
-      "workdir": "holohub_bin"
-    }
-  }
+		"modes": {
+			"replayer": {
+				"description": "Use a pre-recorded Endoscopy video as input",
+				"build": {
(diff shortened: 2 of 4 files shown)


applications/h264/h264_endoscopy_tool_tracking/python/metadata.json
```diff
@@ -48,9 +48,19 @@
...
+		"modes": {
+			"replayer": {
+				"description": "Use a pre-recorded Endoscopy video as input",
+				"build": {
+					"docker_build_args": ["--target", "base"]
+				},
+				"run": {
+					"command": "python3 <holohub_app_source>/h264_endoscopy_tool_tracking.py --data <holohub_data_dir>/endoscopy",
+					"workdir": "holohub_bin",
+					"docker_build_args": ["--target", "run"]
```
wyli: could unset the var using the run args?
```suggestion
					"docker_run_args": ["-e LD_PRELOAD="]
```

tbirdso: This doesn't seem to solve the issue, the root cause is that the CMake build and Python app runtime are occurring in the same environment. I'm seeking a way to have LD_PRELOAD set to a different value between build and runtime.

I guess another way to do this is to require two separate user commands:
```base
# Build without LD_PRELOAD
./holohub build h264_endoscopy_tool_tracking --build-args="--target base"

# Run with LD_PRELOAD
./holohub run h264_endoscopy_tool_tracking --build-args="--target run"
```

We've discussed supporting separate build and runtime container environments with a single preset in the past, seems like this is not supported at the moment but would be a good future goal?

wyli: I see, another option is to set the run.command:  `LD_PRELOAD=/opt/nvidia/holoscan/lib/libgxf_core.so python3 <holohub_app_source>/h264_endoscopy_tool_tracking.py...`

tbirdso: >another option is to set the run.command

I tried this but it failed, attempts to run "LD_PRELOAD=/opt/nvidia/holoscan/lib/libgxf_core.so" as the executable

wyli: ah yes, it doesn't support this type of injection... if it's a python specific issue maybe it's ok to do it in the python script before the hsdk module importing

```py
if os.path.exists("/opt/nvidia/holoscan/lib/libgxf_core.so") and not os.environ.get("LD_PRELOAD"):
    os.environ["LD_PRELOAD"] = "/opt/nvidia/holoscan/lib/libgxf_core.so"
```

applications/h264/h264_endoscopy_tool_tracking/README.md
```diff
@@ -25,16 +25,26 @@ can be specified in the 'h264_endoscopy_tool_tracking.yaml' file.
...
 ./holohub run h264_endoscopy_tool_tracking --language cpp
+```
+
+### Python
+
+Separate build and run commands are required to address the known [symbol loading issue](../README.md#symbol-error-at-load).
+
+```bash
+# C++ version
+./holohub build h264_endoscopy_tool_tracking --language cpp
```
wyli: specify `cpp` here in the python section looks a bit strange...

```suggestion
./holohub build h264_endoscopy_tool_tracking --language python
```

applications/h264/h264_video_decode/README.md
```diff
@@ -22,17 +22,26 @@ input file can be specified in the 'h264_video_decode.yaml' file.
...
-* Building and running the application from the top level Holohub directory:
+### C++
+
+```bash
+./holohub run h264_video_decode --language cpp
+```
+
+### Python
+
+Separate build and run commands are required to address the known [symbol loading issue](../README.md#symbol-error-at-load).
```
wyli, line 35: couldn't find this `../README.md#symbol-error-at-load`, seems the url checker doesn't detect this as well

tbirdso: Confused, I see the link check passed and I can view the linked section at https://github.com/nvidia-holoscan/holohub/blob/744cc99ad448f80e348eedbfe7f50fb35cb83a81/applications/h264/README.md#symbol-error-at-load ?

applications/h264/h264_endoscopy_tool_tracking/python/metadata.json
tbirdso: Gave this a shot but still see the error:
```
[error] [gxf_codelet.cpp:44] Unable to find the GXF type name ('nvidia::gxf::VideoReadBitStream') for the codelet
```

I think `LD_PRELOAD` would need to be set before running `python3` to have effect.

Seems like documenting the user runtime requirement as done in the latest README update is still the best way at the moment:
```
# Note: LD_PRELOAD required to address symbol issue
./holohub run h264_endoscopy_tool_tracking --language python \
    --docker-opts="-e LD_PRELOAD=/opt/nvidia/holoscan/lib/libgxf_core.so"
```

applications/h264/h264_video_decode/README.md
wyli: yes I see that now

applications/h264/h264_endoscopy_tool_tracking/python/metadata.json
wyli: ok, thanks for looking into this

applications/h264/h264_endoscopy_tool_tracking/README.md
```diff
@@ -25,16 +25,26 @@ can be specified in the 'h264_endoscopy_tool_tracking.yaml' file.
...
-# C++ version
 ./holohub run h264_endoscopy_tool_tracking --language cpp
+```
+
+### Python
+
+Separate build and run commands are required to address the known [symbol loading issue](../README.md#symbol-error-at-load).
+
+```bash
+# C++ version
```
wyli:
```suggestion
```

wyli (approved):
looks good to me, looks like the CLI misses some flags or field specifically for run and ignored during build

applications/h264/h264_endoscopy_tool_tracking/README.md
tbirdso: Thanks for catching this 👍

PR #1104, opened 2025-08-26
App: Update `object_detection_torchvision` to eagerly link torchvision symbols

tbirdso:
Background: Upcoming release Holoscan SDK v3.6.0 aims to remove torchvision from the
    release container. Torchvision is not strictly required for the
    Holoinfer Libtorch backend. However, a subsequent issue has been
    observed in running the object_detection_torch application.
    Investigation has revealed that torchvision symbols must be explicitly
    loaded by the executable in order to load the FRCNN ResNet 50 model and
    similar PyTorch models.
    
This change updates the object_detection_torch target to eagerly link in
    torchvision library symbols to ensure that the model load succeeds. Note that it is also required to link Python and Libtorch at build time in order to resolve torchvision symbols.
    
Verified the app runs with v3.5.0 and v3.6.0 with Xvfb on x86. Render test passes locally.

Also makes AJA source optional to reduce build time.

co-authored-by: Frank Lai <frlai@nvidia.com>
co-authored-by: Shekar Dwivedi <shekard@nvidia.com>

### DIFF
--- a/applications/object_detection_torch/CMakeLists.txt
+++ b/applications/object_detection_torch/CMakeLists.txt
@@ -46,6 +46,10 @@ if(holoscan_VERSION GREATER 3.5)
   set_target_properties(torchvision PROPERTIES
       IMPORTED_LOCATION "${TORCHVISION_LIBRARY}"
   )
+  target_link_libraries(torchvision INTERFACE
+    torch
+    Python3::Python
+  )
   add_library(torchvision::torchvision ALIAS torchvision)
 endif()
 
@@ -68,7 +72,7 @@ if(holoscan_VERSION GREATER 3.5)
   # Forcibly link torchvision and its required Python symbols to prevent ResNet runtime import failure
   target_link_libraries(object_detection_torch
     PRIVATE
-    ${TORCH_LIBRARIES}
+    torchvision::torchvision
     "-Wl,--no-as-needed $<TARGET_PROPERTY:Python3::Python,IMPORTED_LOCATION> $<TARGET_PROPERTY:torchvision::torchvision,IMPORTED_LOCATION> -Wl,--as-needed"
   )
 endif()


applications/object_detection_torch/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
-# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+..# SPDX-License-Identifier: Apache-2.0
```
agirault:
```suggestion
# SPDX-License-Identifier: Apache-2.0
```

applications/object_detection_torch/CMakeLists.txt
```diff
@@ -19,6 +19,36 @@ project(object_detection_torch CXX)
 find_package(holoscan 0.6 REQUIRED CONFIG
              PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")
 
+if(holoscan_VERSION GREATER 3.5)
+  # Torchvision is required for loading FRCNN ResNet 50 and similar models
+  # with the Holoinfer Libtorch backend.
+  find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
```
agirault, line 25: Need `Development` as well?

applications/object_detection_torch/CMakeLists.txt
tbirdso: Interesting side effect of the `check_copyright.py --update-current-year` script. cc @wyli  for viz, will fix manually, not a big deal

applications/object_detection_torch/CMakeLists.txt
tbirdso: Yes, see below where `Python3::Python` needs to be linked in order to link Torch and Torchvision symbols in the executable. Could update to Development.Embed.

Ref: https://cmake.org/cmake/help/latest/module/FindPython3.html#imported-targets

applications/object_detection_torch/CMakeLists.txt
```diff
@@ -19,6 +19,36 @@ project(object_detection_torch CXX)
...
+  )
+  find_library(TORCHVISION_LIBRARY
+    NAMES libtorchvision.so _C.so
+    PATHS ${TORCHVISION_SEARCH_DIRS}
+    PATH_SUFFIXES lib lib64
+    REQUIRED
+  )
+  add_library(torchvision SHARED IMPORTED)
+  set_target_properties(torchvision PROPERTIES
+      IMPORTED_LOCATION "${TORCHVISION_LIBRARY}"
```
agirault, line 47: nitpick: could add
```cmake
target_link_libraries(torchvision PRIVATE torch)
```
or 
```cmake
target_link_libraries(torchvision PRIVATE ${TORCH_LIBRARIES})
```
and skip it below

applications/object_detection_torch/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-# SPDX-FileCopyrightText: Copyright (c) 2022-2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
```
agirault, line 4:
```suggestion
# SPDX-License-Identifier: Apache-2.0
```

applications/object_detection_torch/Dockerfile
```diff
@@ -38,5 +38,12 @@ COPY utilities /tmp/scripts/utilities/
...
 
+FROM base as torch
+
 # Install libraries
 RUN pip install torch torchvision
+
+# Set up Holoscan SDK container libtorch to be found with ldconfig for app C++ build and runtime
+RUN echo $(python3 -c "import torch; print(torch.__path__[0])")/lib > /etc/ld.so.conf.d/libtorch.conf \
+    && ldconfig \
+    && ldconfig -p | grep -q "libtorch.so"
```
agirault, line 49: so this container has both torch from hsdk and from pip?

applications/object_detection_torch/Dockerfile
agirault: weird, code showed `..` before but not anymore.

tbirdso: I had pushed an update in https://github.com/nvidia-holoscan/holohub/pull/1104/commits/5c6c348a69622f325444d6e390d4f9d1f686fa60

applications/object_detection_torch/Dockerfile
tbirdso: Yes. PyTorch is required for model generation at app build time, libtorch alone is insufficient.

We could feasibly have two environments for build and run to avoid having both at runtime. IIRC that is on the roadmap but not yet supported smoothly by the CLI with mode presets.

applications/object_detection_torch/CMakeLists.txt
tbirdso: Wouldn't this need to be PUBLIC or INTERFACE? Will update, thanks

tbirdso: Updated

applications/object_detection_torch/CMakeLists.txt
agirault: @ronyrad did you mention seeing this as well?

applications/object_detection_torch/CMakeLists.txt
agirault: no PRIVATE is good since you don't expose headers or symbols

agirault (approved):

PR #1105, opened 2025-08-26
Adjust validation frames threshold

jjomier:
Enhance video frame validation by adding configurable pixel difference threshold. The `check_frames` function now accepts a `threshold` parameter, allowing users to specify the sensitivity of frame comparison. 

Updated command-line argument parsing to include the new threshold option with a default value of 0.5%


tbirdso (approved):
LGTM!

PR #1112, opened 2025-08-28
Refactor CMake configuration for colonoscopy segmentation tests

jjomier:
- Removed the unused `find_package(holoscan)` line from the CMakeLists.txt file for the colonoscopy segmentation application.
- Updated the `holohub.container.ctest` file by removing the CUDA compiler option. Applications should set the NVCC compiler in the CMake files.

### DIFF
--- a/applications/colonoscopy_segmentation/CMakeLists.txt
+++ b/applications/colonoscopy_segmentation/CMakeLists.txt
@@ -61,9 +61,6 @@ if(BUILD_TESTING)
     DEPENDS "colonoscopy_segmentation_test.py"
   )
 
-  # Get GXF_LIB_DIR for Python tests
-  find_package(holoscan)
-
   # Add test
   add_test(NAME colonoscopy_segmentation_python_test
            COMMAND python3 ${CMAKE_CURRENT_BINARY_DIR}/colonoscopy_segmentation_test.py
--- a/utilities/testing/holohub.container.ctest
+++ b/utilities/testing/holohub.container.ctest
@@ -48,7 +48,6 @@ ctest_read_custom_files("${CTEST_BINARY_DIRECTORY}")
 set(configure_options
   ${CONFIGURE_OPTIONS}
   -D HOLOHUB_DATA_DIR=/workspace/holohub/data-${APP}-${TAG}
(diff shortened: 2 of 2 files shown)


wyli (approved):

applications/colonoscopy_segmentation/CMakeLists.txt
```diff
@@ -61,9 +61,6 @@ if(BUILD_TESTING)
     DEPENDS "colonoscopy_segmentation_test.py"
   )
 
-  # Get GXF_LIB_DIR for Python tests
-  find_package(holoscan)
```
tbirdso, line 65: This was NOT unused, the reason is in the comment, removing this will result in a test failure. Please see git blame for commit description

tbirdso: @jjomier can you please revert?

wyli: please see: https://github.com/nvidia-holoscan/holohub/pull/1113

jjomier: I'm confused how the test would work since HSDK was not found by CMake:
`CMake Warning at applications/colonoscopy_segmentation/CMakeLists.txt:65 (find_package):
  By not providing "Findholoscan.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "holoscan",
  but CMake did not find one.

  Could not find a package configuration file provided by "holoscan" with any
  of the following names:

    holoscanConfig.cmake
    holoscan-config.cmake

  Add the installation prefix of "holoscan" to CMAKE_PREFIX_PATH or set
  "holoscan_DIR" to a directory containing one of the above files.  If
  "holoscan" provides a separate development package or SDK, be sure it has
  been installed.`

tbirdso: Discussed with @jjomier  offline, our dashboard indeed shows the test passing despite this `find_package(holoscan)` failing, which is curious.

Let's do a quick test to see what happens with the current test behavior, if it is passing we can remove the GXF_LIB_DIR reference, otherwise I'll add a quick patch to depend on the Holoscan Python wheel instead of the Holoscan C++ package for the library reference. 👍

PR #1113, opened 2025-08-28
Revert "Refactor CMake configuration for colonoscopy segmentation tests"

wyli:
Reverts nvidia-holoscan/holohub#1112

### DIFF
--- a/utilities/testing/holohub.container.ctest
+++ b/utilities/testing/holohub.container.ctest
@@ -48,7 +48,6 @@ ctest_read_custom_files("${CTEST_BINARY_DIRECTORY}")
 set(configure_options
   ${CONFIGURE_OPTIONS}
   -D HOLOHUB_DATA_DIR=/workspace/holohub/data-${APP}-${TAG}
-  -D CMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc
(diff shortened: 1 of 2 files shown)


utilities/testing/holohub.container.ctest
```diff
@@ -48,6 +48,7 @@ ctest_read_custom_files("${CTEST_BINARY_DIRECTORY}")
 set(configure_options
   ${CONFIGURE_OPTIONS}
   -D HOLOHUB_DATA_DIR=/workspace/holohub/data-${APP}-${TAG}
+  -D CMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc
```
tbirdso: I think we can remove this line, just need the `find_package(holoscan)` above to be kept.

wyli:
```suggestion
```

tbirdso:
Looks good to me, but would like approval from @jjomier to confirm

tbirdso:
After investigation @jjomier  was correct, the GXF_LIB_DIR path was extraneous and probably left over from before the Holoscan Python wheel was installed rather than relying on a container HSDK installation

Suggest we close this PR in favor of https://github.com/nvidia-holoscan/holohub/pull/1114

PR #1114, opened 2025-08-28
App: Remove extraneous "colonoscopy_segmentation" test PYTHONPATH

tbirdso:
Followup to e86a13e removing GXF_LIB_DIR from test PYTHONPATH. As of f9d44deb the Holoscan Python package is installed in the Python dist-packages path at container build time, so we do not need to set PYTHONPATH to find it.

The app with replayer preset builds no Python operators, so the HoloHub Python module is not required.

### DIFF
--- a/applications/colonoscopy_segmentation/CMakeLists.txt
+++ b/applications/colonoscopy_segmentation/CMakeLists.txt
@@ -67,8 +67,8 @@ if(BUILD_TESTING)
                    --config ${CMAKE_CURRENT_BINARY_DIR}/colonoscopy_segmentation_testing.yaml
                    --data "${HOLOHUB_DATA_DIR}/colonoscopy_segmentation"
            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
-  set_property(TEST colonoscopy_segmentation_python_test PROPERTY ENVIRONMENT
-               "PYTHONPATH=${GXF_LIB_DIR}/../python/lib:${CMAKE_BINARY_DIR}/python/lib")
+  set_property(TEST colonoscopy_segmentation_python_test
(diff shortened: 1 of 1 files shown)


applications/colonoscopy_segmentation/CMakeLists.txt
```diff
@@ -67,8 +67,8 @@ if(BUILD_TESTING)
                    --config ${CMAKE_CURRENT_BINARY_DIR}/colonoscopy_segmentation_testing.yaml
                    --data "${HOLOHUB_DATA_DIR}/colonoscopy_segmentation"
            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
-  set_property(TEST colonoscopy_segmentation_python_test PROPERTY ENVIRONMENT
-               "PYTHONPATH=${GXF_LIB_DIR}/../python/lib:${CMAKE_BINARY_DIR}/python/lib")
+  set_property(TEST colonoscopy_segmentation_python_test
+                PROPERTY ENVIRONMENT PYTHONPATH="")
```
jjomier, line 71: Maybe we can just remove these lines?

tbirdso: Will the test use the environment PYTHONPATH if we remove this line?

Slightly better practice to ensure that the container PYTHONPATH is not affecting results, but not a real concern either way as we expect `sys.path` to already have the Holoscan Python installation

tbirdso: Did a quick test and confirmed that if PYTHONPATH is not set in test properties then it is picked up from the environment. Not much risk here in `colonoscopy_segmentation` due to the finetuned container environment, but let's keep the empty PYTHONPATH as best practice for controlling the test environment.

PR #1117, opened 2025-09-02
Fix AI Surgical Workflow AJA Format Issue

bhashemian:
This pull request updates the video pipeline for AJA capture card sources to ensure compatibility by dropping the alpha channel from video frames. It introduces a new `FormatConverterOp` to convert from `rgba8888` to `rgb888` format, and updates the flow and configuration accordingly.

AJA source compatibility improvements:

* Changed the input data type for AJA sources from `rgba8888` to `rgb888` in the `compose` method, and added a `FormatConverterOp` named `drop_alpha_channel` to convert and drop the alpha channel for downstream compatibility (`ai_surgical_video.py`).
* Updated the dynamic flow logic to insert the `drop_alpha_channel` operation between the AJA source and the main processing source, ensuring the correct data format is used (`ai_surgical_video.py`).

Configuration updates:

* Added a new `drop_alpha_channel` configuration section to `config.yaml` to specify the format conversion from `rgba8888` to `rgb888` for the `FormatConverterOp`.

PR #1121, opened 2025-09-03
Fix AI Surgical Workflow AJA Format Converter Naming

bhashemian:
Followup fix for #1117.


bhashemian:
@tbirdso, I missed addressing your comment in PR #1117 because it was merged automatically. This PR is to fix that.

PR #1124, opened 2025-09-05
Update Orsi `FormatConverterOp` for CUDA 13 with Backward Compatibility

bhashemian:
This pull request updates the `FormatConverterOp` implementation in `format_converter.cpp` to support CUDA 13.0 with backward compatibility for earlier versions.

**CUDA error handling and NPP integration:**

* Replaces the custom `CUDA_TRY` macro with standardized error handling using `HOLOSCAN_CUDA_CALL_THROW_ERROR` from `holoscan/utils/cuda_macros.hpp`, ensuring consistent and informative error reporting for CUDA API calls. 

**CUDA 13+ compatibility:**

* Adds a conditional initialization for the NPP stream context in `initialize()` to manually populate context fields when running on CUDA 13.0 or newer, addressing changes in CUDA's stream context handling.

**Build configuration updates:**

* Updates the CUDA architecture targets for both segmentation preprocessor and postprocessor operators to only build for architecture 80 (Ampere), removing support for architecture 70 (Volta). (`orsi_segmentation_preprocessor/CMakeLists.txt`, `orsi_segmentation_postprocessor/CMakeLists.txt`).

### DIFF
--- a/operators/orsi/orsi_segmentation_postprocessor/CMakeLists.txt
+++ b/operators/orsi/orsi_segmentation_postprocessor/CMakeLists.txt
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/operators/orsi/orsi_segmentation_postprocessor/CMakeLists.txt
+++ b/operators/orsi/orsi_segmentation_postprocessor/CMakeLists.txt
@@ -28,7 +28,7 @@ add_library(${PROJECT_NAME}
   segmentation_postprocessor.cu
 )
 
-set_target_properties(${PROJECT_NAME}  PROPERTIES CUDA_ARCHITECTURES "70;80")
+set_target_properties(${PROJECT_NAME}  PROPERTIES CUDA_ARCHITECTURES "80")
 
 add_library(holoscan::orsi::segmentation_postprocessor ALIAS ${PROJECT_NAME})
 target_include_directories(${PROJECT_NAME} INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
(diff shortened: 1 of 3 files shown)


operators/orsi/orsi_segmentation_postprocessor/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ add_library(${PROJECT_NAME}
   segmentation_postprocessor.cu
 )
 
-set_target_properties(${PROJECT_NAME}  PROPERTIES CUDA_ARCHITECTURES "70;80")
+set_target_properties(${PROJECT_NAME}  PROPERTIES CUDA_ARCHITECTURES "80")
```
tbirdso: Looks like CUDA Toolkit 13.0 should still support >= 7.5, any concerns about keeping that min version (Quadro RTX compatibility) for now?

```suggestion
set_target_properties(${PROJECT_NAME}  PROPERTIES CUDA_ARCHITECTURES "75;80")
```

operators/orsi/orsi_segmentation_preprocessor/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ add_library(${PROJECT_NAME}
   segmentation_preprocessor.cu
 )
 
-set_target_properties(${PROJECT_NAME}  PROPERTIES CUDA_ARCHITECTURES "70;80")
+set_target_properties(${PROJECT_NAME}  PROPERTIES CUDA_ARCHITECTURES "80")
```
tbirdso:
```suggestion
set_target_properties(${PROJECT_NAME}  PROPERTIES CUDA_ARCHITECTURES "75;80")
```

tbirdso (commented):
Thank you! One question about supported compute capabilities

operators/orsi/orsi_segmentation_postprocessor/CMakeLists.txt
tbirdso: Also, if the Orsi apps are intended to always run on the same system on which they are built, then this target may be able to drop the `CUDA_ARCHITECTURES` altogether and use the [default HoloHub "native" architecture](https://github.com/nvidia-holoscan/holohub/blob/main/CMakeLists.txt#L74) to build for the current GPU architecture.

bhashemian: Thanks, Tom! That's a fair point! I was not sure about Orsi's intention but it makes sense to remove it altogether. I'll make the change and test it.

tbirdso (approved):
Looks good! Please confirm when you're able to verify behavior with default `CUDA_ARCHITECTURES`

PR #1135, opened 2025-09-19
[DRAFT] [DON'T MERGE] Fix dependency with AJA Operator and CUDA stream in Multi AI ultrasound application 

ichergui:


applications/ultrasound_segmentation/python/ultrasound_segmentation.py
```diff
@@ -72,7 +70,11 @@ def compose(self):
...
         )
 
-        is_aja = self.source.lower() == "aja"
+        is_aja = False
+        if self.source.lower() == "aja":
+            from holohub.aja_source import AJASourceOp
+
+            is_aja = True
         if is_aja:
             source = AJASourceOp(self, name="aja", **self.kwargs("aja"))
```
bhashemian: I think this is a simpler way to achieve the same, right? Unless, you want to try/catch if the import works and give a more detailed error message or continue without AJA, which I don't recommend.

```suggestion
        is_aja = self.source.lower() == "aja"
        if is_aja:
            from holohub.aja_source import AJASourceOp
            source = AJASourceOp(self, name="aja", **self.kwargs("aja"))
```

operators/CMakeLists.txt
```diff
@@ -13,9 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+option(HOLOHUB_BUILD_AJA_OPERATOR "Build support for AJA operator" ON)
```
tbirdso: This should already be available from `add_holohub_operator` with the variable `-DOP_aja_source:BOOL=<ON/OFF>`

```suggestion
```

operators/CMakeLists.txt
```diff
@@ -13,9 +13,13 @@
...
+
 # Add operators (in alphabetical order)
 add_holohub_operator(advanced_network)
-add_holohub_operator(aja_source)
+if(HOLOHUB_BUILD_AJA_OPERATOR)
+  add_holohub_operator(aja_source)
+endif()
```
tbirdso: Conditional is redundant with `add_holohub_operator`

```suggestion
add_holohub_operator(aja_source)
```

applications/ultrasound_segmentation/cpp/CMakeLists.txt
```diff
@@ -31,9 +31,10 @@ target_link_libraries(ultrasound_segmentation
    holoscan::ops::inference
    holoscan::ops::segmentation_postprocessor
    holoscan::ops::holoviz
-   holoscan::aja
 )
 
+target_link_libraries(ultrasound_segmentation PRIVATE $<TARGET_NAME_IF_EXISTS:holoscan::aja>)
```
tbirdso: Could simplify by doing this in the same `target_link_libraries` call above with other dependencies
```suggestion
  $<TARGET_NAME_IF_EXISTS:holoscan::aja>
)
```

Here and in other CMakeLists.txt updates throughout

applications/ultrasound_segmentation/cpp/CMakeLists.txt
```diff
@@ -84,9 +85,10 @@ if(BUILD_TESTING)
     holoscan::ops::inference
     holoscan::ops::segmentation_postprocessor
     holoscan::ops::holoviz
-    holoscan::aja
   )
 
+  target_link_libraries(ultrasound_segmentation_cpp_test PRIVATE $<TARGET_NAME_IF_EXISTS:holoscan::aja>)
```
tbirdso: Could simplify (nitpick)

```suggestion
  $<TARGET_NAME_IF_EXISTS:holoscan::aja>
  )
```

tbirdso (changes requested):
Thanks @ichergui ! One concern about `operators/CMakeLists.txt` redundant updates

applications/ultrasound_segmentation/python/ultrasound_segmentation.py
ichergui: Hi @bhashemian 
Thanks for the review. You are right.
I will pushthe updates shortly

operators/CMakeLists.txt
ichergui: Thanks @tbirdso 
I will make changes and push them shortly

applications/ultrasound_segmentation/cpp/CMakeLists.txt
ichergui: I wanted to coherent with the other application. 
I will make the changes and push them

bhashemian (approved):
Thanks @ichergui! LGTM!

PR #1137, opened 2025-09-22
CLI: configurable properties for external usage

wyli:
This is to make various hard-coded properties such as container names, versions configurable. Main benefit is that the external repo can leverage the CLI scripts with flexible project-specific info.


utilities/cli/container.py
```diff
@@ -54,20 +52,53 @@ class HoloHubContainer:
...
+    # Docker and runtime configuration
+    DOCKER_EXE = os.environ.get("HOLOHUB_DOCKER_EXE", "docker")  # Docker executable
+
+    # SDK and path configuration
+    SDK_PATH = os.environ.get("HOLOHUB_SDK_PATH", "/opt/nvidia/holoscan")
+    BASE_SDK_VERSION = os.environ.get("HOLOHUB_BASE_SDK_VERSION", "3.6.0")
+    BENCHMARKING_SUBDIR = os.environ.get(
+        "HOLOHUB_BENCHMARKING_SUBDIR", "benchmarks/holoscan_flow_benchmarking"
+    )
+    DEFAULT_DOCKERFILE = os.environ.get("HOLOHUB_DEFAULT_DOCKERFILE", "Dockerfile")
```
tbirdso: Would an absolute path rather than relative be useful here?

wyli:
> Can we please document these settings in a markdown table somewhere? Would also be good to add to `./holohhub list-env` output for debugging

yes, those are printed in the utils https://github.com/nvidia-holoscan/holohub/blob/5120fd62d30da702f383ef7656963394de0d9e49/utilities/cli/util.py#L1111-L1116

PR #1138, opened 2025-09-23
Update H.264 application with MM 1.5 and CUDA 13 support

mocsharp:
- Added a check in the Dockerfile to enforce CUDA version 13 or higher for H.264 applications starting from Holoscan SDK 3.6.0.
- Updated DeepStream dependencies URL in install_dependencies.sh to version 5.1.0.
- Bumped the version of the GXF MM extensions and CUDA in install_dependencies.sh to 1.5.0 and 13.0, respectively.
- Updated CMakeLists.txt files for C++ and Python projects to require Holoscan version 3.6.1.

These changes ensure compatibility with the latest SDK and improve dependency management.

### DIFF
--- a/applications/h264/Dockerfile
+++ b/applications/h264/Dockerfile
@@ -76,4 +76,3 @@ WORKDIR /
 
 RUN /install_dependencies.sh
 
-ENV LD_PRELOAD=/opt/nvidia/holoscan/lib/libgxf_core.so
--- a/applications/h264/Dockerfile
+++ b/applications/h264/Dockerfile
@@ -25,6 +25,16 @@ FROM ${BASE_IMAGE} as base
 
 ARG DEBIAN_FRONTEND=noninteractive
 
+# Starting from Holoscan SDK 3.6.0, only CUDA 13 is supported.
+# To run H.264 application with CUDA 12, please use tags with holoscan-sdk-3.5.0 or earlier.
+RUN CUDA_VERSION_MAJOR=$(echo $CUDA_VERSION | cut -d '.' -f 1) \
+    && if [ "$CUDA_VERSION_MAJOR" != "13" ]; then \
+        echo "CUDA version $CUDA_VERSION is not supported."; \
+        echo "Starting from Holoscan SDK 3.6.0, H.264 applications only support CUDA 13 or higher."; \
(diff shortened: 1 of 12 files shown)


applications/h264/Dockerfile
```diff
@@ -65,3 +75,5 @@ WORKDIR /
 #ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/lib/aarch64-linux-gnu/tegra/
 
 RUN /install_dependencies.sh
+
+ENV LD_PRELOAD=/opt/nvidia/holoscan/lib/libgxf_core.so
```
tbirdso: We tried setting this in the environment in the past but it impeded data downloads, suggest removing and sticking with the README documented workaround for now

Ref: https://github.com/nvidia-holoscan/holohub/pull/1103/commits/744cc99ad448f80e348eedbfe7f50fb35cb83a81

mocsharp: Thanks @tbirdso. It would be nice if we can include environment variables in the metadata.json section.

mocsharp: I've made a change to the holohub CLI to read env from metadata:

```json
    "run": {
      "command": "python3 <holohub_app_source>/h264_endoscopy_tool_tracking.py --data <holohub_data_dir>/endoscopy",
      "workdir": "holohub_bin",
      "env": {
        "LD_PRELOAD": "/opt/nvidia/holoscan/lib/libgxf_core.so"
      }
    }
```
This would allow users to run the python apps using Holohub CLI, otherwise, they would have to enter the container and run the app manually.

tbirdso: Thanks @mocsharp , looks good, added @wyli  for comment

wyli: thanks, I think it's good to add the flexibility. Not a blocker, but a minor concern: since the CLI supports the `--local` flag, the `run` section might execute either inside a container or without a container. The absolute path `/opt/nvidia/holoscan...` may not always exist. I've thought about adding placeholders such as
```
"env": {
        "LD_PRELOAD": "<holoscan_src>/lib/libgxf_core.so"
}
```
but probably we don't need that complexity for now...

wyli (approved):

PR #1141, opened 2025-09-25
Move app.run() calls to within a main() function so that Application and operator deconstructors reliably get called before the Python process terminates.

ghost:
The `app` object had not been destroyed resulting in missing cleanup. When moving to a function `app` is destroyed before Python terminates.


tbirdso (approved):
Looks good, thank you!

PR #1144, opened 2025-09-28
App: Fix missing object_detection_torch header

tbirdso:
Apply fix for upcoming Holoscan SDK v3.7, which moves the "holoscan::inference::node_type" definition from the "holoscan::infer" target into the "holoscan::holoinfer_util" target

Backwards compatible with <=v3.6

Addresses error:
```
[object_detection_torch/main.cpp:130](https://gitlab-master.nvidia.com/clara-holoscan/clara-holoscan-sdk/blob/09a0250c71581fd51a3f8e4c872a441e40ab1b78/object_detection_torch/main.cpp#L130):75: error: 'node_type' is not a member of 'holoscan::inference'
  130 |     auto configuration = config["generate_boxes"].as<holoscan::inference::node_type>();
      |                                                                           ^~~~~~~~~
```

cc @shekhardw

### DIFF
--- a/applications/object_detection_torch/CMakeLists.txt
+++ b/applications/object_detection_torch/CMakeLists.txt
@@ -66,6 +66,8 @@ target_link_libraries(object_detection_torch
     holoscan::ops::inference
     holoscan::ops::inference_processor
     holoscan::ops::holoviz
+    holoscan::infer
+    $<TARGET_NAME_IF_EXISTS:holoscan::holoinfer_utils>
     $<TARGET_NAME_IF_EXISTS:holoscan::aja>
 )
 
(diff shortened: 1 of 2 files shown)


applications/object_detection_torch/CMakeLists.txt
```diff
@@ -66,6 +66,8 @@ target_link_libraries(object_detection_torch
     holoscan::ops::inference
     holoscan::ops::inference_processor
     holoscan::ops::holoviz
+    holoscan::infer
+    $<TARGET_NAME_IF_EXISTS:holoscan::holoinfer_utils>
```
agirault: switching to the same logic as multiai would be nice for consistency

tbirdso: Updated

PR #1129, opened 2025-09-10
Enhance video streaming tests fixed

cdinea:
This PR implements unit test suites for the streaming_client and streaming_server Holoscan operators.

Update Note: This PR has been synchronized with recent changes to the streaming client and server streaming apps.
The integration test methodology employs the identical workflow and deployment documented in the application README for running both the streaming server and streaming client apps.

**Preconditions Related to Infrastructure**

Software Dependencies:

- NGC CLI for downloading streaming client binaries
- StreamingClient operator and StreamingServer libraries from NGC (libStreamingClient.so, libNvStreamingSession.so, etc.)
- Video data files: surgical_video.gxf_index and surgical_video.gxf_entities

# StreamingClient Operator Testing

## Prerequisites

⚠️ **IMPORTANT**: Before running any tests, you must download the required NGC resources as specified in the main [README.md](README.md).

### Download Required NGC Resources

The StreamingClient operator requires downloading the Holoscan Client Cloud Streaming library from NGC:

```bash
# Download using NGC CLI
cd <your_holohub_path>/operators/streaming_client
ngc registry resource download-version nvidia/holoscan_client_cloud_streaming:0.1
unzip -o holoscan_client_cloud_streaming_v0.1/holoscan_client_cloud_streaming.zip

# Copy the appropriate architecture libraries to lib/ directory
# For x86_64 systems:
cp lib/x86_64/* lib/
# For aarch64 systems:
# cp lib/aarch64/* lib/

# Clean up architecture-specific directories and NGC download directory
rm -rf lib/x86_64 lib/aarch64
rm -rf holoscan_client_cloud_streaming_v0.1
```

**NGC Resource URL**: https://catalog.ngc.nvidia.com/orgs/nvidia/resources/holoscan_client_cloud_streaming

📋 **Note**: Tests may fail or behave unexpectedly if these dependencies are not properly installed.

## Automatic Test Data Download

The streaming client tests automatically download surgical video data for realistic testing:

- **Video Data**: Endoscopy sample data (854x480, 30fps) is automatically downloaded from NGC
- **GXF Entities**: Video files are converted to GXF entities during the build process
- **Test Data Location**: Data is placed in `${CMAKE_BINARY_DIR}/data/endoscopy/`
- **Symlinks Created**: `surgical_video.gxf_entities` and `surgical_video.gxf_index` are symlinked for easy access
- **Self-Sufficient Tests**: No manual data preparation required - tests download what they need

[...]

PR #1147, opened 2025-09-30
Enhanced unified video streaming app + integration testing

cdinea:
[to be added]


applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required(VERSION 3.20)
+project(video_streaming_demo_enhanced LANGUAGES NONE)
+
+# Check if this application should be built
+if(NOT APP_video_streaming_demo_enhanced)
+  return()
+endif()
```
tbirdso, line 23: Please use the standard HoloHub `add_holohub_application` function rather than re-implementing the app check here

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+               sleep 30
+
+               # Check if client is still running (indicating successful connection)
+               if kill -0 \$CLIENT_PID 2>/dev/null; then
+                 echo '✓ Client is still running - connection successful'
+                 CLIENT_SUCCESS=1
+
+                 # Stop client gracefully
+                 echo 'Stopping client...'
+                 kill -TERM \$CLIENT_PID 2>/dev/null || true
```
tbirdso: Is this the generally recommended user approach to stop the client process?

applications/video_streaming_demo_enhanced/Dockerfile
```diff
@@ -75,9 +97,8 @@ WORKDIR /workspace/holohub
 
 # Set environment variables
 ENV HOLOSCAN_INPUT_PATH=/workspace/data/endoscopy
-
-# Enable autocomplete
-RUN echo "source /etc/bash_completion" >> ~/.bashrc
+ENV PYTHONPATH=/workspace/holohub/applications/streaming_server_demo/python:${PYTHONPATH}
```
tbirdso: I don't see this path in the PR, remove?
```suggestion
```

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+               echo '✗ Integration test FAILED'
+               exit 1
+             fi
+           "
+           WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+
+  set_tests_properties(video_streaming_integration_test PROPERTIES
+                       TIMEOUT 300
+                       PASS_REGULAR_EXPRESSION "Integration test PASSED"
+                       FAIL_REGULAR_EXPRESSION "Integration test FAILED")
```
tbirdso, line 285: This is on the right track. Please add further regex checks to ensure that each step of expected server behavior ran. If only the final output string is checked, it might miss instances where the return code is correct but server behavior did not match expectations.

We recently learned that PASS_REGULAR_EXPRESSION and FAIL_REGULAR_EXPRESSION are not very good at handling multiple test criteria. @grlee77  has added a test validation script I would recommend using here. The script will be published in Holoscan SDK v3.7, in the meantime please see the internal link: https://gitlab-master.nvidia.com/holoscan/holoscan-sdk/-/blob/main/public/scripts/test_pattern_validation.py

Log regex checks should verify that each application
1. Initialized
2. Connected
3. Streamed data
4. Exited gracefully

For example, if a log message such as  ` HOLOSCAN_LOG_INFO("Using resource width: {}", resource_config.width);` always runs when the server initializes, and we know the test image width is 920, we should check for the string `Using resource width: 920` in output and the test should fail if it is not found.

Could you please indicate what are some log messages we should always see if the test runs successfully? See related [previous discussion](https://github.com/nvidia-holoscan/holohub/pull/1129#discussion_r2376850907).

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,156 @@
...
+               CLIENT_PID=\$!
+
+               # Let client stream for 30 seconds
+               echo 'Letting client stream for 30 seconds...'
+               sleep 30
+
+               # Check if client is still running (indicating successful connection)
+               if kill -0 \$CLIENT_PID 2>/dev/null; then
+                 echo '✓ Client is still running - connection successful'
+                 CLIENT_SUCCESS=1
```
tbirdso: I don't necessarily agree that the application running is a good indicator of successful streaming. Applications can stall on exit for a number of reasons.

We should verify via log messages and visual output frame checks that streaming succeeds.

applications/video_streaming_demo_enhanced/README.md
```diff
@@ -0,0 +1,670 @@
...
+
+# Clean up
+rm -rf holoscan_server_cloud_streaming_v0.2
+```
+
+## Running the Applications
+
+The unified application provides both client and server as separate components:
+
+**⚠️ Important: Enhanced applications require Holoscan SDK 3.5.0. Use the base image parameter to ensure compatibility.**
```
tbirdso: Can we test with latest (v3.7.0)?

If there are known issues, can we please document them in the README?

applications/video_streaming_demo_enhanced/README.md
```diff
@@ -0,0 +1,670 @@
...
+
+### 1. Start the Streaming Server
+
+```bash
+# From holohub root directory - with Holoscan 3.5.0 base image
+./holohub run --docker-opts='-e EnableHybridMode=1' --base-img=nvcr.io/nvidia/clara-holoscan/holoscan:v3.5.0-dgpu video_streaming_demo_enhanced server --language cpp
```
tbirdso: Seems like the Dockerfile now has `ENV EnableHybridMode=1`, can we remove the explicit parameter here?

applications/video_streaming_demo_enhanced/README.md
```diff
@@ -0,0 +1,670 @@
...
+**Analyzing the log:**
+```bash
+# Check test status
+grep "Integration test PASSED" integration_test.log
+
+# Check frame counts
+grep "CLIENT: Received frame" integration_test.log | tail -5
+
+# Check for errors
+grep -i "error\|fail\|crash" integration_test.log
```
tbirdso, line 541: We should include these checks in automatic test output checks (PASS_REGULAR_EXPRESSION, etc)

applications/video_streaming_demo_enhanced/README.md
```diff
@@ -0,0 +1,670 @@
...
+
+### Continuous Integration
+
+The integration test is designed for CI/CD pipelines:
+
+```bash
+# CI-friendly command with timeout and exit codes
+timeout 300 ./applications/video_streaming_demo_enhanced/integration_test.sh
+echo "Integration test exit code: $?"
+```
```
tbirdso: We should use `./holohub test` for CI/CD, similar to:
```
./holohub test integration_test
```

tbirdso (commented):
Initial review pass focused on integration test design, please see feedback

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: thank you for the feedback @tbirdso - I believe I ran into build errors when trying to use add_holohub_application , would you consider this to be a P0 item and needs revisiting?

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: thank you @tbirdso - not sure how else we could stop a client process to stop running in a test environment.

applications/video_streaming_demo_enhanced/Dockerfile
cdinea: thank you @tbirdso - PYTHONPATH deleted, already set in a previous line

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: thank you for the feedback @tbirdso - it is a great idea to have the test pattern validator , I took note on an internal JIRA item to address this in a subsequent PR.

> If only the final output string is checked, it might miss instances where the return code is correct but server behavior did not match expectations.

Given the way the Streaming Server and the Streaming Client work, I believe this is partially valid but not entirely accurate. Let me break down what the test actually does:

- Process Health (lines 89-102): Uses kill -0 to verify the server process is alive and stable
- Client Connection (lines 117-129): Verifies the client process stays running for 30 seconds (indicating successful connection)
- Exit Codes (lines 144, 147): Returns 0 for success, 1 for failure
- Output Strings (lines 152-155): CMake's PASS_REGULAR_EXPRESSION and FAIL_REGULAR_EXPRESSION match against output

> Could you please indicate what are some log messages we should always see if the test runs successfully? See related https://github.com/nvidia-holoscan/holohub/pull/1129#discussion_r2376850907.

Please refer to the detailed description in the [README](https://github.com/nvidia-holoscan/holohub/blob/e5460621b3c9965ef328e99017e91f7dfe989b86/applications/video_streaming_demo_enhanced/README.md#integration-testing)

applications/video_streaming_demo_enhanced/README.md
cdinea: hi @tbirdso -  I have tested with Holoscan 3.6 and encountered ABI compatibility issues. The Streaming Client and Server components include bundled libraries (libcudart and others) that conflict with Holoscan 3.6's runtime dependencies. While I have not validated against version 3.7, similar compatibility challenges are likely to persist due to the fundamental ABI mismatch.
Resolving this would require recompiling the entire Streaming SDK stack against the target Holoscan version. Given the scope and timeline constraints, I recommend maintaining compatibility with Holoscan 3.5 for this release cycle. I will document this version requirement and the underlying technical constraints in README

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: Added more detailed testing against log messages in [this](https://github.com/nvidia-holoscan/holohub/pull/1147/commits/0ca9707730dc9402edded04de099c42bf602d8d9) commit

applications/video_streaming_demo_enhanced/README.md
cdinea: hi @tbirdso - based on the integration test we outlined, my understanding is that only the apps will be run with ./holohub test

`# launch server
./holohub test --docker-opts='-e EnableHybridMode=1' streaming_server_demo_04_80_tensor_streaming --language cpp 2>&1 > streamignserver.log &  # here I am using '>' instead of '| tee' to not showing anything in the terminal

sleep 10

./holohub test --docker-opts='-e EnableHybridMode=1 --device /dev/video0' streaming_client_demo_04_80_streaming --language cpp --config streaming_client_demo.yaml 2>&1 > streamingclient.log &  # here I am using '>' instead of '| tee' to not showing anything in the terminal

sleep 30

# wait for both server and client to terminates (timeout = 10 secs)
wait -t 10

kill -9 %1 %2

# check the log files
cat streamingserver.log
cat streamingclient.log`


I need to scope out what it takes for ./holohub test to be able to run nested tests with ./holohub test

tbirdso: I'm confused. Initial design discussion indeed had two calls to `./holohub test`, but the implementation you've prepared here only requires a single call.

The procedure you've shared above matches the design you shared previously, where `./holohub test` is called twice to set up the client and server separately.

The implementation in this PR is a different design, where a single `./holohub test` call and CTest entry is responsible for launching both applications.

I.e. I see the following test implementation in the current PR:
```
# Add integration test for video streaming demo
  add_test(NAME video_streaming_integration_test
           COMMAND bash -c "...
# This launches the server executable directly
./applications/video_streaming_demo_enhanced/video_streaming_demo_server/cpp/streaming_server_demo_enhanced --config applications/video_streaming_demo_enhanced/video_streaming_demo_server/cpp/streaming_server_demo.yaml > \$SERVER_LOG 2>&1 &
...
# This launches the client executable directly
./applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/streaming_client_demo_enhanced --config applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/streaming_client_demo_replayer.yaml > \$CLIENT_LOG 2>&1 &
...
```

I would suggest sticking with the current implementation, one `./holohub test` call that runs the full test with the client and server together is preferable to two separate calls to `./holohub test`. I do not see a need for nested `./holohub test` calls here.

applications/video_streaming_demo_enhanced/README.md
tbirdso: Got it, thanks for the explanation. Agreed that documenting and pinning to 3.5.0 is sufficient.

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+
+             echo ''
+             echo '=== Test Results Summary ==='
+             echo \"Server checks passed: \$SERVER_SUCCESS/4 (minimum required: 3)\"
+             echo \"Client checks passed: \$CLIENT_SUCCESS/3 (minimum required: 2)\"
+
+             # Determine overall success
+             # Require at least 3/4 server checks and 2/3 client checks
+             if [ \$SERVER_SUCCESS -ge 3 ] && [ \$CLIENT_SUCCESS -ge 2 ]; then
+               STREAMING_SUCCESS=1
```
tbirdso: What is the justification for these minimum bounds on acceptance?

It seems like we should always expect 100% pass, anything less is a test failure.

Also, I count 5 server checks.. the implementation with hardcoded constants here will not scale if new checks are added and not accounted for on this line. It seems like the current implementation claims 80% minimum bound for the server but is actually 60%.

Please update the test logic so that the existence of _any_ failure will result in an overall test failure.

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+               echo '✓ Server: Downstream connection established'
+               SERVER_SUCCESS=\$((SERVER_SUCCESS + 1))
+             else
+               echo '✗ Server: Downstream connection not established'
+             fi
+             
+             # Check for frame processing
+             if grep -q 'Processing UNIQUE frame' \$SERVER_LOG; then
+               FRAME_COUNT=\$(grep -c 'Processing UNIQUE frame' \$SERVER_LOG)
+               echo \"✓ Server: Processed \$FRAME_COUNT unique frames\"
```
tbirdso: What is the expected lower and upper bound on the frame count? Current implementation could pass with 0 unique frames, which seems not expected

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+               SERVER_SUCCESS=\$((SERVER_SUCCESS + 1))
+             else
+               echo '✗ Server: No frames processed'
+             fi
+             
+             # Check for frame stats (optional but good indicator)
+             if grep -q 'Frame Processing Stats' \$SERVER_LOG; then
+               echo '✓ Server: Frame processing statistics logged'
+               SERVER_SUCCESS=\$((SERVER_SUCCESS + 1))
+             fi
```
tbirdso: The integration test should not have optional checks, we should have clear expectations and acceptance criteria for deterministic test behavior. Please clarify whether or not this is expected output for the integration test, and if the output is expected then please make the check required for the test to pass.

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
```diff
@@ -0,0 +1,1601 @@
...
+    }
+    shape_str += "]";
+    HOLOSCAN_LOG_INFO("Tensor shape={}", shape_str);
+
+    // 🔧 ENHANCED DEBUG: Write tensor to disk before processing - standardized to every 10th frame
+    debug_frame_counter++;
+    if (debug_frame_counter % 10 == 0) {  // Save every 10th frame to match server frequency
+      HOLOSCAN_LOG_INFO("💾 DEBUG: Writing input tensor to disk (frame {})", debug_frame_counter);
+      writeTensorToDisk(tensor, "client_input_tensor", debug_frame_counter);
+    }
```
tbirdso: Please guard debug functionality behind a disabled-by-default flag, or remove entirely

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+
+             echo ''
+             echo '=== Verifying Client Logs ==='
+             
+             # Check for successful frame sending
```
tbirdso: Thank you for adding regex checks. It looks like the test is still missing regex coverage for client frame reception, would you please add that as well?

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,236 @@
...
+             # Check for downstream connection
+             if grep -q 'Downstream connection established' \$SERVER_LOG; then
+               echo '✓ Server: Downstream connection established'
+               SERVER_SUCCESS=\$((SERVER_SUCCESS + 1))
+             else
+               echo '✗ Server: Downstream connection not established'
+             fi
+             
+             # Check for frame processing
+             if grep -q 'Processing UNIQUE frame' \$SERVER_LOG; then
```
tbirdso: This check looks good and verifies that the `StreamingServerUpstreamOp::compute` method has run.

Please also add a regex check to verify that frames are handled in `StreamingServerDownstreamOp`.

tbirdso (commented):
Thank you for adding regex checks in the integration test, added feedback.

applications/video_streaming_demo_enhanced/CMakeLists.txt
tbirdso: Yes, please reach out internally for help if you encounter issues with HoloHub infrastructure.

bhashemian:
cursor review

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
bhashemian, line 1119: The hot path generates very chatty INFO logs.

Every tensor triggers a multi-line INFO dump that includes information like shape, per-10-frame disk writes, and so on.
To keep logs actionable in production, please consider demoting most of these to DEBUG (or remove entirely) and leaving only one concise INFO summary.

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
```diff
@@ -0,0 +1,1601 @@
...
+
+    // FIXED: Replace continue with early return - validate tensor before processing
```
bhashemian: Here and in several other places, there are comments that do't seem to be relevant, like `FIXED:`, `🔧 POTENTIAL FIX:`, `CRITICAL FIX:`, etc. Could you please remove them?

operators/video_streaming/streaming_server_enhanced/streaming_server_upstream_op.cpp
```diff
@@ -0,0 +1,444 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
bhashemian: Here and in all the newly added files:
```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

operators/video_streaming/streaming_client_enhanced/metadata.json
```diff
@@ -0,0 +1,45 @@
+{
```
bhashemian, line 1: @cdinea for the operators, could you please remove the `_enhanced` from the folder names? 
Also do you think that it makes sense to simply have: 
- `operators/video_creaming/client`
- `operators/video_creaming/server`

**DEPENDS ON:**
@wyli does the `./holohub build` only uses the immediate folder name? For instance, if we have two operators defined under the same folder name but with different parent folder names, e.g.,  `operators/a/foo/my_op.cpp` and `operators/b/foo/your_op.cpp`, what does happen when calling `./holohub build foo`? can we or should we call `./holohub build a/foo`?

operators/video_streaming/streaming_client_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,282 @@
...
+  ${CMAKE_BINARY_DIR}/operators/streaming_client/libPoco.so
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+  ${STREAMINGCLIENT_DIR}/lib/${ARCH_DIR}/libssl.so.3
+  ${CMAKE_BINARY_DIR}/operators/streaming_client/libssl.so.3
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+  ${STREAMINGCLIENT_DIR}/lib/${ARCH_DIR}/libcrypto.so.3
+  ${CMAKE_BINARY_DIR}/operators/streaming_client/libcrypto.so.3
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+  ${STREAMINGCLIENT_DIR}/lib/${ARCH_DIR}/libcudart.so.12.0.107
+  ${CMAKE_BINARY_DIR}/operators/streaming_client/libcudart.so.12.0.107
```
bhashemian, line 155: Can we not depends on a specific CUDA minor version?

applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,109 @@
...
+cmake_minimum_required(VERSION 3.20)
+project(streaming_client_demo_enhanced CXX)
+
+# This application is built by the parent add_holohub_application macro
+
+find_package(holoscan 1.0 REQUIRED CONFIG
+             PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")
+
+# Find the streaming client operator directory
+set(OPERATOR_CLIENT_DIR "${CMAKE_SOURCE_DIR}/operators/video_streaming/streaming_client_enhanced/holoscan_client_cloud_streaming/lib/x86_64")
```
bhashemian: I see hard coded architecture here `/lib/x86_64`. Is it supposed to work on `x86_64` only?

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
```diff
@@ -0,0 +1,1601 @@
...
+          HOLOSCAN_LOG_DEBUG("Pre-send validation:");
+          HOLOSCAN_LOG_DEBUG("  - Frame valid: {}", frame.isValid() ? "YES" : "NO");
+          HOLOSCAN_LOG_DEBUG("  - Frame data: {}", frame.getData() ? "VALID" : "NULL");
+          HOLOSCAN_LOG_DEBUG("  - Frame size: {} bytes", frame.getDataSize());
+          HOLOSCAN_LOG_DEBUG("  - Client streaming: {}", client_->isStreaming() ? "YES" : "NO");
+          HOLOSCAN_LOG_DEBUG("  - Upstream ready: {}", client_->isUpstreamReady() ? "YES" : "NO");
+
+          send_success = client_->sendFrame(frame);
+          if (send_success) {
+            HOLOSCAN_LOG_INFO("✅ Frame sent successfully on attempt {}",
```
bhashemian: I am not so sure of using icons and emojis in the code. Having a series of tests that provides a summary of what has passed or failed would be much cleaner and easier to maintain? @tbirdso what do you think?

bhashemian (changes requested):
@cdinea thanks so much for this contribution. I have left few comments.

One challenge with this PR is the excessive logging information and unnecessary comments that reduce code readability. I would greatly appreciate it if you could make them concise and remove any unnecessary logging and comments.

operators/video_streaming/streaming_client_enhanced/CMakeLists.txt
cdinea: thank you for the feedback @bhashemian - these libraries came bundled together with the streaming and any modification would risk ABI compatibility issues and potential runtime failures.

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
cdinea: I appreciate the feedback. To clarify, these are Unicode characters (some from symbol blocks like ✅❌⚠️, others from emoji blocks like 🔍💾🎯) being used as visual markers in logging. I'd like to explain why I prefer keeping them for the current implementation:

- Visual scanning: When analyzing thousands of log lines, these characters provide instant visual cues (✅ successes, ❌ failures, ⚠️ warnings) that help quickly identify patterns
- Reduced debugging time: During live integration testing with the Holoscan Streaming Stack, spotting these markers has significantly improved issue identification
- Context categorization: Different symbols mark different operation types (🔍 investigation, 💾 disk I/O, 📊 data analysis)

applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/CMakeLists.txt
cdinea: Good catch! This is indeed an oversight. The libraries are actually provided for both x86_64 and aarch64 architectures in the NGC package. The server CMakeLists.txt already has proper architecture detection, but I missed implementing it in the client CMakeLists.txt. I'll update it to match the server's approach to support both architectures properly.

operators/video_streaming/streaming_server_enhanced/streaming_server_upstream_op.cpp
cdinea: fixed copyright header lines in [this](https://github.com/nvidia-holoscan/holohub/pull/1147/commits/17c4493821fe9ff2107b13212910d22dd6944f8b) commit

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
cdinea: the comments are relevant to the code block , I rephrased the comments to be more straightforward

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
cdinea: addressed the feedback in [this](https://github.com/nvidia-holoscan/holohub/pull/1147/commits/3730f367e6fbc071ee557b010766475a62d104d5) commit

operators/video_streaming/streaming_client_enhanced/metadata.json
cdinea: It makes sense to keep the "streaming" in the name as it is not just a simple client and server - there are other apps in holohub that have client and server arch and we want to differentiate that this is a streaming stack, even at the subdirectory level. The streaming designation serves as both a functional descriptor and an architectural marker. I will look into removing the "enhanced" add-on.

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: addressed the feedback in [this](https://github.com/nvidia-holoscan/holohub/pull/1147/commits/c6f937dfadd26b6a2d7a7f150e9eddffaf44b201) commit

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: addressed feedback in a recent commit

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
cdinea: All debug disk writes in the client operator are now guarded behind #ifdef HOLOSCAN_DEBUG_FRAME_WRITING (commit 36462a1a), matching the pattern already used in the server operators.

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: The "Frame Processing Stats" check is now explicitly required with proper error reporting (commit 6ebf7548). Since the log is deterministically produced every 30 frames and the test requires 6/6 server checks to pass, marking it as "optional" was inconsistent.

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: I've added minimum frame count thresholds to all streaming checks (commit 53478f3a). The test now requires ≥100 frames for all counts to verify sustained bidirectional streaming, not just a single frame. This threshold is conservative (~15% of typical 500-600 frame counts)

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: The test now requires 100% pass rate (lines 232-237). The success condition uses strict equality (-eq 6 and -eq 4) rather than minimum bounds, so any single check failure will cause the overall test to fail. The counts are correct: 6 server checks (client connected, upstream/downstream connections, upstream processing, downstream processing, frame stats) and 4 client checks (frames sent, frames received, validation passed, client started).

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: I modified the CMakeLists to use `add_holohub_application` instead of `add_subdirectory` and removed conditional build checks and manual operator enabling. This broke a few things and the client app run command has been modified.

applications/video_streaming_demo_enhanced/README.md
cdinea: Please note that we already included automated checks and this is referring to manual inspection of the integration log that is generated

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: Unfortunately, by applying this change, the integration test is not valid anymore as when I  changed to add_holohub_application(), it creates these options:
`APP_video_streaming_demo_client`
`APP_video_streaming_demo_server`
But the test infrastructure is setting:
`APP_video_streaming_demo_enhanced (line 546)`
This variable doesn't exist anymore, so nothing gets built - I will revert the changes to keep add_subdirectory, as I invested a lot of time to make the integration test working and this change affects the entire testing.

applications/video_streaming_demo_enhanced/README.md
cdinea: I tried running both of the apps without this docker opts and the client app terminates unexpectedly.I am logging this in JIRA to investigate but for now, we should keep it as par of the --docker-opts

applications/video_streaming_demo_enhanced/README.md
cdinea: The proposed modification

> We should use ./holohub test for CI/CD, similar to:

> ./holohub test integration_test

 would create a recursive execution pattern, as integration_test.sh already invokes:
`./holohub test video_streaming_demo_enhanced --base-img=nvcr.io/nvidia/clara-holoscan/holoscan:v3.5.0-dgpu --cmake-options="-DBUILD_TESTING=ON" --ctest-options="-R video_streaming_integration_test -V" --verbose 2>&1 | tee applications/video_streaming_demo_enhanced/integration_test.log`

My implementation follows the understanding that we should replace the` ./holohub run` command with appropriate modifications to enable both client and server applications to execute under the` ./holohub test `.
The wrapper script (integration_test.sh) provides essential development and debugging capabilities:

- Coordinated startup/shutdown of client-server components
- Log aggregation and analysis
- Timing synchronization between services
- Environment setup and validation

As an option, I can document that for CI/CD we can run directly the holohub test command 

```
./holohub test video_streaming_demo_enhanced \
  --base-img=nvcr.io/nvidia/clara-holoscan/holoscan:v3.5.0-dgpu \
  --cmake-options="-DBUILD_TESTING=ON" \
  --ctest-options="-R video_streaming_integration_test -V" \
  --verbose
```

applications/video_streaming_demo_enhanced/CMakeLists.txt
tbirdso: Please use `add_holohub_application(video_streaming_demo_enhanced ...)` to reflect the structure proposed in this PR.

Please reach out internally for help if you encounter issues with HoloHub infrastructure.

applications/video_streaming_demo_enhanced/CMakeLists.txt
tbirdso: One possible approach for graceful exit at test time could be to add a [`CountCondition`](https://docs.nvidia.com/holoscan/sdk-user-guide/api/cpp/classholoscan_1_1countcondition.html) so that the application stops after a set number of frames. See [Endoscopy Tool Tracking testing](https://github.com/nvidia-holoscan/holohub/blob/main/applications/endoscopy_tool_tracking/cpp/CMakeLists.txt#L70) for an example.

I'm not sure if that CountCondition approach would be directly applicable or if there might be a similar approach with a fixed expected frame count applicable to the video streaming test here. Do you see any obstacles?

Not a blocker, the timed approach and manual kill here is OK for integration coverage, frame count and frame testing could be a future enhancement.

applications/video_streaming_demo_enhanced/CMakeLists.txt
tbirdso: > Client Connection (lines 117-129): Verifies the client process stays running for 30 seconds (indicating successful connection)

This check is not sufficient to guarantee expected runtime operation

> Please refer to the detailed description in the [README](https://github.com/nvidia-holoscan/holohub/blob/e5460621b3c9965ef328e99017e91f7dfe989b86/applications/video_streaming_demo_enhanced/README.md#integration-testing)

Great, thanks for updating, please cover these expected log messages in regex pattern checks.

applications/video_streaming_demo_enhanced/CMakeLists.txt
tbirdso: > The success condition uses strict equality (-eq 6 and -eq 4) rather than minimum bounds, so any single check failure will cause the overall test to fail.

Please do not rely on fixed counts for checks, but instead mark the test as failed if any failure is present. Fixed counts can go stale if new checks are added.

operators/video_streaming/streaming_client_enhanced/streaming_client.cpp
tbirdso: > Having a series of tests that provides a summary of what has passed or failed would be much cleaner and easier to maintain?

I agree that a more granular set of tests could be nice, not necessarily a blocker. Possible design could be:
- Have one CTest entry with `add_test(video_streaming_integration ...)` that runs the client and server together and prints to a log, does basic checks to make sure app exit codes and basic output did not indicate error
- Have followup CTest entries such as `add_test(video_streaming_integration_server_upstream_check ...)` that depend on the integration test log output and parse for specific output, like making sure that the server upstream functionality ran as expected, client received frames, certain frame counts were met, certain performance was met, etc

In the interest of moving the contribution forward, I could be okay with keeping those checks as part of a single test entry for now, so long as it's clear that the log checks are performed and working as expected.

README.md
```diff
@@ -110,6 +110,8 @@ Run the following command to build the development container for a given project
 ./holohub build-container [project_name]
 ```
 
+For Holoscan SDK v3.6.1 and later, you can specify the CUDA version using the `--cuda` option (e.g., `--cuda 12` or `--cuda 13`), which automatically selects the appropriate container tag and passes the CUDA version as a build argument.
```
tbirdso, line 113: Please rebase on `main`, cannot properly review with changes from `main` appearing in the GitHub diff

```
git fetch origin
git rebase origin/main
```

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,274 @@
...
+  )
+endif()
+
+# Enable required operator
+set(OP_video_streaming ON CACHE BOOL "Build the video_streaming operator" FORCE)
+
+# Build client and server subdirectories directly
+# Note: We use add_subdirectory instead of add_holohub_application because:
+# 1. Both client and server are part of the same integration test
+# 2. The test infrastructure sets APP_video_streaming_demo_enhanced (parent name)
```
tbirdso: I don't see where this is set under the current implementation

applications/video_streaming_demo_enhanced/CMakeLists.txt
```diff
@@ -0,0 +1,274 @@
...
+# 3. Using add_holohub_application would create separate APP_*_client and APP_*_server options
```
tbirdso: `add_holohub_application(video_streaming_demo_enhanced ...)` will define `APP_video_streaming_demo_enhanced`.

applications/video_streaming_demo_enhanced/CMakeLists.txt
tbirdso: On review the updated server and client log checks in the inlined integration test shell script are indeed updated to cover log messages demonstrating `start` and `compute` functionality. Thanks!

applications/video_streaming_demo_enhanced/README.md
tbirdso: > My implementation follows the understanding that we should replace the ./holohub run command with appropriate modifications to enable both client and server applications to execute under the./holohub test.

Yes! This approach is good.

> The wrapper script (integration_test.sh) provides essential development and debugging capabilities:
> Coordinated startup/shutdown of client-server components
> Log aggregation and analysis
> Timing synchronization between services
> Environment setup and validation

I don't see any of this in the script. Can you please point to where this happens?

As it currently stands the `integration_test.sh` script seems redundant with directly calling `./holohub test`.

bhashemian:
@cdinea could you please address the linting issue? It’s improtant to fix it because it affects the readability and maintainability of the code. Splitting the code into some methods could solve the problem. Thanks

```console
operators/video_streaming/streaming_client_enhanced/streaming_client.cpp:1351:  Small and focused functions are preferred: StreamingClientOp::compute() has 536 non-comment lines (error triggered by exceeding 500 lines).  [readability/fn_size] [1]
```

README.md
cdinea: thank you for the for the feedback - I think I rebased a few times, it looks like @bhashemian rebased again yesterday.

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: rephrased comment and removed guard checking

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: updated comment to reflect unconditional build

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: Thank you for the feedback @tbirdso  - I have updated the parent `applications/CMakeLists.txt` to use `add_holohub_application(video_streaming_demo_enhanced ...)` instead of `add_subdirectory()`.

Within the application itself, the client and server components still use `add_subdirectory()` because they are tightly coupled components that must be built together for the integration test to function properly.

I noticed that several other applications in `applications/CMakeLists.txt` also use `add_subdirectory()` at the parent level:

```cmake
add_subdirectory(orsi)
add_subdirectory(nvidia_nim)
add_subdirectory(ehr_query_llm)
add_subdirectory(slang)
```

Could you clarify whether:
1. These applications should also be migrated to `add_holohub_application()`?
2. Or is `add_subdirectory()` acceptable when an application contains multiple sub-applications/components?

This will help me understand if the current structure (using `add_holohub_application` at parent level + `add_subdirectory` for client/server components) is the correct pattern, or if there's a preferred approach for applications with multiple components.

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: Thank you for the feedback @tbirdso ! I looked at the Endoscopy Tool Tracking example to see how the `CountCondition` is used : by modifying the YAML config at test time (changing `count: 0` to `count: 10`).

For the video streaming application, there are a few challenges with using `CountCondition`:

- bidirectional Streaming Verification Gets Complex - in our current approach we wait 30 seconds, then verify BOTH directions have 100+ frames. If we use CountCondition , the client stops after a while, let's say sending 300 frames
but the server frames might still be in transit -- the challenge here would be how long do we wait after client stops to verify server sent frames back?

- another issues is that CountCondition doesn't test sustained streaming.In the current test we verify that streaming works continuously for a duration (real-world scenario) but CountCondition tests "send N frames then stop" (less realistic). In production, real streaming applications run for hours, not "until N frames"

- another interesting scenario is , for example, in our test we stream for 30 seconds, expect 100+ frames in each direction at ~30fps. For CountCondition scenario it is unclear if we send 300 frames, but server might receive fewer due to buffering/latency. How many server→client frames should we expect in this scenario?

CountCondition seems suited for unit tests (process N items), but for integration testing, the timed approach better reflects real streaming usage patterns where applications run for durations, not frame counts.It is an interesting solution and I look into it for a subsequent PR if we consider it.

tbirdso: Makes sense, not a blocker 👍

applications/video_streaming_demo_enhanced/CMakeLists.txt
tbirdso: Thank you, these application groups all use `add_holohub_application(...)` under the nested subdirectory, please see https://github.com/nvidia-holoscan/holohub/blob/main/applications/orsi/CMakeLists.txt#L62 as an example.

applications/video_streaming_demo_enhanced/README.md
cdinea: thank you for the feedback @tbirdso . From my perspective, having the wrapper script is a good tool for debugging as it provides

1. **Environment setup**: Sets `HOLOHUB_BASE_SDK_VERSION=3.5.0` (line 40)
2. **Docker cache cleanup**: `docker system prune` for fresh builds (line 35)
3. **Shows current git commit**: For debugging which code version is tested (line 31)
4. **Log capture**: `| tee integration_test.log` (line 41)
5. **Adds required flags**: `--base-img=nvcr.io/.../holoscan:v3.5.0-dgpu` (line 41)
but  the client and server synchronization is done in the CMakeLists

In my opinion, ./holohub test is suited for CI/CD, this wrapper is more suited for developer convenience.

tbirdso: Thanks for the correction, that's fine.

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: Thank you for the feedback @tbirdso. I understand the concern about hardcoded counts becoming stale. 

However, the counting approach provides valuable diagnostic information that a simple failure flag doesn't:

1. **Diagnostic clarity**: Shows "Server checks: 5/6 passed" vs just "FAILED" 
2. **Partial success visibility**: Helps identify if it's a complete failure vs one flaky check
3. **CI/CD insights**: In logs, seeing "3/6" vs "6/6" immediately shows severity

we could keep counting for informational purposes, but use a **failure flag** for the actual pass/fail decision.

README.md
bhashemian: The branch is up to date with main but the rebase would cause a conflict. Although it is trivial to resolve, we can proceed as is since we will squash merge it.

applications/video_streaming_demo_enhanced/CMakeLists.txt
cdinea: I pushed a commit to enhanced the counting checks and add flag for failure/success for integration testing and a subsequent PR to update the README

bhashemian:
Thanks @cdinea for addressing the comments. @tbirdso, do you have any other comments here? If not, I’ll locally test and merge this PR.

bhashemian:
@cdinea, could you please include an image for this application in the README? It would enhance the visual appeal of the documentation card view for this application on the website. Thanks

bhashemian:
@cdinea running the server enhanced applications failed with the following error:
```
FileNotFoundError: [Errno 2] No such file or directory: '/workspace/holohub/build/video_streaming_demo_server/applications/video_streaming_demo_enhanced/video_streaming_demo_server/cpp/streaming_server_demo_enhanced_wrapper.sh'
```

cdinea:
thank you for the feedback @bhashemian - I have tested the app yesterday but I will test again and report back - it looks like one of the files might have got lost during rebase or cherry picking.

cdinea:
@bhashemian  I updated the READMEs- all the required files are present, but by changing the build structure and use add_holohub_applicationinstead of add_subdirectory for the parent app, the holohub run commands have changed and some READMEs have not been updated with the new holohub run commands.I have tested on my end , please let me know if you encounter any issues.I   also added a screenshot in the main app README

cdinea:
FYI Once this PR is merged - I will rebase [this](https://github.com/nvidia-holoscan/holohub/pull/1152) PR that has the Python bindings and one integration test and open it for review

applications/video_streaming_demo_enhanced/screenshot_streaming_demo.png
bhashemian, line 1: @cdinea thanks for adding a photo for this application. Is there any reason to include the VSCode screenshot? I think it would be better to only include the application's output image, and not the vscode.

cdinea: thank you @bhashemian for the prompt review - I added the vscode background to show the client and server running  in the console - do you prefer to only add the holoviz window with the surgical video?

applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/CMakeLists.txt
bhashemian: @cdinea any update on this?

applications/video_streaming_demo_enhanced/screenshot_streaming_demo.png
bhashemian: Yes, please. We want a title image which can just represent the application. Doesn't need to be detailed.

cdinea: I pushed the commit with the screenshot for the Holoviz window

bhashemian:
@cdinea I tested the application with HSDK CUDA12 image, and it works perfectly. Thanks

Howerver, running the client with CUDA 13 image fails with this error:
```
/workspace/holohub/build/video_streaming_demo_enhanced/applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/streaming_client_demo_enhanced: error while loading shared libraries: libcudart.so.12: cannot open shared object file: No such file or directory
```
Here is the base image that I used: `--base-img=nvcr.io/nvidia/clara-holoscan/holoscan:v3.7.0-cuda13`

operators/video_streaming/metadata.json
```diff
@@ -0,0 +1,51 @@
...
+                "affiliation": "NVIDIA"
+            }
+        ],
+        "version": "3.5.0",
+        "changelog": {
+            "3.5.0": "Unified video streaming operators with client and server components"
+        },
+        "language": ["C++"],
+        "platforms": ["x86_64", "aarch64"],
+        "tags": ["Video", "Streaming", "Client", "Server", "Networking", "Cloud", "Bidirectional"],
```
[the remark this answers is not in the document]
cdinea: @bhashemian is this feedback valid and needs to be addressed?

cdinea:
@bhashemian I am rebasing the branch now and will ping once done

operators/video_streaming/metadata.json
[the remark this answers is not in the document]
bhashemian: Yes but I'll take care of that.

applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/CMakeLists.txt
cdinea: updated in a recent commit to address both aarch for the client

operators/video_streaming/metadata.json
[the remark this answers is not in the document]
cdinea: is this feedback related to [this ](https://github.com/nvidia-holoscan/holohub/pull/1165)PR which has been merged ?

it looks like there might be a bug, since "Video" is part of the approved list 
"Video" ✅ (approved)

But it might be checking: "Networking" instead of "Networking and Distributed Computing" ❌
Or checking if "Cloud" is a valid category ?

probably this PR introduces a list that does not match the current tags and an option would be to remove non-category tags like "Client", "Server", "Cloud", "Bidirectional" - maybe it is good to document this somewhere so our developers can chose only from the approved list

applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/CMakeLists.txt
bhashemian: @cdinea could you please double check if this has been updated? Thanks

cdinea: I have done the aarch detection change, but it might have got lost in rebasing.I jsut pushed a new commit here https://github.com/nvidia-holoscan/holohub/pull/1147/commits/5a30ef224bbf9920f8144d7c2e6e444536255689

applications/video_streaming_demo_enhanced/video_streaming_demo_client/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+# The client requires a server connection to function properly
+# if(BUILD_TESTING)
+#   add_test(NAME streaming_client_demo_enhanced_cpp_test
+#            COMMAND streaming_client_demo_enhanced
+#                    --config ${CMAKE_CURRENT_BINARY_DIR}/streaming_client_demo_replayer.yaml
+#            WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+#   set_tests_properties(streaming_client_demo_enhanced_cpp_test PROPERTIES
+#                        PASS_REGULAR_EXPRESSION "StreamingClient created successfully"
+#                        FAIL_REGULAR_EXPRESSION "[^a-z]Error;ERROR;Failed")
+# endif()
```
bhashemian: @cdinea Can we remove this commented block?

cdinea: removed unnecessary code blocks from both client and server in this commit https://github.com/nvidia-holoscan/holohub/pull/1147/commits/2445e1fed4e4d0096443d8c569c7e098a0ae91be

bhashemian (approved):

PR #1149, opened 2025-10-02
CLI: add cuda version option

wyli:
adds a `--cuda` option, starting from Holoscan SDK v3.6.1, there are three container tags:
- cuda13: CUDA 13 (x86_64, Jetson Thor)
- cuda12-dgpu: CUDA 12 dGPU (x86_64, IGX Orin dGPU, Clara AGX dGPU, GH200)
- cuda12-igpu: CUDA 12 iGPU (Jetson Orin, IGX Orin iGPU, Clara AGX iGPU)
For SDK versions < 3.6.1, returns the old format (dgpu/igpu).

when --cuda is not specified and hsdk version is >= 3.6.1, the default is set via `get_default_cuda_version`


utilities/cli/util.py
```diff
@@ -314,27 +314,89 @@ def check_nvidia_ctk(min_version: str = "1.12.0", recommended_version: str = "1.
...
-    except subprocess.CalledProcessError:
-        return "dgpu"
-
+    # Check for iGPU (Orin integrated GPU)
+    if "Orin (nvgpu)" in gpu_name:
+        return "igpu"
     return "dgpu"
 
 
+def get_default_cuda_version() -> str:
```
wyli, line 352: please confirm this one, I copy it from @agirault script, guess the logic would change in the next few GA releases? @tbirdso

tbirdso: An alternative might be to check the CUDA driver version:
- <580: default to CUDA 12
- \>=580: default to CUDA 13

wyli: any recommendations? @agirault

utilities/cli/util.py
```diff
@@ -314,27 +314,92 @@ def check_nvidia_ctk(min_version: str = "1.12.0", recommended_version: str = "1.
...
+        cuda_version: CUDA major version (e.g., 12, 13). If None, uses platform default.
+        sdk_version: SDK version string (e.g., "3.6.0", "3.6.1", "3.7.0").
+
+    Returns:
+        The appropriate container tag string
+    """
+    try:
+        sdk_ver = parse_semantic_version(sdk_version)
+    except (ValueError, IndexError):
+        sdk_ver = (3, 7, 0)
```
bhashemian: Do we expect to maintain this extra versioning here?
What do think of importing `BASE_SDK_VERSION` (which should always be a valid semantic version) instead of this? This way we would have only one source or truth for the default version.

wyli: this is the smallest version number enabling the new container naming pattern, I think it's a good fallback for now. the `BASE_SDK_VERSION` now supports env var settings https://github.com/nvidia-holoscan/holohub/blob/aaddca50a5cb96849b70258683a16494d37bb57f/utilities/cli/container.py#L67C24-L67C65 may not follow the pattern

bhashemian: Sure, it’s not a blocker, but we should consider making it a single source of truth, using the default container version as the fallback. Thanks

wyli: sounds good, added a variable for the default

utilities/cli/util.py
agirault: ah yes, I actually implemented this over the weekend for isaac os, will share on slack

agirault:
```bash
get_default_cuda_version() {
  # Default to cuda 13 if nvidia-smi is not available.
  if ! command -v nvidia-smi >/dev/null 2>&1; then
    echo " > Warning: nvidia-smi not found, default cuda version is 13" >&2
    printf '%s' "13"
    return 0
  fi

  # Check the driver version using nvidia-smi.
  local driver_version
  if ! driver_version=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader 2>/dev/null); then
    echo " > Warning: Unable to detect NVIDIA driver version, default cuda version is 13" >&2
    printf '%s' "13"
    return 0
  fi

  local driver_major_version=$(echo "$driver_version" | cut -d. -f1)
  if [[ "$driver_major_version" -ge 580 ]]; then
    printf '%s' "13"
  else
    printf '%s' "12"
  fi
}
```

PR #1161, opened 2025-10-08
Fix HDMI input on the AJA KONA X card

nv-bogdan:
This PR enables the AJA source operator to utilize the HDMI input on the AJA KONA X card, which supports both SDI and HDMI connections.


nv-bogdan:
@jjomier , I'm OK to merge.

PR #1171, opened 2025-10-17
Updating cuda quantum application

jjomier:
## Summary by CodeRabbit

* **Chores**
  * Optimized CUDA Quantum application startup by simplifying runtime configuration
  * Consolidated application dependencies for streamlined installation and setup

### DIFF
--- a/applications/cuda_quantum/requirements.txt
+++ b/applications/cuda_quantum/requirements.txt
@@ -1,2 +1,1 @@
 cudaq
-holoscan


applications/cuda_quantum/requirements.txt
```diff
@@ -1,2 +1,2 @@
-cuda_quantum
+cudaq
 holoscan
```
bhashemian:
```suggestion
```

bhashemian: It is already present inside the docker container and also the package name has been changed to `holoscan-cu12` and `holoscan-cu13`, so it would be best to remove them and depends on the existing Holoscan SDK.

PR #1183, opened 2025-10-24
[yolo_model_deployment]: Patch export-det.py to use dynamo=False

mimiliaogo:
This PR solves two current issues in `yolo_model_deployment`: 
1. `ModuleNotFoundError: No module named 'onnxscript'`
2. Yolov8 model predicts incorrect results

Adds `dynamo=False` to the `torch.onnx.export` call to force the legacy
TorchScript exporter, as the default Dynamo backend produces incorrect
results or fails with YOLOv8 models.

Note: this PR is tested on X86.

## Summary by CodeRabbit

* **New Features**
  * YOLO model deployment export workflow now includes a pre-export patch step that adjusts ONNX export behavior (ensures dynamo=False), improving compatibility and predictability of exported models.


applications/yolo_model_deployment/patch_export.py
```diff
@@ -0,0 +1,24 @@
...
+
+import re
+
+# Read the file
+with open("YOLOv8-TensorRT/export-det.py", "r") as f:
+    content = f.read()
+
+# Adds `dynamo=False` to the `torch.onnx.export` call to force the legacy
+# TorchScript exporter, as the default Dynamo backend produces incorrect
+# results or fails with YOLOv8 models.
```
tbirdso, line 13: Is there an associated GitHub issue we could reference for future knowledge?
```suggestion
# Adds `dynamo=False` to the `torch.onnx.export` call to force the legacy
# TorchScript exporter, as the default Dynamo backend produces incorrect
# results or fails with YOLOv8 models.
# Refer to: https://github.com/triple-Mu/YOLOv8-TensorRT/issues/309
```

If not, can we open an upstream GitHub issue?

tbirdso (approved):
Thanks @mimiliaogo !

Could you please update the PR and/or commit description to note which target platforms you tested?

bhashemian:
@coderabbitai has been disabled. Enabling it to have a quick review.

applications/yolo_model_deployment/patch_export.py
mimiliaogo: It's not documented somewhere, but I've opened an issue here: https://github.com/triple-Mu/YOLOv8-TensorRT/issues/309.

tbirdso:
Thank you!

PR #1202, opened 2025-10-29
Fix build issues for applications

jjomier:
Added dockerfile location to metadata files

## Summary by CodeRabbit

* **Chores**
  * Added Dockerfile references to ProHawk Video Replayer (C++ & Python) and VPI Stereo metadata.
  * Added a Dockerfile for the gRPC H.264 endoscopy tracking app and removed its prior metadata reference.
  * Simplified base image selection and removed the CUDA version guard in the H.264 container build.
  * Removed application build steps from the Network Radar Pipeline container build.

### DIFF
--- a/applications/distributed/grpc/grpc_h264_endoscopy_tool_tracking/cpp/Dockerfile
+++ b/applications/distributed/grpc/grpc_h264_endoscopy_tool_tracking/cpp/Dockerfile
@@ -23,16 +23,6 @@ FROM nvcr.io/nvidia/clara-holoscan/holoscan:v${BASE_SDK_VERSION}-cuda13 as base
 
 ARG DEBIAN_FRONTEND=noninteractive
 
-# Starting from Holoscan SDK 3.6.0, only CUDA 13 is supported.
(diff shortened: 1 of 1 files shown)


tbirdso (approved):

applications/distributed/grpc/grpc_h264_endoscopy_tool_tracking/cpp/Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+
+ARG BASE_IMAGE
+ARG GPU_TYPE
+
+FROM ${BASE_IMAGE} as base
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Starting from Holoscan SDK 3.6.0, only CUDA 13 is supported.
+# To run H.264 application with CUDA 12, please use tags with holoscan-sdk-3.5.0 or earlier.
```
tbirdso: Could we update the existing H.264 Dockerfile that is already used across apps, rather than forking and adding a new condition for just this app?

jjomier: We need a new dockerfile because we have grpc dependencies.

PR #1210, opened 2025-11-04
Added coverage capability to Holohub CLI

jjomier:
## Summary by CodeRabbit

* **New Features**
  * Added --language flag to select cpp or python for tests.
  * Added --coverage flag to enable test coverage collection and XML output.

* **Documentation**
  * Added examples showing --coverage and --language usage.

* **Tests**
  * Added unit tests validating language and coverage flag forwarding and build/run behavior.
  * Added test-side support for generating coverage reports.

* **Chores**
  * Normalized Docker stage syntax, made Python executable used by tests configurable, and added .nv/ to .gitignore.

### DIFF
--- a/applications/multiai_endoscopy/python/multi_ai.py
+++ b/applications/multiai_endoscopy/python/multi_ai.py
@@ -29,8 +29,6 @@ from holoscan.operators import (
 )
 from holoscan.resources import UnboundedAllocator
 
-from holohub.aja_source import AJASourceOp
-
 
 class DetectionPostprocessorOp(Operator):
     """Example of an operator post processing the tensor from inference component.
@@ -156,7 +154,12 @@ class MultiAIDetectionSegmentation(Application):
 
         # start constructing app
         is_aja = self.source.lower() == "aja"
-        SourceClass = AJASourceOp if is_aja else VideoStreamReplayerOp
+        if is_aja:
+            from holohub.aja_source import AJASourceOp
+
+            SourceClass = AJASourceOp
+        else:
+            SourceClass = VideoStreamReplayerOp
         source_kwargs = self.kwargs(self.source)
         if self.source == "replayer":
             video_dir = os.path.join(self.sample_data_path, "endoscopy")


utilities/cli/holohub.py
```diff
@@ -914,10 +953,19 @@ def handle_test(self, args: argparse.Namespace) -> None:
         if args.verbose:
             ctest_cmd += "-VV "
 
+        # If coverage is requested, ensure gcov is available; install gcc if missing (as root)
+        if getattr(args, "coverage", False):
+            preinstall_cmd = (
+                "if ! command -v gcov >/dev/null 2>&1; then "
+                "(apt-get update && apt-get install -y gcc) || true; fi; "
```
wyli: I think this installation commands should be merged into the new `utilities/setup/coverage.sh`? (and if that's the case the 'as_root' change could be removed as well)

jjomier: Good catch - let me remove

applications/colonoscopy_segmentation/CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ if(BUILD_TESTING)
 
   # Add test
   add_test(NAME colonoscopy_segmentation_python_test
-           COMMAND python3 ${CMAKE_CURRENT_BINARY_DIR}/colonoscopy_segmentation_test.py
+           COMMAND ${PYTHON_EXECUTABLE_TEST} ${CMAKE_CURRENT_BINARY_DIR}/colonoscopy_segmentation_test.py
```
tbirdso, line 66: Will we need to do this replacement for all Python-based tests? Not a blocker, can be a followup

applications/volume_rendering/python/CMakeLists.txt
```diff
@@ -37,9 +37,8 @@ if(HOLOHUB_DOWNLOAD_DATASETS)
 endif()
```
tbirdso, line 37: I see an internal volume_rendering C++ coverage report in CDash, looks like coverage is being reported as missing for third party libraries under `_deps`.

How can a developer tell gcov to ignore certain directories or projects in the coverage report under the current `./holohub test` proposal here? Is it possible? Not a blocker

utilities/cli/holohub.py
tbirdso: I see the inline `apt-get install -y gcc` is removed in the latest version, can we add gcc installation in `utilities/setup/coverage.sh`?

jjomier: Actually gcc is needed for compiling so we should be good.

applications/colonoscopy_segmentation/CMakeLists.txt
jjomier: Yes - was planning a follow up MR

applications/volume_rendering/python/CMakeLists.txt
jjomier: Great point. Added the CTEST_CUSTOM_COVERAGE_EXCLUDE
much better : http://cdash.nvidia.com/viewCoverage.php?buildid=150778

tbirdso (approved):

PR #1213, opened 2025-11-04
Fix Build Issue with OpenIGTLink 3D Slicer Application

bhashemian:
OpenIGTLink 3D Slicer Application (`openigtlink_3dslicer_python`) fails to build in Python language. This PR fixes it by removing the unnecessary data dependency link to C++ target while building for Python.

## Summary by CodeRabbit

* **Chores**
  * Updated copyright year headers.
  * Removed an unnecessary build dependency declaration to streamline the build process.
  * Consolidated and cleaned up build-script formatting for clearer, single-block termination.

### DIFF
--- a/operators/tensor_to_file/CMakeLists.txt
+++ b/operators/tensor_to_file/CMakeLists.txt
@@ -34,4 +34,7 @@ target_include_directories(tensor_to_file PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
 )
 
-add_subdirectory(python)
+# Python module
+if(HOLOHUB_BUILD_PYTHON)
+  add_subdirectory(python)
+endif()


applications/openigtlink_3dslicer/python/CMakeLists.txt
```diff
@@ -30,6 +30,4 @@ if(HOLOHUB_DOWNLOAD_DATASETS)
     GXF_ENTITIES_FRAMERATE 30
     ALL
   )
-
-  add_dependencies(openigtlink_3dslicer colonoscopy_segmentation_data)
```
jjomier, line 34: Is this correct? We still need to download the data right?

jjomier: actually the data is always downloaded so it might work. Have you tested?

bhashemian: Yes, I have tested it locally, and it builds and downloads the data properly.

jjomier (approved):

PR #1216, opened 2025-11-06
Dev: Refactor optional components in default Dockerfile

tbirdso:
Refactors optional components in the top-level default development Dockerfile to remove less common components from the critical container build path.

Some components may be generally unused in the default container and could be removed. Comprehensive survey of component usage is out of scope for this change.

Users may opt into components installation by targeting specific Docker layers:
```
./holohub run-container --build-args="--target benchmarking-setup"
```

## Summary by CodeRabbit

* **Refactor**
  * Restructured the container build into a clearer multi-stage pipeline with dedicated stages for development, benchmarking, and tooling.

* **Chores**
  * Added a build-time configuration option and a development environment variable to simplify workflows.
  * Consolidated setup for benchmarking, Qcap, and DDS tooling into separate stages.
  * Expanded stage-specific notes and updated copyright year.

### DIFF
--- a/operators/deidentification/pixelator/test_pixelator.py
+++ b/operators/deidentification/pixelator/test_pixelator.py
@@ -19,9 +19,9 @@ from holoscan.core import Operator
 from .pixelator import PixelatorOp
 
 try:
-    from holoscan.code import BaseOperator
+    from holoscan.core import BaseOperator
 except ImportError:
-    from holoscan.code import _Operator as BaseOperator
+    from holoscan.core import _Operator as BaseOperator
 
 
 def test_pixelator_op_init(fragment):


Dockerfile
```diff
@@ -99,11 +104,37 @@ COPY benchmarks/holoscan_flow_benchmarking/requirements.txt /tmp/benchmarking_re
...
+FROM dev as yuan-qcap
+
+# Qcap dependency
+RUN apt update \
+    && apt install --no-install-recommends -y \
+        libgstreamer1.0-0 \
+        libgstreamer-plugins-base1.0-0 \
+        libgles2 \
+        libopengl0
+
```
bhashemian, line 126: Does this really need to be in the Holohub default image? Can't we make specific Dockerfile for those operators?

tbirdso: Probably not, and other components such as DDS may also be removable. I'd rather keep those decisions out of the scope of this PR, would it be OK if we follow up on those decisions in a later change?

Dockerfile
```diff
@@ -67,18 +72,18 @@ RUN /tmp/scripts/holohub setup && rm -rf /var/lib/apt/lists/*
...
-        libopengl0
+# --------------------------------------------------------------------------
+#
+# Set up common packages for advanced development with
+# Holoscan SDK Flow Benchmarking performance tools
+#
+# --------------------------------------------------------------------------
+FROM dev as benchmarking-setup
+
+ARG CMAKE_BUILD_TYPE
```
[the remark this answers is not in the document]
tbirdso: Not sure this is correct, but risk of updating is low.

Dockerfile
```diff
@@ -57,6 +55,13 @@ RUN if ! python3 -m pip --version >/dev/null 2>&1; then \
         curl -sS https://bootstrap.pypa.io/get-pip.py | ${PYTHON_VERSION} \
     ; fi
```
bhashemian, line 61: This layer only verifies the presence of Python and installs it if it’s missing. So naming it `holohub-cli-setup` could be confusing, wouldn’t it?

Dockerfile
```diff
@@ -67,18 +72,18 @@ RUN /tmp/scripts/holohub setup && rm -rf /var/lib/apt/lists/*
 # Enable autocomplete
 RUN echo ". /etc/bash_completion.d/holohub_autocomplete" >> /etc/bash.bashrc
```
bhashemian, line 78: This is also necessary for anyone who wants to just run the application (without developing anything), right? Does it make sense to merge it with `holohub-cli-setup`?

Dockerfile
tbirdso: The goal of this stage is to verify that prerequisites for running HoloHub CLI are met, it happens that the only prerequisite needed here is installing Python3 + `pip`

Would renaming to `holohub-cli-prerequisites` help clarify?

Dockerfile
tbirdso: I see where you're coming from, but the goal of the stages here should be:
1. Set up any prerequisites for the HoloHub CLI
2. Use HoloHub CLI to set up the development container, i.e. prepare a common build and runtime environment

We should keep packages for general user and dev experience in this stage. Is there a better name than `dev` you would suggest?

bhashemian: Maybe just `holohub-cli`? Not a blocker.

bhashemian (approved):
LGTM! Just it would be great if the name of the stages could be improved for clarity as discussed. Thanks

Dockerfile
tbirdso: Updated 👍

PR #1227, opened 2025-11-11
Enable fetch_holohub_operator to pull branch and apply patches

nv-mhaselton:
## Context

Add `fetch_holohub_operator` CMake functionality to checkout specific branches/tags and optionally apply patches to fetched operators.

## Description

Adds three new optional parameters to `fetch_holohub_operator`:

1. **`BRANCH`** - Checkout specific branches or tags (defaults to "main")
2. **`DEPTH`** - Control git clone depth (defaults to 1, use 0 for full history) 
3. **`PATCH_COMMAND`** - Apply patches after checkout

Example usage:
```cmake
include(${HOLOHUB_XR_ROOT}/cmake/FetchHolohubOperator.cmake)

# Fetch the XR operators.
fetch_holohub_operator(xr
  REPO_URL ${HOLOHUB_XR_ROOT}
  BRANCH ${HOLOHUB_BRANCH}
  DEPTH 0
  PATCH_COMMAND git apply "${CMAKE_SOURCE_DIR}/third_party/holohub/fix-xr-xrendframe.patch"
                && git apply "${CMAKE_SOURCE_DIR}/third_party/holohub/fix-xr-xrsession-openxr-version.patch"
)
```

## Summary by CodeRabbit

* **New Features**
  * Added an option to supply local patch files that are applied during repository retrieval (supports specifying branch; patches applied early in the checkout flow).
  * Validates patch file existence and fails fast if a patch is missing.

* **Documentation**
  * Updated docs and examples to show the new patch option and usage.

### DIFF
--- a/cmake/FetchHolohubOperator.cmake
+++ b/cmake/FetchHolohubOperator.cmake
@@ -27,17 +27,24 @@
 #   PATH - The path to the operator within the Holohub repository (defaults to OPERATOR_NAME)
 #   REPO_URL - The URL of the Holohub repository (defaults to git@github.com:nvidia-holoscan/holohub.git)
 #   BRANCH - The branch to checkout (defaults to "main")
+#   DEPTH - Git clone depth (defaults to 1 for shallow clone, use 0 for full history)
 #   DISABLE_PYTHON - Whether to build Python bindings
+#   PATCH_COMMAND - Custom command to run after checkout (e.g., to apply patches)
+#
+# Note: When using PATCH_COMMAND with patches created from older commits, you may need to increase
+#       DEPTH or set it to 0 to include sufficient git history for the patch to apply correctly.
 #
 # Example:
 #   fetch_holohub_operator(realsense_camera)
 #   fetch_holohub_operator(dds_operator_base PATH dds/base)
 #   fetch_holohub_operator(custom_operator REPO_URL "https://github.com/custom/holohub.git")
 #   fetch_holohub_operator(custom_operator BRANCH "dev")
+#   fetch_holohub_operator(custom_operator DEPTH 0)
 #   fetch_holohub_operator(custom_operator DISABLE_PYTHON)
+#   fetch_holohub_operator(custom_operator PATCH_COMMAND git apply ${CMAKE_CURRENT_SOURCE_DIR}/fix.patch)
 function(fetch_holohub_operator OPERATOR_NAME)
 
-  cmake_parse_arguments(ARGS "DISABLE_PYTHON" "PATH;REPO_URL;BRANCH" "" ${ARGN})
+  cmake_parse_arguments(ARGS "DISABLE_PYTHON" "PATH;REPO_URL;BRANCH;DEPTH" "PATCH_COMMAND" ${ARGN})
 
   if(NOT ARGS_REPO_URL)
     set(ARGS_REPO_URL "https://github.com/nvidia-holoscan/holohub.git")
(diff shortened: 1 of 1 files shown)


cmake/FetchHolohubOperator.cmake
```diff
@@ -70,7 +82,9 @@ function(fetch_holohub_operator OPERATOR_NAME)
             cmake/pybind11/
             cmake/pydoc/
             operators/operator_util.hpp
+        && git fetch --tags
```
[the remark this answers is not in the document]
nv-mhaselton: switched to git clone --branch. ...

cmake/FetchHolohubOperator.cmake
```diff
@@ -70,7 +82,8 @@ function(fetch_holohub_operator OPERATOR_NAME)
             cmake/pybind11/
             cmake/pydoc/
             operators/operator_util.hpp
-        && git checkout ${ARGS_BRANCH}
+        && git checkout
```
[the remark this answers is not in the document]
nv-mhaselton: updated to include branch name

cmake/FetchHolohubOperator.cmake
```diff
@@ -55,13 +57,23 @@ function(fetch_holohub_operator OPERATOR_NAME)
...
 
+  # Build patch application command if patches are provided
+  set(PATCH_COMMAND "")
+  if(ARGS_PATCHES)
+    foreach(PATCH_FILE ${ARGS_PATCHES})
+      # Convert to absolute path if needed
+      get_filename_component(PATCH_FILE_ABS "${PATCH_FILE}" ABSOLUTE)
+      list(APPEND PATCH_COMMAND "&&" "git" "apply" "${PATCH_FILE_ABS}")
+    endforeach()
+  endif()
```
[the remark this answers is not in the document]
nv-mhaselton: added suggested apply --check

cmake/FetchHolohubOperator.cmake
```diff
@@ -55,13 +57,28 @@ function(fetch_holohub_operator OPERATOR_NAME)
...
+
   # Fetch Holohub repository
   include(FetchContent)
   FetchContent_Declare(
       holohub_${OPERATOR_NAME}
       SOURCE_DIR "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src"
       DOWNLOAD_COMMAND
-        git clone --depth 1 --no-checkout ${ARGS_REPO_URL} "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src"
+        git clone --depth 1 --branch ${ARGS_BRANCH} --no-checkout ${ARGS_REPO_URL} "${FETCHCONTENT_BASE_DIR}/holohub_${OPERATOR_NAME}-prefix/src"
```
agirault: so the branch was just ignored until now? :D 

Should we make this a git ref to be able to provide better pinning?

nv-mhaselton: The below `&& git checkout ${ARGS_BRANCH}` was checking out main by default, but this did not seem to find a tag if it was not part of the initial shallow clone.

wyli (approved):
looks good to me, a minor concern apart from Alexis comment is that when the patch was created based on a previous commit, the git clone --depth 1 may not have the info for applying the patch, would be great to either document or add flexible depth.

cmake/FetchHolohubOperator.cmake
```diff
@@ -71,6 +88,7 @@ function(fetch_holohub_operator OPERATOR_NAME)
             cmake/pydoc/
             operators/operator_util.hpp
         && git checkout ${ARGS_BRANCH}
+        ${PATCH_COMMAND}
```
tbirdso: Why not use CMake's [PATCH_COMMAND](https://cmake.org/cmake/help/latest/module/ExternalProject.html#patch-step-options)? Looks like this is currently adding patching as part of the DOWNLOAD_COMMAND step. Could be nicer for CMake devs, not a blocker though

nv-mhaselton: Good point, I'll push an update to use the standard cmake parameter.

cmake/FetchHolohubOperator.cmake
nv-mhaselton: I believe it should be fine for users to make this configuration when using the CMake function?

nv-mhaselton:
@wyli, @tbirdso, @agirault - I've added optional parameters for DEPTH and aligned PATCHES to follow the PATCH_COMMAND standard. Please let me know if you'd like any additional changes 🙇

PR #1240, opened 2025-11-17
CLI mode update: accept cmd param and environment variables

wyli:
CLI mode enhancements:
- all modes allow for command line parameter overriding
  - previous behaviour: default mode (can override), explicit mode (no override)
  - this PR: default mode (can override), explicit mode (can override)
- this PR adds documentation and project schema about the existing 'mode.run.env' field, the field is used to define runtime environment variable
- add support of mode-level environment variable `mode.env` so that we specify the environment in `metadata.json`

## Summary by CodeRabbit

* **New Features**
  * Mode-level and runtime-only environment variables supported, with clear precedence rules.

* **Improvements**
  * CLI parameters now always override mode settings when provided; informative warnings shown instead of hard failures.
  * Validation improved with suggestions for unknown configuration keys and clearer messaging about overrides.

* **Tests**
  * Tests updated/added to cover mode override semantics and environment propagation.

### DIFF
--- a/utilities/cli/holohub.py
+++ b/utilities/cli/holohub.py
@@ -1075,12 +1075,9 @@ class HoloHubCLI:
         if args.local or os.environ.get("HOLOHUB_BUILD_LOCAL"):
             build_env = None
             if mode_config:
-                build_env = {}
-                if mode_config.get("env"):
-                    build_env.update(mode_config["env"])
                 build_cfg = mode_config.get("build", {})
                 if isinstance(build_cfg, dict) and build_cfg.get("env"):
-                    build_env.update(build_cfg["env"])
+                    build_env = dict(build_cfg["env"])
             self.build_project_locally(
                 project_name=args.project,
                 language=args.language if hasattr(args, "language") else None,
@@ -1202,12 +1199,9 @@ class HoloHubCLI:
             else:
                 build_env = None
                 if mode_config:
-                    build_env = {}
-                    if mode_config.get("env"):
-                        build_env.update(mode_config["env"])
                     build_cfg = mode_config.get("build", {})
                     if isinstance(build_cfg, dict) and build_cfg.get("env"):
-                        build_env.update(build_cfg["env"])
+                        build_env = dict(build_cfg["env"])
 
                 build_dir, project_data = self.build_project_locally(
                     project_name=args.project,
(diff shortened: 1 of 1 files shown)


wyli:
>  extending it to the upper level commands so that we can set env variables like `HOLOHUB_BUILD_LOCAL` or `HOLOHUB_BASE_SDK_VERSION`?

yes, this PR adds the support in metadata.json `mode.env`, build/run/install apps can have these in mode: `"env": {"HOLOHUB_BUILD_LOCAL": "1", "HOLOHUB_BASE_SDK_VERSION": "3.8.0"}`

bhashemian:
> > extending it to the upper level commands so that we can set env variables like `HOLOHUB_BUILD_LOCAL` or `HOLOHUB_BASE_SDK_VERSION`?
> 
> yes, this PR adds the support in metadata.json `mode.env`, build/run/install apps can have these in mode: `"env": {"HOLOHUB_BUILD_LOCAL": "1", "HOLOHUB_BASE_SDK_VERSION": "3.8.0"}`

Thanks @wyli, I tested it with video_streaming_server application. It works for `"env": {"HOLOHUB_BUILD_LOCAL": "1"}` but doesn't acknowledge `"env": {"HOLOHUB_BASE_SDK_VERSION": "3.5.0"}` and still use `3.8.0` for building the docker while setting `export HOLOHUB_BASE_SDK_VERSION=3.5.0` is working as expected!

wyli:
> > > extending it to the upper level commands so that we can set env variables like `HOLOHUB_BUILD_LOCAL` or `HOLOHUB_BASE_SDK_VERSION`?
> > 
> > 
> > yes, this PR adds the support in metadata.json `mode.env`, build/run/install apps can have these in mode: `"env": {"HOLOHUB_BUILD_LOCAL": "1", "HOLOHUB_BASE_SDK_VERSION": "3.8.0"}`
> 
> Thanks @wyli, I tested it with video_streaming_server application. It works for `"env": {"HOLOHUB_BUILD_LOCAL": "1"}` but doesn't acknowledge `"env": {"HOLOHUB_BASE_SDK_VERSION": "3.5.0"}` and still use `3.8.0` for building the docker while setting `export HOLOHUB_BASE_SDK_VERSION=3.5.0` is working as expected!

confirm "HOLOHUB_BASE_SDK_VERSION": "3.5.0" doesn't work, I'll have a look and have follow-up PR if it's a minor feature request..

if the variables are for docker build I've tested this works: ```./holohub build-container --build-args="--build-arg BASE_SDK_VERSION=3.6.0 --build-arg BASE_IMAGE=nvcr.io/nvidia/clara-holoscan/holoscan:v3.6.0-dgpu"```

utilities/cli/README.md
```diff
@@ -265,6 +266,10 @@ Each mode is defined as a named object under the `modes` key:
       "production": {
         "description": "High-performance mode with GPU acceleration",
         "requirements": ["gpu", "model", "tensorrt"],
+        "env": {
+          "CUDA_VISIBLE_DEVICES": "0",
```
tbirdso: Wouldn't we generally >=1 GPUs for runtime?

Is there a good way to either set env only for build time, or to "unset" a certain variable at runtime?

wyli: thanks, updated the example, also added 'mode.build.env' following the same pattern, now support build time only vars.

PR #1241, opened 2025-11-17
CLI: add a version checking mechanism

wyli:
appending a version check between the wrapper's `CLI_PINNED_COMMIT` and the currently downloaded version hash `.cli_commit_hash` when running the wrapper script
```
════════════════════════════════════════════════════
⚠️  ./holohubCLI_wrapper CLI version does not match CLI_PINNED_COMMIT.
Current:  304ac263 (utilities/cli/.cli_commit_hash)
Expected: 04ac2639
To reinstall pinned CLI, run:
  CLI_FORCE_UPDATE=1 ./holohubCLI_wrapper --help
════════════════════════════════════════════════════
```

CLI wrapper workflow after this PR:
- first time run (or after removing the existing CLI module)
  - wrapper runs git clone and writes `.cli_commit_hash` file to track the upstream commit
- Normal wrapper script usage after downloading
  - verion_check.py is launched to after every CLI command for local consistency
  - prints a reminder to update CLI when wrapper's required version is different from the local version
 - user can refresh the cli with `CLI_FORCE_UPDATE=1 ./holohubCLI_wrapper --help`

## Summary by CodeRabbit

* **New Features**
  * CLI installer now supports a pinned-commit workflow, records the deployed commit, and can run a post-run version check while preserving exit behavior.
* **Bug Fixes / UX**
  * Adds a clear bordered warning when local and pinned commits mismatch; update check can be skipped via an environment flag.
* **Tests**
  * Added test covering the update-check and mismatch warning output.
* **Chores**
  * Added an ignore pattern to avoid committing CLI update metadata.

### DIFF
--- a/tutorials/holohub_operators_external_applications/holohubCLI_wrapper
+++ b/tutorials/holohub_operators_external_applications/holohubCLI_wrapper
@@ -52,11 +52,6 @@ TMP_HOLOHUB_DIR="${SCRIPT_PATH}/.tmp-holohub"
 REPO_URL="${CLI_REPO_URL:-https://github.com/nvidia-holoscan/holohub.git}"
 CLI_PINNED_COMMIT="${CLI_PINNED_COMMIT:-304ac2639ab004368c8ed9b3f2a47de2521969ca}"
 
-if [ -z "${CLI_PINNED_COMMIT}" ]; then
-    echo "Error: CLI_PINNED_COMMIT must be set to a valid commit hash."
-    exit 1
-fi
-
 cleanup_tmp() {
     rm -rf "${TMP_HOLOHUB_DIR}"
 }


tutorials/holohub_operators_external_applications/holohubCLI_wrapper
```diff
@@ -16,23 +16,44 @@
 # limitations under the License.
 
 SCRIPT_PATH="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_URL="${CLI_REPO_URL:-https://github.com/nvidia-holoscan/holohub.git}"
```
tbirdso, line 52: Do we expect to ever host the CLI at another repo?

tutorials/holohub_operators_external_applications/holohubCLI_wrapper
```diff
@@ -16,23 +16,44 @@
...
-# The script checks if utilities/cli exists in the current directory,
-# if not it pulls the holohub CLI from the holohub repo
-# To update the holohub CLI, remove the utilities/cli directory and run the script again
-if [ ! -d "${SCRIPT_PATH}/utilities/cli" ]; then
-    git clone --depth 1 --no-checkout https://github.com/nvidia-holoscan/holohub.git ${SCRIPT_PATH}/.tmp-holohub/holohub \
-        && cd ${SCRIPT_PATH}/.tmp-holohub/holohub \
+# Function to download CLI from holohub repo (this is a demo omitting error handling for brevity)
+download_cli() {
+    echo "Downloading holohub CLI..."
+    PINNED_COMMIT="${CLI_PINNED_COMMIT:-}"
```
tbirdso: Why define REPO_URL at the top level but PINNED_COMMIT nested locally?

tutorials/holohub_operators_external_applications/holohubCLI_wrapper
```diff
@@ -16,23 +16,44 @@
...
+if [ "${CLI_FORCE_UPDATE}" = "1" ] || [ ! -d "${SCRIPT_PATH}/utilities/cli" ]; then
+    download_cli
 fi
 
 export HOLOHUB_CMD_NAME=${BASH_SOURCE[0]}
 
+# Run the CLI command
 PYTHONPATH=${PYTHONPATH}:${SCRIPT_PATH} python3 ${SCRIPT_PATH}/utilities/cli/holohub.py "$@"
+CLI_EXIT_CODE=$?
+# Check for CLI updates after command completes
```
tbirdso: Any advantage to checking after vs before?

utilities/cli/version_check.py
```diff
@@ -0,0 +1,144 @@
...
+import subprocess
+import time
+from pathlib import Path
+
+
+def check_for_cli_updates() -> None:
+    """
+    Check if a newer version of the CLI is available.
+
+    - Checks once per 24 hours (configurable via CLI_CHECK_INTERVAL)
```
agirault: That seems a bit overengineered? I was hoping of something simpler:
- if using a git tag or specific git ref, ensure the sha requested matches the sha cached
- if using a branch, only check if there is a more recent sha than the sha cached if explicitly asking to update

utilities/cli/version_check.py
```diff
@@ -0,0 +1,144 @@
...
+    - Fails silently on network issues or missing files
+
+    Environment Variables:
+        CLI_SKIP_UPDATE_CHECK: Set to "1" to disable update checks
+        CLI_CHECK_INTERVAL: Seconds between checks (default: 86400 = 24 hours)
+        CLI_REPO_URL: Custom repository URL (default: https://github.com/nvidia-holoscan/holohub.git)
+        HOLOHUB_BRANCH: Branch to track (default: main)
+        CLI_PINNED_COMMIT: If set, pin the CLI to this commit hash (short or full).
+            - When set, remote update checks are skipped.
+            - If the local CLI commit differs from the pinned commit, a warning is shown.
```
agirault: That seems a bit overengineered? I was hoping of something simpler:

- if using a git tag or specific git ref, ensure the sha requested matches the sha cached
- if using a branch, only check if there is a more recent sha than the sha cached if explicitly asking to update


```suggestion
    Environment Variables:
        HOLOHUB_CLI_REPO_URL: Custom repository URL (default: https://github.com/nvidia-holoscan/holohub.git).
        HOLOHUB_CLI_FORCE_UPDATE: Set to "1" to update the version of the CLI when HOLOHUB_CLI_GIT_REF is not a specific commit hash (ex: branch, tag that got updated).
        HOLOHUB_CLI_GIT_REF: version of the CLI to fetch as a git reference (branch, tag, commit hash)
            - If using a branch or tag, use HOLOHUB_CLI_FORCE_UPDATE to refetch potential updates.
            - If using a specific commit hash, the CLI will automatically get updated if the cached CLI commit hash does not match the requested hash.
```

agirault: @coderabbitai can you chime in?

agirault: That's my thinking also. Eventually we'll probably want to use a proper package manager for this like pip so the less code we write to maintain and drop the better IMO.

And `HOLOHUB_BRANCH` coud be kept for compatibility purposes and marked as deprecated with warning to switch to `HOLOHUB_CLI_GIT_REF`.

tutorials/holohub_operators_external_applications/holohubCLI_wrapper
agirault: right, seems like REPO_URL isn't needed outside of download_cli. If the concern was to keep the default value of top of the script, could have a separate variable called `HOLOHUB_CLI_DEFAULT_REPO_URL` then used here as default.

tutorials/holohub_operators_external_applications/holohubCLI_wrapper
wyli: could be set to a fork of this repo for example for debugging

tutorials/holohub_operators_external_applications/holohubCLI_wrapper
wyli: yes, printing the upgrade reminder at the end of any build/run logging is hopefully more noticeable

utilities/cli/version_check.py
wyli: I think regular checks and reminders for the update is a good feature, otherwise the users will have to remember when and how to reinstall CLI which is not ideal

agirault: I think it's not necessary in general, people should be using a specific sha/version instead of a non unique git ref, and that's then a non-issue. Package managers generally don't even do that, and if we point out the ones that do, I think we'd rather depend on an actual package manager to handle versions of our CLI, and not write and maintain our own.

And even if we agreed to say "we need that", that's not what we're after today as a quick and simple way to just strictly pin the cli version.

PR #1246, opened 2025-11-19
CLI add info for sccache

wyli:
` ./holohub env-info` prints the relevant config:
```
...
sccache Information:
  HOLOHUB_ENABLE_SCCACHE: false (disabled)
  sccache binary: (not found in PATH)
  sccache version: (unavailable)
  Effective SCCACHE_DIR: /home/user/.cache/sccache
  SCCACHE_* environment variables: (none set)
...
```

## Summary by CodeRabbit

* **New Features**
  * Environment diagnostics now include SCCACHE details: enabled status, binary path, version, and effective cache directory.

* **Chores**
  * SCCACHE cache directory is resolved dynamically from the build environment when enabled.
  * Exposed environment variables updated to include SCCACHE enablement for clearer diagnostics.


utilities/cli/util.py
```diff
@@ -1353,9 +1354,41 @@ def collect_env_info() -> None:
...
+    effective_dir = get_sccache_dir()
+    print(f"  Effective SCCACHE_DIR: {effective_dir}")
+
+    sccache_items = sorted(
+        (k, os.environ.get(k) or "(not set)") for k in os.environ if k.startswith("SCCACHE_")
+    )
+    if sccache_items:
+        print("  SCCACHE_* environment variables:")
+        for key, value in sccache_items:
+            print(f"    {key}: {value}")
```
agirault: nitpick: looping twice here, could do in a single loop


Also could skip SCCACHE_DIR since already printed?

utilities/cli/util.py
```diff
@@ -1353,9 +1354,41 @@ def collect_env_info() -> None:
...
+    enable_val, enabled = get_env_bool("HOLOHUB_ENABLE_SCCACHE", default=False)
+    print(f"  HOLOHUB_ENABLE_SCCACHE: {enable_val} ({'enabled' if enabled else 'disabled'})")
+
+    sccache_bin = shutil.which("sccache")
+    version = run_info_command(["sccache", "--version"]) if sccache_bin else None
+    print(f"  sccache binary: {sccache_bin or '(not found in PATH)'}")
+    print(f"  sccache version: {version or '(unavailable)'}")
+
+    effective_dir = get_sccache_dir()
+    print(f"  Effective SCCACHE_DIR: {effective_dir}")
```
agirault: would `Local` be more meaningful than `Effective`? Or can this be run inside the container as well?

utilities/cli/util.py
```diff
@@ -1353,9 +1354,41 @@ def collect_env_info() -> None:
...
+    """Collect and display sccache-related information"""
+    print(f"\n{Color.blue('sccache Information:')}")
+
+    enable_val, enabled = get_env_bool("HOLOHUB_ENABLE_SCCACHE", default=False)
+    print(f"  HOLOHUB_ENABLE_SCCACHE: {enable_val} ({'enabled' if enabled else 'disabled'})")
+
+    sccache_bin = shutil.which("sccache")
+    version = run_info_command(["sccache", "--version"]) if sccache_bin else None
+    print(f"  sccache binary: {sccache_bin or '(not found in PATH)'}")
+    print(f"  sccache version: {version or '(unavailable)'}")
```
agirault, line 1376: This is only valuable inside the container, not sure that it makes sense in env-info?

utilities/cli/util.py
agirault, line 1390: I do print this everytime we start building if sccache is enabled, since I feel like it's just as valid as the CMAKE_ variables that are printed after by cmake to understand the env. Do you think we should remove it from the build if we can see them here instead?

utilities/cli/util.py
```diff
@@ -1353,9 +1354,41 @@ def collect_env_info() -> None:
...
+
+    effective_dir = get_sccache_dir()
+    print(f"  Effective SCCACHE_DIR: {effective_dir}")
+
+    sccache_items = sorted(
+        (k, os.environ.get(k) or "(not set)") for k in os.environ if k.startswith("SCCACHE_")
+    )
+    if sccache_items:
+        print("  SCCACHE_* environment variables:")
```
agirault, line 1388: This won't take into account default values we set inside the container, but I guess that's fine?

utilities/cli/util.py
tbirdso: Is `sccache` support only available in containers, or could devs also use sccache for local development?

utilities/cli/util.py
wyli: Yes I think the build print is also good as part of log for any debugging

utilities/cli/util.py
agirault: they can use it locally as well, they'd need to call setup to install it. My point was more that for containers, you would not have (or care about) sccache locally.

Looks like the other `collect_*_info()` methods are only focused on system specific info, this this isn't

wyli: if we run `./holohub setup` locally, the sccache is available for local dev, no?

utilities/cli/util.py
wyli: yes, I think currently we need `./holohub env-info` and `./holohub run-container -- ./holohub env-info` to get both inside and outside info...

tbirdso (approved):

PR #1256, opened 2025-11-25
Enhance Holohub CLI Environment Handling 

bhashemian:
## Summary by CodeRabbit

* **Refactor**
  * Unified environment and path-mapping across build, run, and install flows; consolidated workdir resolution and local vs container branching.

* **New Features**
  * Verbose path-mapping output and two-stage placeholder resolution for paths and env vars.
  * New env-update behavior supporting prepend/append semantics and templated replacements.
  * Exposed configurable CLI prefix for path resolution.

* **Bug Fixes**
  * Improved env precedence, language-aware defaults, and clearer build/run/install messaging.

* **Tests**
  * Expanded tests for placeholder replacement, mode env precedence, and isolated environment handling.

* **Documentation**
  * Expanded CLI docs with modes, examples, env precedence, and migration guidance.

### DIFF
--- a/utilities/cli/README.md
+++ b/utilities/cli/README.md
@@ -387,23 +387,14 @@ Both modes automatically share the same Docker image name (`holohub:myapp`), so 
     ```json
     "modes": {
       "mode_name": {
+        "run": {
+          "env": {
+            "PATH": "<PATH>:<holohub_app_bin>/bin"  # append to the updated PATH by mode_name.env
+            "CMAKE_BUILD_TYPE": "Release"  # override the existing CMAKE_BUILD_TYPE
+          }
+        },
         "env": {
-          "PATH": "<PATH>:<holohub_app_bin>/bin"  # append to the updated PATH by modes.env
-        }
-      }
-      "env": {
-        "PATH": "<holohub_bin>/bin:<PATH>"  # prepend to the existing PATH in the environment
-      }
-    }
-    ```
-
-  - If it is without appending or prepending, it will override the existing environment variable, like:
-
-    ```json
-    "modes": {
-      "mode_name": {
-        "env": {
-          "PATH": "<holohub_app_source>/bin"  # override any existing PATH
+          "PATH": "<holohub_bin>/bin:<PATH>"  # prepend to the existing PATH in the environment
         }
       }
     }


utilities/cli/README.md
```diff
@@ -355,13 +374,42 @@ For cases where build and run containers need different Docker configurations, y
...
+
+    ```json
+    "modes": {
+      "mode_name": {
+        "env": {
+          "PATH": ":<holohub_app_bin>/bin"  # append to the existing PATH
+        }
+      }
+      "env": {
+        "PATH": "<holohub_bin>/bin:"  # prepend to the existing PATH
```
tbirdso: Nitpick: Could please we have an explicit substitution for clarity, such as follows:
```suggestion
          "PATH": "<PATH>:<holohub_app_bin>/bin"  # append to the existing PATH
        }
      }
      "env": {
        "PATH": "<holohub_bin>/bin:<PATH>"  # prepend to the existing PATH
```

Having explicit `<PATH>` would align better to the existing substitution approach with `<holohub_bin>`, etc and would be clearer for developers

bhashemian:
@coderabbitai full review

utilities/cli/README.md
bhashemian: I like your suggestion but for that we should resolve all the env vars since the user could potentially use other variable names to append to another one, like:
```
"PATH": "<LD_LIBRARY_PATH>:<holohub_app_bin>/bin"
```
This could open a door to various cases to consider and we should define a way to distinguish between placeholders (`<holohub_app_bin>`) and the env vars (`<PATH>`, `<my_other_env_var>`, `<holohub_my_env_var>`, etc.).

I thought it would be simpler to not explicitly mention the variable name as we only allow to append/prepend to the same variable, what do you think?

bhashemian: Having said that, I still like being explicit too. What about considering anything starting with `holohub_` (the `prefix`) as the placeholder, which means not allowing env var starting with `holohub_`, and the rest as the env var?

bhashemian: I made them explicit. Thanks for your suggestion.

utilities/cli/README.md
```diff
@@ -355,13 +374,42 @@ For cases where build and run containers need different Docker configurations, y
...
+    "modes": {
+      "mode_name": {
+        "env": {
+          "PATH": "<holohub_app_source>/bin"  # override the existing PATH
+        }
+      }
+    }
+    ```
+
+  - The inner `env` object takes the precedence over the top-level `env` object, and the top-level `env` object takes precedence over the CLI environment variables.
```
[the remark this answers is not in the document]
bhashemian: @coderabbitai the README has been updated. Please check again.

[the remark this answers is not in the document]
bhashemian: @coderabbitai I have updated it, please check again?

bhashemian:
> Took a quick look, changes LGTM
> 
> @bhashemian did you encounter an issue with any specific apps that prompted this work? Could you please add a quick note on motivation in the description to help with understanding? Thanks!

Thanks @tbirdso! Sure, the initial motivation was to enable following in the metadata.json:

https://github.com/AndreasHeumann/holohub/blob/6005d3b6e959136d6e61e4c3f78fd6eb3e61c401/applications/pipeline_visualization/python/metadata.json#L35-L50

PR #1262, opened 2025-11-26
Download Datatsets Hosted on Google Drive

bhashemian:
## Summary by CodeRabbit

* **New Features**
  * Optional gated automatic dataset download for EndoNeRF (images, depth, masks, poses), with poses_bounds placed in the dataset root and completion stamping.
  * Downloader adds a --dataset_name option and improves Google Drive downloads (gdown fallback and safer handling).

* **Documentation**
  * Dataset step now requires agreeing to dataset terms, shows how to disable auto-download, and adds run-training and dynamic-rendering guidance.

* **Chores**
  * Container build now installs Google Drive download tooling.

### DIFF
--- a/utilities/download_ngc_data
+++ b/utilities/download_ngc_data
@@ -29,11 +29,6 @@ while [[ $# -gt 0 ]]; do
       shift # past argument
       shift # past value
       ;;
-    --dataset_name)
-      DATASET_NAME="$2"
-      shift # past argument
-      shift # past value
-      ;;
     --download_dir)
       DOWNLOAD_DIR="$2"
       shift # past argument
@@ -41,6 +36,11 @@ while [[ $# -gt 0 ]]; do
       ;;
     --download_name)
       DOWNLOAD_NAME="$2"
+      shift # past argument
+      shift # past value
+      ;;
+    --dataset_name)
+      DATASET_NAME="$2"
       shift # past argument
       shift # past value
       ;;
@@ -97,7 +97,7 @@ set -- "${POSITIONAL_ARGS[@]}" # restore positional parameters
 display_usage() {
 	echo "Usage"
 	echo ""
-	echo "  download_ngc_data [options] --url <url> --dataset_name <dataset_name> --download_dir <download_dir> --download_name <download_name>"
+	echo "  download_ngc_data [options] --url <url> --download_dir <download_dir> --download_name <download_name> [--dataset_name <dataset_name>] "
 	echo ""
 	echo "This script downloads resources from NGC and optionally runs a script to convert video files to GXF entities."
   echo ""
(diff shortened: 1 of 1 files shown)


bhashemian:
`gdown` can't download folders with  more than 50 files!

utilities/download_ngc_data
```diff
@@ -305,8 +305,12 @@ fi
 # Make the download directory
 mkdir -p ${download_dir}
 
-if [ "${url:0:8}" == "https://" ]; then
-
+if [[ ${url} == *"drive.google.com"* ]]; then
+    download_command=gdown
+    if ! command -v gdown > /dev/null; then
+      fatal R "Please install " W "gdown" R " command to download datasets. Follow the instructions in https://github.com/wkentaro/gdown to install gdown."
+    fi
```
tbirdso, line 325: Nitpick: could we check for the `down` existence at the site where it is called rather than when we set the command to use here?

bhashemian: Thanks @tbirdso for your feedback but checking it here is consistent with other checks that we have for `wget` and `ngc`. Is there any reason you prefer to check later when it is called?

tbirdso: > checking it here is consistent with other checks

Sounds good 👍 

> Is there any reason you prefer to check later when it is called?

Just future proofing in case the initial selection was ever decoupled from the execution site such that the user had a chance to install `gdown` in between. Not a likely scenario with regards to current implementation, not a blocker 👍

bhashemian:
@coderabbitai update summary

utilities/download_ngc_data
```diff
@@ -29,6 +29,11 @@ while [[ $# -gt 0 ]]; do
       shift # past argument
       shift # past value
       ;;
+    --dataset_name)
```
tbirdso: Could we add the same in Holoscan SDK upstream for our next upcoming release? https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/scripts/download_ngc_data

utilities/download_ngc_data
```diff
@@ -92,14 +97,15 @@ set -- "${POSITIONAL_ARGS[@]}" # restore positional parameters
 display_usage() {
 	echo "Usage"
 	echo ""
-	echo "  download_ngc_data [options] --url <url> --download_dir <download_dir> --download_name <download_name>"
+	echo "  download_ngc_data [options] --url <url> --dataset_name <dataset_name> --download_dir <download_dir> --download_name <download_name>"
```
tbirdso: If I'm reading correctly, it seems like `download_name` and `dataset_name` may match by default. To keep backwards compat, could we make `dataset_name` an optional argument with a default value of `download_name`?

```suggestion
	echo "  download_ngc_data [options] --url <url> --download_dir <download_dir> --download_name <download_name> [--dataset_name <dataset_name>] "
```

+ update at call sites

tbirdso (approved):
Looks good! Please see question about preserving `download_ngc_data` interface backwards compatibility

utilities/download_ngc_data
bhashemian: That's a good suggestion, thanks! `--dataset_name` is now optional and defaults to value of `--download_name`

utilities/download_ngc_data
bhashemian: Sure!

PR #1269, opened 2025-12-01
Add ZLIB dependency to volume renderer CMake configuration

jjomier:
- Included ZLIB as a required package in the CMakeLists.txt for the volume renderer.
- Updated target_link_libraries to link against ZLIB::ZLIB, ensuring proper integration for compression functionalities.

## Summary by CodeRabbit

* **Chores**
  * Build system configuration updated with new dependencies.


tbirdso:
@jjomier  is this a new feature, or addressing a gap in the existing feature support?

Could you please update the description to clarify motivation?

jjomier:
> * d ZLIB as a required package in the CMakeLists.txt for the volume re

This was a bug. I believe zlib is only brought up with the volume_loader, but using only volume_renderer would fail to find zlib.h

PR #1277, opened 2025-12-04
CLI: enable ./holohub test for different project types

wyli:
- extend the current `./holohub test app_name` to test different types of projects (operators, tutorials...)
- fixes an issue when forming docker image with certain app names
  - `ERROR: failed to build: invalid tag "holohub-myapp:test:b377f6a5bfde": invalid reference format`

## Summary by CodeRabbit

* **New Features**
  * Added support for new project types (GXF extensions and tutorials).
  * Tests now run in per-project contexts with repository-wide fallback.

* **Bug Fixes**
  * Project names are now sanitized for Docker/container operations.

* **Refactor**
  * Build and test workflows derive prefixes and project names dynamically from project metadata.

* **Documentation**
  * Added usage notes for running tests via the tooling.


utilities/cli/container.py
```diff
@@ -331,13 +332,16 @@ def get_conditional_options(
     @property
     def image_name(self) -> str:
         if self.dockerfile_path != HoloHubContainer.default_dockerfile():
-            return f"{self.CONTAINER_PREFIX}:{self.project_metadata.get('project_name', '')}"
+            project_tag = self.get_project_name()
+            if project_tag:
+                return f"{self.CONTAINER_PREFIX}:{project_tag}"
+            return self.CONTAINER_PREFIX
```
tbirdso, line 338: What is an example of a tag that might need this logic? Wondering if we need to append a default `:latest`, etc

utilities/cli/container.py
```diff
@@ -417,6 +421,17 @@ def dockerfile_path(self) -> Path:
         # Strategy 5-6: Fall back to default Dockerfile
         return HoloHubContainer.default_dockerfile()
 
+    def get_project_name(self) -> str:
+        """Return docker-safe project name."""
```
tbirdso, line 425: Should we enforce that project names themselves must be docker-safe?

Where does the `test:...` name mentioned in the PR description come from?

utilities/cli/container.py
wyli: this is the legacy name that we've been using when the app has it's own dockerfile, `./holohub run qt_video_replayer --dryrun` -> `-t holohub:qt_video_replayer`

utilities/cli/container.py
wyli: that actual case I faced was `XrFrameOp` -> `ERROR: failed to build: invalid tag "holohub-XrFrameOp:5e46dc2c8aa7": repository name must be lowercase`. 

> enforce that project names themselves must be docker-safe

it's a good idea, need more much more investigations...I take a note.  because currently `add_holohub_application` somehow rely on the fact that app name is a folder name for `add_subdirectory`

PR #1285, opened 2025-12-09
Add the support for using AJA driver 17.6

anishag14:
Update AJA source to latest libajantv2 and add libudev dependency.

## Summary by CodeRabbit

* **Chores**
  * Added an AJA-specific build stage to the container build to enable capture card support.
  * Updated a native AJA library reference to a newer pinned version for improved stability.

* **Documentation**
  * Updated quick-start instructions to show how to build and invoke the AJA-targeted binary.

### DIFF
--- a/applications/orsi/CMakeLists.txt
+++ b/applications/orsi/CMakeLists.txt
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2022-2023 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -143,3 +143,8 @@ RUN echo 'export JREHOME=$(readlink /etc/alternatives/java | sed -e "s/\/bin\/ja
 #
 # --------------------------------------------------------------------------
 FROM holohub-cli AS holohub-dev
+
+RUN apt update \
+    && apt install --no-install-recommends -y \
+        libudev-dev \
+    && rm -rf /var/lib/apt/lists/*
```
tbirdso, line 149: Since most apps do not rely on AJA, could we please move the AJA dependency to a dedicated stage in the Dockerfile that is not part of the default path? See the `yuan-qcap` or `dds` stages above for examples.

The command to run the development environment with AJA build support would then be:
```
./holohub run-container --build-args="--target holohub-aja"
```

anishag14: Done

PR #1288, opened 2025-12-10
Align `download_ngc_data` script with the one in HSDK

bhashemian:
Updating download_ngc_data` script in the HSDK, I noticed that there have been updates to the script in the HSDK that haven’t been propagated here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error handling now halts immediately on missing or invalid inputs, preventing partial runs and unclear outcomes.
  * User-facing messages and exit behavior around dataset naming and help now behave more consistently.
  * Path resolution for the data conversion step has been improved so the process runs reliably regardless of invocation context.

### DIFF
--- a/utilities/download_ngc_data
+++ b/utilities/download_ngc_data
@@ -257,7 +257,7 @@ display_usage() {
 
 if [ "${HELP}" == "YES" ]; then
   display_usage
-  exit 1
+  fatal
 fi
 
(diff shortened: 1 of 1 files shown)


bhashemian:
@tbirdso it seems that the changes introduced in HSDK needs some refactoring. I'll update them here and then update the HSDK as well.

PR #1296, opened 2025-12-11
Upgrade `cpplint` to Avoid Deprecation Warning

bhashemian:
Current deprecation warning:
```console
/workspace/venvqa/lib/python3.12/site-packages/cpplint.py:57: DeprecationWarning: module 'sre_compile' is deprecated
  import sre_compile
```
and here are the list of newly flagged by this upgrade which are filtered:

| Warning Type | Description | Count |
|--------------|-------------|-------|
| `whitespace/newline` | Controlled statements inside brackets of if/for/while clause should be on a separate line | ~470 |
| `build/c++17` | `<filesystem>` is an unapproved C++17 header | 20 |
| `whitespace/indent_namespace` | Do not indent within a namespace | 46 |
| `readability/nolint` | Not in a NOLINT block | 2 |

## Summary by CodeRabbit

* **Chores**
  * Updated code linting configuration to refine quality checks.
  * Upgraded linting tool to the latest version for improved code analysis capabilities.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1304, opened 2025-12-15
Fix `whitespace-newline` Issues Flagged by CPPLINT

bhashemian:
This is a followup PR for https://github.com/nvidia-holoscan/holohub/pull/1296, where this linting issue was reported after upgrading `cpplint`.

| Warning Type | Description | 
|--------------|-------------|
| `whitespace/newline` | Controlled statements inside brackets of if/for/while clause should be on a separate line |

### DIFF
--- a/applications/adv_networking_bench/README.md
+++ b/applications/adv_networking_bench/README.md
@@ -63,24 +63,31 @@ found in this directory.
 
 ### Requirements
 
-This application requires all configuration and requirements from the [Advanced Network library](/operators/advanced_network/README.md).
+This application requires all configuration and requirements from the [Advanced Network library](/operators/advanced_network/README.md) first.
 
 ### Build Instructions
 
-Please refer to the top level Holohub README.md file for information on how to build this application.
+#### DPDK
 
 ```bash
-# Regular build with DPDK and GPUNetIO support
-./holohub build adv_networking_bench --build-args="--target gpunetio" --configure-args="-D ANO_MGR:STRING=gpunetio"
+./holohub build adv_networking_bench --language=cpp
 ```
 
-For **Rivermax** support, follow the prerequisites from the [operators/advanced_network/README.md](/operators/advanced_network/README.md) README, then build the application with the following command:
+#### GPUNetIO
 
 ```bash
-./holohub build adv_networking_bench --build-args="--target rivermax" --configure-args="-D ANO_MGR:STRING=rivermax"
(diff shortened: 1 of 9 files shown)


tbirdso:
> where this linting issue was reported after upgrading cpplint

Hi @bhashemian , is there any chance we can configure cpplint to preserve behavior across versions? Will be painful if we need to reformat across all projects every time we update linting tools

bhashemian:
> > where this linting issue was reported after upgrading cpplint
> 
> Hi @bhashemian , is there any chance we can configure cpplint to preserve behavior across versions? Will be painful if we need to reformat across all projects every time we update linting tools

@tbirdso I agree with you, but linting tools are improving and now support more linting issues. We can either filter them, as I did before, or fix them. Initially, I suppressed them to maintain the same behavior as before, but I thought you were suggesting fixing them.

I don’t think it’s that crucial. My suggestion is that we proceed with this fix and, for future linting issues, decide per case whether to fix or filter them. Does that make sense to you?

tbirdso:
> Initially, I suppressed them to maintain the same behavior as before, but I thought you were suggesting fixing them.

I apologize, my earlier [comment](https://github.com/nvidia-holoscan/holohub/pull/1296#pullrequestreview-3572065744) wasn't clear. Generally my take is:
- Pure styling issues can be suppressed. I wasn't understanding before that we are already suppressing these `whitespace/indent_namespace` warnings and they would not block PRs, thanks for doing that. Based on that, I would consider the status quo sufficient to address these warnings.
- Issues other than styling should be investigated. I was initially concerned about the `<filesystem> is an unapproved C++17 header.  [build/c++17]` warnings in the previous PR, but on review I see that <filesystem> indeed exists in C++17 ([ref](https://en.cppreference.com/w/cpp/filesystem.html)) and the cpplint message is misleading or wrong. As a result I'm OK with the existing suppression and consider this sufficiently addressed under the status quo.

Generally wouldn't be concerned with styling fixes, but I'm concerned that merging style changes across the breadth of projects in this PR could result in HoloHub developers needing to rebase or resolve merge conflicts without tangible benefit.

My preference is that we close this PR and consider the cpplint issues already addressed sufficiently by your previous PR. Is that okay with you?

bhashemian:
Thanks @tbirdso for the clarification. It’s wasted time for me, but it makes sense for Holohub. I agree with less unnecessary restriction.

PR #1308, opened 2025-12-16
Fix Basic Networking Ping Application

bhashemian:
## Summary by CodeRabbit

* **New Features**
  * Added mode-based execution with three modes: flexible (default), receiver (rx), and transmitter (tx).

* **Documentation**
  * Simplified README quick start with rx/tx examples and added language-specific and config-file usage examples; minor wording/heading tweaks.

* **Improvements**
  * Expanded tested SDK compatibility to include 3.9.0.
  * Updated command-line argument syntax reference to use --run-args.

* **Bug Fixes / Behavior**
  * Made receiver/transmitter config detection more permissive and added a default for max burst interval.

### DIFF
--- a/applications/basic_networking_ping/README.md
+++ b/applications/basic_networking_ping/README.md
@@ -56,7 +56,7 @@ There are three launch profiles configured for this application:
 1. **(gdb) basic_networking_ping/cpp RX**: Launch Basic Networking Ping with the [RX configurations](basic_networking_ping_rx.yaml).
 2. **(gdb) basic_networking_ping/cpp TX**: Launch Basic Networking Ping with the [TX configurations](basic_networking_ping_tx.yaml).
 3. **(compound) basic_networking_ping/cpp TX & RX**: Launch both 1 and 2 in parallel.
-   This launch profile launches the receiver follow by the transmitter.
+   This launch profile launches the receiver followed by the transmitter.
 
 ### Python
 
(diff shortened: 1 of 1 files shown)


applications/basic_networking_ping/cpp/metadata.json
```diff
@@ -32,9 +33,29 @@
...
-			"workdir": "holohub_app_bin"
+		"default_mode": "flexible",
+		"modes": {
+			"flexible": {
+				"description": "Start the receiver or transmitter with the specified arguments",
+				"run": {
+					"command": "<holohub_app_bin>/basic_networking_ping",
+					"workdir": "holohub_app_bin"
+				}
+			},
```
tbirdso, line 44: Why not have RX or TX as default? Concerned that the need for a "flexible" mode may point to insufficient CLI flexibility in accepting user argument overrides?

dylan-eustice: Maybe this is just my ignorance of how Holohub uses metadata.json, but it's not clear to me how the `flexible` mode is used within the application. No issues or confusion on the rest of the PR though.

bhashemian: @tbirdso The reason is `--run-args` is just appending arguments to the run command, and cannot replace one. We could potentially modify the application to run in RX or TX mode by default but not from Holohub side, right?

tbirdso: Gotcha, like you said I see that `--run-args` extends the run command, and also that at least the Python `basic_networking_ping` app is not currently designed to handle more than one argument, so passing `--run-args` to any tx/rx pre-defined app is certain to throw a runtime error

Approach seems reasonable for balancing app flexibility with HoloHub CLI behavior today 👍

bhashemian: @dylan-eustice the flexible mode is the default mode so will be invoke when no mode is set. It is just to keep backward-compatibility and flexibility of using custom config file. The expected use as mentioned in the README is:
```
./holohub run basic_networking_ping --run-args another_config.yaml
```
Note that this command will invoke `flexible` mode.

PR #1322, opened 2026-01-02
[NEW APP] BCI Visualization with Kernel Flow2 and Real-Time Volume Rendering

mimiliaogo:
This PR introduces a non-invasive BCI visualization application, developed in collaboration with Kernel, that reconstructs and renders HbO activities of a 3D brain in real time.
## What’s included
1. Kernel reconstruction pipeline: end-to-end processing from fNIRS moments input to reconstructed HbO volumes.
2. Converts the reconstructed HbO output into the format required by the volume renderer.
3. Volume renderer enhancements: enables streaming and multithreaded execution, adds CUDA stream handling, and includes several bug fixes identified during integration.

### TODO
1. Add a more detailed README (Kernel Flow2, fNIRS, and pipeline overview).
2. Refactor and consolidate data inputs and upload datasets to cloud.

## Summary by CodeRabbit

* **New Features**
  * New real-time BCI app: SNIRF NIRS streaming, GPU-accelerated reconstruction, voxel→volume resampling, and interactive 3D volume rendering with camera feedback.
  * Docker support, application manifest/metadata, and pinned Python requirements for reproducible runs.

* **Improvements**
  * Volume renderer: CUDA stream integration, camera-pose caching, and reduced busy-waiting for smoother rendering.
  * Added color-buffer passthrough and Python bindings to avoid rendering bottlenecks.

* **Documentation**
  * Detailed README, example configs, and runtime/config files for setup and transfer-function tuning.


applications/bci_visualization/operators/color_buffer_passthrough/cpp/color_buffer_passthrough.cpp
bhashemian, line 1: Would it be possible to create an explicit `cpp` folder please?

operators/volume_renderer/dataset.hpp
```diff
@@ -56,11 +56,12 @@ class Dataset {
    * then the range is calculated form the data. For example a full range of a uint8 data type is
    * defined by {0.f, 255.f}.
    * @param tensor volume data
+   * @param cuda_stream CUDA stream
    */
   void SetVolume(Types type, const std::array<float, 3>& spacing,
                  const std::array<uint32_t, 3>& permute_axis, const std::array<bool, 3>& flip_axes,
                  const std::vector<clara::viz::Vector2f>& element_range,
-                 const nvidia::gxf::Handle<nvidia::gxf::Tensor>& tensor);
+                 const nvidia::gxf::Handle<nvidia::gxf::Tensor>& tensor, cudaStream_t cuda_stream);
```
tbirdso, line 64: Any anticipated impact to other demo apps already relying on the operator (i.e. `volume_rendering`)?

Can we add a default to preserve API backwards compat? `cudaStream_t cuda_stream = 0`?

operators/volume_renderer/volume_renderer.cpp
```diff
@@ -293,8 +296,24 @@ bool VolumeRendererOp::Impl::receive_volume(InputContext& input, Dataset::Types
...
+    nvidia::gxf::Expected<nvidia::gxf::Entity> entity = volume.value();
+    const gxf_result_t result = cuda_stream_handler_.fromMessage(context.context(), entity);
+    if (result != GXF_SUCCESS) {
+      throw std::runtime_error("Failed to get the CUDA stream from incoming messages");
+    }
+    const cudaStream_t cuda_stream = cuda_stream_handler_.getCudaStream(context.context());
+
+    // MEMORY LEAK FIX: Clear old volumes before receiving new ones for streaming mode
+    // The Dataset class accumulates frames in a vector, which causes unbounded memory growth
+    // For real-time streaming (BCI visualization), we only need the latest frame
```
tbirdso: Nitpick: seems like this applies to any streaming, not specific to BCI demo
```suggestion
    // For real-time streaming, we only need the latest frame
```

operators/volume_renderer/volume_renderer.cpp
```diff
@@ -521,7 +539,7 @@ void VolumeRendererOp::setup(OperatorSpec& spec) {
   spec.input<std::array<uint32_t, 3>>("mask_permute_axis").condition(ConditionType::kNone);
   spec.input<std::array<bool, 3>>("mask_flip_axes").condition(ConditionType::kNone);
 
-  spec.output<holoscan::gxf::Entity>("color_buffer_out");
+  spec.output<holoscan::gxf::Entity>("color_buffer_out").condition(ConditionType::kNone); // (Mimi): remove backpressure to avoid deadlock when running in a loop
```
tbirdso: Could this behavior change break existing apps? I see "fault" documented as the default downstream queue policy ([ref](https://github.com/nvidia-holoscan/holoscan-sdk/blob/main/examples/multi_branch_pipeline/cpp/multi_branch_pipeline.cpp#L79))

Could we please verify for backwards compat with `volume_rendering`? An alternative could be to use a passthrough operator in the BCI demo app to remove backpressure without directly changing `VolumeRendererOp`

operators/volume_renderer/volume_renderer.cpp
```diff
@@ -817,17 +841,24 @@ void VolumeRendererOp::compute(InputContext& input, OutputContext& output,
...
   std::shared_ptr<VideoBufferBlob> depth_buffer_blob;
-  if (depth_buffer) { depth_buffer_blob = std::make_shared<VideoBufferBlob>(depth_buffer); }
-  impl_->image_service_->Render(color_buffer->video_frame_info().width,
-                                color_buffer->video_frame_info().height,
-                                color_buffer_blob,
-                                depth_buffer_blob);
-
-  // then wait for the message with the encoded data
-  const auto image = impl_->image_service_->WaitForRenderedImage();
+  if (impl_->dataset_.GetNumberFrames() > 0) {
```
tbirdso, line 849: For my understanding, what would cause the HolovizOp to tick without any volume provided? Incoming config values?

operators/volume_renderer/volume_renderer.cpp
```diff
@@ -817,17 +841,24 @@ void VolumeRendererOp::compute(InputContext& input, OutputContext& output,
...
+                                  color_buffer_blob,
+                                  depth_buffer_blob);
+
+    // then wait for the message with the encoded data
+    const auto image = impl_->image_service_->WaitForRenderedImage();
+  } else {
+    // no density volume, sleep for the time slot
+    clara::viz::RenderSettingsInterface::AccessGuard access(impl_->render_settings_interface_);
+    HOLOSCAN_LOG_INFO("No density volume, sleeping for {} milliseconds", access->time_slot.Get());
+    std::this_thread::sleep_for(std::chrono::milliseconds(static_cast<int>(access->time_slot.Get())));
```
tbirdso: Why sleep if there is nothing to render? Does this block other operators from using the thread?

applications/bci_visualization/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+project(bci_visualization LANGUAGES NONE)
+
+find_package(holoscan 2.0 REQUIRED CONFIG
+             PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")
+
+add_subdirectory(operators)
+
+# Enable the operators
+add_library(bci_visualization INTERFACE)
+target_link_libraries(bci_visualization INTERFACE holoscan::core holoscan::ops::holoviz color_buffer_passthrough)
```
tbirdso: I don't think these are needed for the Python app. Could you please help confirm?

```suggestion
```

applications/bci_visualization/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE} AS base
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Install curl for downloading data files
+RUN apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/*
+
+# Add schemas directory to PYTHONPATH to enable importing operators
+ENV PYTHONPATH=$PYTHONPATH:/workspace/holohub
```
tbirdso: Not sure about this, typically C++ operators would be built in a directory such as `/workspace/holohub/build/bci_visualization/python/operators/holohub` and pure Python modules would be available in subdirectories such as `/workspace/holohub/applications/bci_visualization/operators`.

Could you comment on what behavior you're seeing without this PYTHONPATH setting?

applications/bci_visualization/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+# Install Python dependencies
+RUN pip install nibabel nilearn
+
+# Install Kernel's dependencies
+RUN pip install \
+    numpy \
+    scipy \
+    h5py
```
tbirdso: Suggest moving to a `requirements.txt` file for potential reuse outside of docker container environments

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+from holoscan.resources import CudaStreamPool, UnboundedAllocator
+from holoscan.schedulers import EventBasedScheduler, MultiThreadScheduler
+# Import local operators
+from operators.voxel_stream_to_volume import VoxelStreamToVolumeOp
+
+from holohub.color_buffer_passthrough import ColorBufferPassthroughOp
+from holohub.volume_renderer import VolumeRendererOp
+
+class BciVisualizationApp(Application):
+    """BCI Visualization Application with ClaraViz."""
```
tbirdso: Could you please add another line or two expanding on the application overview? Would be nice to include the compatible device (Kernel Flow2) and high level behavior (replay device recording, reconstruct activation map volume, render with ClaraViz / HoloViz). Can keep detailed explanation in README

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+            name="cuda_stream",
+            dev_id=0,
+            stream_flags=0,
+            stream_priority=0,
+            reserved_size=1,
+            max_size=5,
+        )
+
+        # Selection
+        selected_channel = 0  # 0: HbO, 1: HbR
```
tbirdso: Should this be an application parameter?

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+        if self._density_max:
+            volume_renderer_args["density_max"] = self._density_max
+
+        volume_renderer = VolumeRendererOp(
+            self,
+            name="volume_renderer",
+            config_file=self._rendering_config,
+            allocator=volume_allocator,
+            alloc_width=1024,
+            alloc_height=768,
```
tbirdso: For standard practice, consider moving these to an application YAML configuration file

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+        color_buffer_passthrough = ColorBufferPassthroughOp(
+            self,
+            name="color_buffer_passthrough",
+        )
+
+        # Connect operators
+
+        kernel_data = Path("/workspace/holohub/data/kernel")
+        from streams.snirf import SNIRFStream
+        stream = SNIRFStream(kernel_data / "data.snirf")
```
tbirdso: Suggest prioritizing data paths from app YAML file or `HOLOSCAN_INPUT_PATH` standard environment varaible, with fallback to inline expected filepaths

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+        from reconstruction import ReconstructionApplication
+        reconstruction_application = ReconstructionApplication(
+            stream=stream,
+            jacobian_path=kernel_data / "flow_mega_jacobian.npy",
+            channel_mapping_path=kernel_data / "flow_channel_map.json",
+            voxel_info_dir=kernel_data / "voxel_info",
+            coefficients_path=kernel_data / "extinction_coefficients_mua.csv",
+            use_gpu=True,
+        )
+        reconstruction_application.compose(self, voxel_to_volume)
```
tbirdso: Interesting, can't recall seeing nested applications like this, is this supported?

Does this result in two Holoscan applications running in parallel, with two distinct executors? Any advantage to having two apps instead of one unified pipeline? Any performance concerns?

Should we consider exploring Holoscan SDK subgraphs for reuse instead of nesting applications? https://docs.nvidia.com/holoscan/sdk-user-guide/api/cpp/classholoscan_1_1subgraph.html

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+
+        # voxel_to_volume → volume_renderer
+        self.add_flow(voxel_to_volume, volume_renderer, {
+            ("volume", "density_volume"),
+            ("spacing", "density_spacing"),
+            ("permute_axis", "density_permute_axis"),
+            ("flip_axes", "density_flip_axes"),
+        })
+        # Add mask connections to VolumeRendererOp
+        self.add_flow(voxel_to_volume, volume_renderer, {
```
tbirdso: Any reason to split up into two `add_flow` calls as done here, as opposed to all in one call?

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+        # volume_renderer ↔ holoviz
+        self.add_flow(volume_renderer, color_buffer_passthrough,
+                      {("color_buffer_out", "color_buffer_in")})
+        self.add_flow(color_buffer_passthrough, holoviz, {("color_buffer_out", "receivers")})
+        self.add_flow(holoviz, volume_renderer, {("camera_pose_output", "camera_pose")})
+
+
+def main():
+    
+
```
tbirdso: Nitpick: spacing
```suggestion
```

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+    parser.add_argument(
+        "-a",
+        "--density_max",
+        action="store",
+        type=int,
+        dest="density_max",
+        help="Set the maximum of the density element values. If not set this is calculated from the"
+        "volume data. In practice CT volumes have a maximum value of 3071 which corresponds to"
+        "the upper value of the Hounsfield scale range usually used.",
+    )
```
tbirdso: Do we expect users will typically want to provide these on the command line? Or, could we consolidate in an app YAML file?

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,240 @@
...
+        help="Path to the mask NIfTI file containing 3D integer labels (I, J, K). Optional.",
+    )
+
+    parser.add_argument(
+        "-h", "--help", action="help", default=argparse.SUPPRESS, help="Help message"
+    )
+    
+    args = parser.parse_args()
+    
+    # Parse roi_labels from comma-separated string to list of integers
```
tbirdso: Might be more user friendly to read/write a list in an app YAML file rather than comma-separated list on command line

applications/bci_visualization/metadata.json
```diff
@@ -0,0 +1,39 @@
...
+			"1.0": "Initial Python Release"
+		},
+		"holoscan_sdk": {
+			"minimum_required_version": "0.6.0",
+			"tested_versions": [
+				"0.6.0",
+				"1.0.0",
+				"2.0.0",
+				"2.1.0"
+			]
```
tbirdso: Is this accurate?

applications/bci_visualization/metadata.json
```diff
@@ -0,0 +1,39 @@
...
+		],
+		"tags": ["Healthcare AI", "Visualization", "BCI", "Hemodynamic", "Volume Rendering", "Neuroscience", "Python"],
+		"ranking": 1,
+		"requirements": {
+			"python": ["numpy", "cupy"]
+		},
+		"run": {
+			"command": "python3 bci_visualization.py --config <holohub_data_dir>/bci_visualization/config.json --density_min -100 --density_max 100 --mask_path <holohub_data_dir>/bci_visualization/anatomy_labels_high_res.nii.gz",
+			"workdir": "holohub_app_source"
+		}
```
tbirdso, line 75: Suggest omitting most parameters from command string and instead providing as default, overridable values with argparse in Python, so that the app runs out of the box but users can also provide their own values as desired.

I.e. use the command `python3 bci_visualization.py` and expect that either of the following are valid:
```
./holohub run bci_visualization   # run with defaults
./holohub run bci_visualization --run-args="--density_min -110 --density_max 110 ..."    # run with user-specified overrides
```

If you think it would be valuable, you could also define multiple presets in `metadata.json` to run with different default settings, such as `./holohub run bci_visualization hbr`, etc

applications/bci_visualization/operators/color_buffer_passthrough/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+  EXPORT holoscan-ops
+)
+
+if(HOLOHUB_BUILD_PYTHON)
+  add_subdirectory(python)
+endif()
+
+
+
+
```
tbirdso, line 50: Nitpick: spacing
```suggestion

```

applications/bci_visualization/operators/color_buffer_passthrough/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+  PRIVATE
+  holoscan::core
+)
+
+target_include_directories(color_buffer_passthrough INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
+
+install(
+  TARGETS color_buffer_passthrough
+  EXPORT holoscan-ops
+)
```
tbirdso, line 42: Does this app need to be portable, i.e. build in one environment and run in another? If not, can remove `install()` functionality to reduce maintenance burden

```suggestion
```

applications/bci_visualization/operators/color_buffer_passthrough/cpp/color_buffer_passthrough.cpp
```diff
@@ -0,0 +1,44 @@
...
+  auto cuda_streams = input.receive_cuda_streams("color_buffer_in");
+
+  output.emit(color_message.value(), "color_buffer_out");
+}
+
+}  // namespace holoscan::ops
+
+
+
+
```
tbirdso, line 44: Nitpick: spacing, here and in other files
```suggestion

```

applications/bci_visualization/operators/color_buffer_passthrough/color_buffer_passthrough.hpp
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef BCI_VISUALIZATION_COLOR_BUFFER_PASSTHROUGH_HPP
+#define BCI_VISUALIZATION_COLOR_BUFFER_PASSTHROUGH_HPP
```
tbirdso: Nitpick: suggest `#pragma once` to reduce include guard maintenance

applications/bci_visualization/operators/reconstruction/__init__.py
```diff
@@ -0,0 +1,21 @@
+from .build_rhs_operator import BuildRHSOperator
```
tbirdso, line 6: Needs copyright header (here and other files)

applications/bci_visualization/operators/reconstruction/build_rhs_operator.py
```diff
@@ -0,0 +1,191 @@
...
+    OutputContext,
+)
+from ..stream import SampleOutput
+from .types import BuildRHSOutput, VoxelMetadata
+
+logger = logging.getLogger(__name__)
+
+
+class BuildRHSOperator(Operator):
+    """Convert realtime moments tensors into trimmed RHS/Jacobian batches."""
```
tbirdso:
```suggestion
    """Convert realtime moments tensors into trimmed Right-Hand Side (RHS)/Jacobian batches."""
```

tbirdso (commented):
Looking good so far! Took a first high level review pass, will take a second pass for Python app details

applications/bci_visualization/utils/reconstruction/assets.py
```diff
@@ -0,0 +1,124 @@
...
+from .data_utils import get_channel_mask
+from .hbo import ExtinctionCoefficient, HbO 
+from .types import ChannelHeadsetMapping
+
+logger = logging.getLogger(__name__)
+USE_GPU_DEFAULT = True
+
+
+@dataclass(frozen=True)
+class Assets:
```
tbirdso, line 22: It might not be easy to find classes and functions defined in `__init__.py`. Can we move definitions to a dedicated file for development, and then possibly re-import here if desired?

applications/bci_visualization/processing/reconstruction/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+from .hbo import ExtinctionCoefficient, HbO 
+from .types import ChannelHeadsetMapping
+
+logger = logging.getLogger(__name__)
+USE_GPU_DEFAULT = True
+
+
+@dataclass(frozen=True)
+class Assets:
+    mega_jacobian: NDArray[np.float32]  # float32 array (feature-channels, voxels)
```
tbirdso: What is this asset and how should it be used? Could you please provide a more detailed docstring?

applications/bci_visualization/utils/reconstruction/assets.py
```diff
@@ -0,0 +1,124 @@
...
+    mega_jacobian: NDArray[np.float32]  # float32 array (feature-channels, voxels)
+    channel_mapping: ChannelHeadsetMapping  # mapping of channels to jacobian channel indices
+    mua: NDArray[np.float32]  # absorption coefficients
+    musp: NDArray[np.float32]  # reduced scattering coefficients
+    idxs_significant_voxels: NDArray[np.int_]  # indices of significant voxels in mega_jacobian
+    ijk: NDArray[np.float32]  # voxel ijk coordinates
+    xyz: NDArray[np.float32]  # voxel xyz coordinates
```
tbirdso, line 39: Is this a lookup array? If so:
- Is this guaranteed to correspond to the underlying memory ordering (row/column major)?
- Is there a closed form for predicting ijk / xyz from a given voxel index?
- Is XYZ in reference to a vertex or to the voxel center?
- Was this structure inspired by any existing medical image standard definition? Would be good to document for users

applications/bci_visualization/processing/reconstruction/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+    xyz: NDArray[np.float32]  # voxel xyz coordinates
+    wavelengths: NDArray[np.int_]  # wavelengths
+    resolution: Tuple[float, float, float]
+    extinction_coefficients: list[ExtinctionCoefficient]  # HbO and HbR extinction coefficients
+
+
+_assets: Assets | None = None
+
+REG_DEFAULT = 0.1
+RESHAPING_ORDER = "F"
```
tbirdso: Please add descriptive docstrings

applications/bci_visualization/utils/reconstruction/assets.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def _load_assets(
+    *,
+    mega_jacobian_path: pathlib.Path | str,
+    channel_mapping_path: pathlib.Path | str,
+    coefficients_path: pathlib.Path | str,
+    voxel_info_dir: pathlib.Path | str,
+) -> Assets:
+    """Load large reconstruction assets on demand.
```
tbirdso, line 56: Nitpick: suggest adding a "Throws" section to mention possible errors from `np.load`, such as if any assets do not exist in the provided path location(s). https://numpy.org/devdocs/reference/generated/numpy.load.html

applications/bci_visualization/utils/reconstruction/assets.py
```diff
@@ -0,0 +1,124 @@
...
+    _idxs_significant_voxels = np.load(pathlib.Path(idxs_significant_voxels_path))
+    _ijk = np.load(pathlib.Path(ijk_path))
+    _xyz = np.load(pathlib.Path(xyz_path))
+    _wavelengths = np.load(pathlib.Path(wavelengths_path))
+    _resolution = tuple(np.load(pathlib.Path(resolution_path)).tolist())
+
+    logger.info("Reconstruction assets initialization complete.")
+
+    return Assets(
+        mega_jacobian=cast(NDArray[np.float32], _mega_jacobian),
```
tbirdso, line 110: Not sure if `cast` is needed for `NDArray` as done here, or should the result of `np.load` typically be type compliant already?

applications/bci_visualization/utils/reconstruction/assets.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def get_assets(
+    jacobian_path: pathlib.Path | str,
+    channel_mapping_path: pathlib.Path | str,
+    voxel_info_dir: pathlib.Path | str,
+    coefficients_path: pathlib.Path | str,
+) -> Assets:
+    global _assets
+    if _assets is None:
```
tbirdso, line 130: Nitpick: might be nice to check for existence of each asset, in case there is ever a situation where assets have been only partially initialized?

applications/bci_visualization/processing/reconstruction/data_utils.py
```diff
@@ -0,0 +1,47 @@
...
+            detectors = srcs[str(src)][str(det_module)]
+            jacobian_index = detectors[str(det)]
+        except KeyError:
+            raise ValueError(
+                f"Channel without jacobian mapping (src_module={src_module}, src={src}, det_module={det_module}, det={det})",
+            )
+
+        channel_mask[jacobian_index] = True
+
+    assert np.any(channel_mask)
```
tbirdso: Seems like all-zero channel mask is unexpected but not necessarily at the level of a critical error. Could we log an error or warning rather than `assert`?

applications/bci_visualization/processing/reconstruction/gpu.py
```diff
@@ -0,0 +1,30 @@
...
+
+logger = logging.getLogger(__name__)
+
+try:
+    import cupy  # type: ignore
+
+    logger.info("CuPy is available for GPU computations.")
+except ImportError:
+    cupy = None
+    logger.info("CuPy is not available.")
```
tbirdso: Is there any advantage to supporting environments without CuPy? If not, could we simplify with a hard requirement?

```suggestion
import cupy
```

applications/bci_visualization/processing/reconstruction/gpu.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def get_array_module(use_gpu: bool = False) -> tuple[ModuleType, bool]:
+    """Get the appropriate array module (NumPy or CuPy) based on GPU availability.
```
tbirdso: Holoscan SDK does not generally support CPU-only environments. Suggest standardizing to drop CPU-only support as not supported by the platform, not a hard blocker for this app though.

applications/bci_visualization/processing/reconstruction/gpu.py
```diff
@@ -0,0 +1,30 @@
...
+    Args:
+        use_gpu: Whether to attempt to use GPU arrays.
+    Returns:
+        A tuple containing the array module and a boolean indicating if GPU is used.
+    """
+    # GPU-only refactor: reconstruction pipeline should always run on CuPy when requested.
+    if not use_gpu:
+        raise ValueError("GPU-only pipeline: get_array_module(use_gpu=False) is not supported")
```
tbirdso: Nitpick: "GPU-only" seems misleading, seems like proposed modes are either "CPU + GPU" or "CPU-only"? Data may not reside primarily on the CPU, but the CPU is still used for scheduling regardless. Suggest renaming

applications/bci_visualization/utils/reconstruction/hbo.py
```diff
@@ -0,0 +1,129 @@
...
+import csv
+from pathlib import Path
+from typing import NamedTuple, Tuple
+
+import numpy as np
+from numpy.typing import NDArray
+
+from .gpu import get_array_module
+
+class ExtinctionCoefficient(NamedTuple):
```
tbirdso, line 15: Please add documentation

applications/bci_visualization/processing/reconstruction/hbo.py
```diff
@@ -0,0 +1,129 @@
...
+    HbO: float
+    deoxyHb: float
+    Water: float
+    Lipids: float
+    LuTex: float
+    GdTex: float
+
+    @classmethod
+    def from_csv(cls, path: Path) -> list[ExtinctionCoefficient]:
+        return [
```
tbirdso: Why a list? Is the ordering meaningful here? Any reason to prefer a list of tuples over some other array structure?

applications/bci_visualization/processing/reconstruction/hbo.py
```diff
@@ -0,0 +1,129 @@
...
+            Wavelengths (nm).
+        idxs_significant_voxels
+            Indices of significant voxels.
+
+        Returns
+        -------
+        data_hbo : NDArray[np.float32]
+            Oxygenated hemoglobin in voxel space.
+        data_hbr : NDArray[np.float32]
+            Deoxygenated hemoglobin in voxel space.
```
tbirdso: What are the units on HbO / HbR? µM?

applications/bci_visualization/processing/reconstruction/hbo.py
```diff
@@ -0,0 +1,129 @@
...
+        """Converts mua to Hb in voxel space.
+
+        Parameters
+        ----------
+        data_mua : NDArray[np.float32]
+            µa reconstructed values in voxel space.
+        wavelengths : tuple
+            Wavelengths (nm).
+        idxs_significant_voxels
+            Indices of significant voxels.
```
tbirdso: What does "significant" mean here?

applications/bci_visualization/streams/__init__.py
```diff
@@ -0,0 +1,2 @@
+def dist3d(x1: float, y1: float, z1: float, x2: float, y2: float, z2: float) -> float:
```
tbirdso: Not easy to find definitions in `__init__.py`, could we move this to a subfile in the module?

applications/bci_visualization/streams/snirf.py
```diff
@@ -0,0 +1,150 @@
...
+from . import dist3d
+from .base_nirs import BaseNirsStream
+
+logger = logging.getLogger(__name__)
+
+NUM_MOMENTS = 3
+NUM_WAVELENGTHS = 2
+
+
+class SNIRFChannel(NamedTuple):
```
tbirdso, line 22: Please add docstrings (Copilot might help)

applications/bci_visualization/streams/snirf.py
```diff
@@ -0,0 +1,150 @@
...
+    wavelength: int
+    source_module: int
+    source_number: int
+    detector_module: int
+    detector_number: int
+    sds: float
+
+
+class SNIRFStream(BaseNirsStream):
+    def __init__(self, snirf_file: Path | str) -> None:
```
tbirdso, line 41: Please add docstring mentioning the SNIRF format spec: https://github.com/fNIRS/snirf

Would be nice to also have functional test coverage to verify basic SNIRF reader + streaming coverage independent of the demo pipeline. Suggest PyTest to align with other HoloHub Python apps

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
```diff
@@ -0,0 +1,302 @@
...
+        self.affine = None
+
+        # Metadata, set from the first frame, reused for subsequent frames
+        self.dims = None  # np.array([X, Y, Z], dtype=np.uint32)
+        self.out_spacing = None  # np.ndarray float32 (3,)
+        self.permute_axis = None  # np.ndarray uint32 (3,)
+        self.flip_axes = None  # np.ndarray bool (3,)
+        self.roi_mask = None  # np.ndarray bool (I, J, K)
+
+        # Labels for brain anatomy (DEPRECATED)
```
tbirdso: Keep or remove?

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
```diff
@@ -0,0 +1,302 @@
...
+        spec.output("mask_flip_axes").condition(ConditionType.NONE)
+
+    def compute(self, op_input, op_output, context):
+        # Receive Hb voxel data (cupy array)
+        hb_voxel = op_input.receive("hb_voxel_data")  # (I, J, K)
+        cuda_stream = op_input.receive_cuda_stream("hb_voxel_data")
+
+        # Check voxel data is valid
+        if not isinstance(hb_voxel, cp.ndarray):
+            raise ValueError("VoxelStreamToVolume: Invalid voxel data, expected cupy array")
```
tbirdso: Nitpick: suggest printing the received type in the error message for debugging

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
```diff
@@ -0,0 +1,302 @@
...
+                f"shape: {self.mask_voxel_raw.shape}, values: {np.unique(self.mask_voxel_raw)}"
+            )
+        except Exception as e:
+            raise RuntimeError(
+                f"VoxelStreamToVolume: Failed to load mask NIfTI '{self.mask_nifti_path}': {e}"
+            ) from e
+
+    def setup(self, spec: OperatorSpec):
+        spec.input("affine_4x4").condition(ConditionType.NONE)  # (4, 4), only emit at the first frame
+        spec.input("hb_voxel_data")  # (I, J, K, n_channels)
```
tbirdso: Specified as ndim == 4 here, but line 108 below checks for ndim == 3 ? Is this correct?

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
```diff
@@ -0,0 +1,302 @@
...
+        # 1. Get the Orientation String
+        # nibabel returns where the axis points TO (e.g., 'RAS')
+        to_codes = aff2axcodes(affine_4x4)
+        print("to_codes: ", to_codes)
+
+        # EXPLANATION OF MAPPING:
+        # NIfTI API (nifti_dmat44_to_orientation) returns constants like NIFTI_L2R ("Left to Right").
+        # The C++ code maps NIFTI_L2R -> 'L' (The "From" direction).
+        # Nibabel returns 'R' (The "To" direction) for that same axis.
+        # Therefore, to generate the string the C++ code expects, we must invert the Nibabel codes.
```
tbirdso: Nitpick: Not sure I follow this motivation. It doesn't seem like any to/from direction strings are returned from this function, but rather a set of axes permutations and flips. Couldn't that be handled internally from the Nibabel convention without needing to reference the `volume_loader` / NIFTI C++ convention?

Not a blocker, if it works as-is then OK to move forward. If needed, would be good to include references to C++ spec:
- https://github.com/NIFTI-Imaging/nifti_clib/blob/master/nifti2/nifti2_io.h#L570
- https://github.com/nvidia-holoscan/holohub/blob/main/operators/volume_loader/nifti_loader.cpp#L39

applications/bci_visualization/operators/reconstruction/build_rhs_operator.py
```diff
@@ -0,0 +1,191 @@
...
+
+            if self._jacobian_cache is None:
+                channel_indices = self._get_channel_indices(flowaxis_optodes)
+                channel_indices_gpu = cp.asarray(channel_indices, dtype=cp.int64)
+
+                jacobians = self._mega_jacobians_gpu[channel_indices_gpu, :, :, :, :]  # 5d
+
+                # swap axes so it's features, channels first
+                jacobians = jacobians.transpose(1, 0, 2, 3, 4)
+                # reshape to 3d and use F ordering
```
tbirdso: Nitpick
```suggestion
                # reshape to 3d and use Fortran-style ordering
```

applications/bci_visualization/metadata.json
```diff
@@ -0,0 +1,39 @@
+{
+	"application": {
+		"name": "BCI Visualization with ClaraViz",
+		"authors": [
+			{
+				"name": "Holoscan Team",
+				"affiliation": "NVIDIA"
+			}
+		],
```
tbirdso, line 29:
```suggestion
			{
				"name": "Holoscan Team",
				"affiliation": "NVIDIA"
			},
			{
			    "name": "Kernel Team",
			    "affiliation": "Kernel"
			},
			{
			    "name": "Mimi Liao",
			    "affiliation": "NVIDIA"
			},
			{
			    "name": "Gabe Lerner",
			    "affiliation": "Kernel"
			}
		],
```

tbirdso:
+ @gabelerner-kernel could you please help review as well?

applications/bci_visualization/Dockerfile
techops-kernel: i'll open a PR to add the real time SDK package and code here after me merge this one

applications/bci_visualization/operators/reconstruction/normalize_operator.py
```diff
@@ -0,0 +1,217 @@
...
+            for idx_feature in range(num_features):
+                row_normalizers[idx_feature::num_features, wavelength_idx] = HARD_CODED_NORMALIZERS[
+                    wavelength_idx
+                ][idx_feature]
+
+        assert not xp.any(xp.isnan(row_normalizers))
+        self._hard_coded_row_normalizers_cache = row_normalizers
+        return row_normalizers
+
+    def _compute_row_normalizers(
```
techops-kernel: we can always rm this and default to the hard coded normalizers method for this repo and add a comment explaining how people can get their own normalization consts (just take them from an offline NIFTI for example)

applications/bci_visualization/operators/reconstruction/normalize_operator.py
```diff
@@ -0,0 +1,217 @@
...
+        return row_normalizers
+
+    def _normalize_batch(self, batch: BuildRHSOutput) -> Tuple[List[WavelengthSystem], int] | None:
+        num_cols = batch.data_jacobians.shape[-1]
+        num_significant = int(batch.idxs_significant_voxels.size)
+        num_absorbers, remainder = divmod(num_cols, num_significant)
+        assert not remainder
+
+        # GPU-only: always use CuPy, and rely on upstream CUDA stream propagation for correctness.
+        xp = cp
```
techops-kernel: if we're gonna always use cupy can just rename `xp` to `cp` in this file

applications/bci_visualization/operators/reconstruction/solver_operator.py
```diff
@@ -0,0 +1,95 @@
...
+            result = self._solve_batch(batch)
+            op_output.emit(result, "result")
+
+    def _solve_batch(self, batch: NormalizedSolveBatch) -> SolverResult:
+        num_wavelengths = len(batch.systems)
+        num_significant_voxels = int(batch.idxs_significant_voxels.size)
+        num_cols_expected = batch.num_absorbers * num_significant_voxels
+
+        # GPU-only: always use CuPy.
+        xp = cp
```
techops-kernel: same here can just rename `xp` to `cp`

applications/bci_visualization/processing/reconstruction/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+import numpy as np
+from numpy.typing import NDArray
+
+# reexport
+from .data_utils import get_channel_mask
+from .hbo import ExtinctionCoefficient, HbO 
+from .types import ChannelHeadsetMapping
+
+logger = logging.getLogger(__name__)
+USE_GPU_DEFAULT = True
```
techops-kernel: can just remove this const and also prob remove the `gpu.py` file/interface

applications/bci_visualization/utils/reconstruction/assets.py
techops-kernel: yea i think that's directionally right, can just be in `assets.py` or something

applications/bci_visualization/processing/reconstruction/__init__.py
techops-kernel: this is a huge sparse 2D matrix where first dim is feature-channel indices and second dim is voxel indices. the order of the feature-channel indices comes from `_get_channel_indices`, voxel indices are just the indices of the ijk in order i believe

applications/bci_visualization/processing/reconstruction/__init__.py
techops-kernel: i'm not sure why this is even a const in our code, we can prob just hardcode the "F" for fortran order in the 2 places it's used

applications/bci_visualization/utils/reconstruction/assets.py
techops-kernel: it's type compliant but the linter doesn't know what type a file would be in code, if we don't really care for type annotations we can remove the cast

applications/bci_visualization/processing/reconstruction/data_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+import numpy as np
+from numpy.typing import NDArray
+
+from .types import ChannelHeadsetMapping
+
+logger = logging.getLogger(__name__)
+
+
+def get_channel_mask(
```
techops-kernel: @mimiliaogo i don't think this method is used anymore since your refactor

applications/bci_visualization/processing/reconstruction/data_utils.py
techops-kernel: it would be an error because we'd essentially select nothing out of the jacobian and there wouldn't be any outputs downstream in the pipeline. the reason it could happen is if your channel data doesn't match what's on disk w/ the assets. a subreason could be that you just have no good channels (not coupled well) or you're using the wrong device type (for example the data comes from a Flow 2.0 but the jacobian is from a Flow 2.1)

applications/bci_visualization/processing/reconstruction/gpu.py
techops-kernel: i'd just remove this whole file, it seems like we only want to support cp going forward in this repo

applications/bci_visualization/processing/reconstruction/hbo.py
techops-kernel: i think one can refactor it to be a dict with a wavelength key and then below on line 73 when we do this:
```
        wavelength_rows = [ext for ext in self._coefficients if ext.Wavelength == round(wavelength)]
```
it would be a lookup in the dict

a lot of this is just copy/pasted code from our internal repo but we can certainly clean it up here!

applications/bci_visualization/utils/reconstruction/assets.py
techops-kernel: not all voxels in the full i*j*k cube are in the ijk/xyz vector, so there's no closed form, they're just the voxels we chose and are used for creating the affine. NIFTI is the inspiration.

applications/bci_visualization/processing/reconstruction/hbo.py
techops-kernel: yea

applications/bci_visualization/processing/reconstruction/hbo.py
techops-kernel: Significant means that the sensitivity is not negligible. The threshold for significant is defined as 1e-3 (have to double check this number) of max sensitivity (jacobian) across voxels.

operators/volume_renderer/dataset.hpp
mimiliaogo: It is already handled by the underlying implementation - if no input stream available, it will return the default stream (cudaStreamDefault).

operators/volume_renderer/volume_renderer.cpp
mimiliaogo: This happens only at the very beginning to wait for **first** available volumes. After this, volume is cached and rendered even if no volume comes in, so no busy waiting anymore.

In our case, the first volume appears after reconstruction pipeline finishes collecting baseline.

No blocking because we have multiple threads running concurrently.

operators/volume_renderer/volume_renderer.cpp
mimiliaogo: HolovizOp only ticks when there's render input. Volume renderer will always emit color_message regardless. In our case, it shows all black at the beginning. And since Holoviz ticks with volume renderer at high FPS, no "foce-quit" window pops out.

operators/volume_renderer/volume_renderer.cpp
mimiliaogo: I verified that the volume_rendering app still works correctly with the current changes. However, I agree that modifying the output port directly in operator introduces risk if it's running in a concurrent pipelines which will cause the 'receive errors' due to default queue policy. That said, backpressure remains an issue when volume renderer and downstream operator is running in a loop concurrently - user needs to handle themselves.

I've moved the changes outside of the operator in next commit.

applications/bci_visualization/Dockerfile
mimiliaogo: @techops-kernel Is the real-time SDK for Flow2 device?

applications/bci_visualization/bci_visualization.py
mimiliaogo: Will unify as single app.

operators/volume_renderer/volume_renderer.cpp
tbirdso: Thank you!

operators/volume_renderer/volume_renderer.cpp
tbirdso: > In our case, it shows all black at the beginning.

Got it, does this also apply to other apps using `VolumeRendererOp` as well? Will the first frame always be empty/black in any scenario, or does this change on line 846 risk omitting valid first frames in other apps?

If this is app-specific behavior, another passthrough op with custom handling might be a better fix than modifying `VolumeRendererOp`

operators/volume_renderer/volume_renderer.cpp
tbirdso: Got it - I wasn't understanding initially that `time_slot` is the expected amount of time that `VolumeRendererOp` will take to render. I see now in https://github.com/NVIDIA/clara-viz/blob/main/src/claraviz/interface/RenderSettingsInterface.h#L82:

> Time slot for a frame in ms. If rendering single images, the renderer returns after this time.
     If rendering a video stream the time is clamped to the frame time.
     If the value is 0.0 the time slot is ignored.

Makes sense to hold the thread resource in `VolumeRendererOp` for that period of time in the expectation that the density volume input will be valid on the next frame. No concerns 👍

applications/bci_visualization/Dockerfile
techops-kernel: @mimiliaogo  yes correct! i'll open a PR after this one is in. it'll be a new "stream" python file and then a check to choose between SDK and SNIRF stream (SDK by default if you don't provide snirf path else snirf)

applications/bci_visualization/streams/snirf.py
gabelerner-kernel: RE: test, i can add it in a subsequent PR, although it would require committing a SNIRF file to the repo which we'd try to keep small, or we can switch to a library like pysnirf instead of my manual code

applications/bci_visualization/streams/__init__.py
gabelerner-kernel: it's unused in this repo will remove

applications/bci_visualization/utils/reconstruction/assets.py
gabelerner-kernel: i don't think there will be, we can't run the pipeline partially

applications/bci_visualization/README.md
```diff
@@ -0,0 +1,54 @@
+# BCI Visualization
+
+## Running
+1. Download data
+   * Please download data from [here](https://drive.google.com/drive/folders/1RpQ6UzjIZAr90FdW9VIbtTFYR6-up7w2) and put everything under `data/bci_visualization`.
```
bhashemian: The download link is not accessible for public. Can we hosted them somewhere else?
We cannot give public access to the content on NVIDIA GDrive.

applications/bci_visualization/README.md
```diff
@@ -0,0 +1,54 @@
...
+   * Includes an activation volume and a segmentation volume.
```
bhashemian:
```suggestion
   * The data includes an activation volume and a segmentation volume.
```

applications/bci_visualization/README.md
```diff
@@ -0,0 +1,54 @@
...
+2. Run application
+   ```bash
+   ./holohub run bci_visualization
+   ```
+   This command will build the docker and run the application.
+
+## Expected Results
+
+![Example output for BCI Visualization](docs/brain.png)
```
bhashemian: `docs/brain.png` doesn't exist! Could you please check on that?

applications/bci_visualization/README.md
bhashemian: Could you please provide the expected folder structure of the data?

applications/bci_visualization/requirements.txt
```diff
@@ -0,0 +1,4 @@
+nibabel==5.3.2
+numpy==2.3.3
+scipy==1.16.3
+h5py==3.15.1
```
bhashemian:
```suggestion
h5py==3.15.1

```

applications/bci_visualization/streams/__init__.py
bhashemian, line 1: To reduce cluttering, can we please move `streams` under `processing` (and maybe rename it to something more generic), so we can have:
```
bci_visualization/processing/
├── __init__.py
├── reconstruction
│   ├── __init__.py
│   ├── data_utils.py
│   ├── gpu.py
│   ├── hbo.py
│   ├── reg_inv.py
│   └── types.py
└── streams
    ├── __init__.py
    ├── base_nirs.py
    └── snirf.py
```

bhashemian (commented):
Thanks @mimiliaogo for submitting this PR. I reviewed it from the user experience perspective and made a few comments to address.

applications/bci_visualization/README.md
mimiliaogo: We'll have a new README soon, this is just temporary

applications/bci_visualization/streams/__init__.py
mimiliaogo: Streams shouldn't be put in processing, streams are responsible for simulating the input stream.
Agree `processing` is confusing, will rename it to `utils`.

operators/volume_renderer/volume_renderer.cpp
mimiliaogo: For normal application which output density volumes from the very beginning, it will just do normal render as the if condition do.

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
mimiliaogo: yea, the mapping is not necessary.

applications/bci_visualization/README.md
```diff
@@ -0,0 +1,231 @@
...
+```bash
+./holohub run bci_visualization
+```
+
+### Expected Data Folder Structure
+
+After downloading and extracting the dataset, your `data/bci_visualization` folder should have this structure:
+
+```
+data/bci_visualization/
```
mimiliaogo: @techops-kernel Can you help review if the below description look correct?

applications/bci_visualization/docs/brain_activity_example.gif
mimiliaogo, line 1: TODO: upload a better resolution gif

applications/bci_visualization/utils/reconstruction/assets.py
```diff
@@ -0,0 +1,139 @@
...
+
+from .hbo import ExtinctionCoefficient
+from .types import ChannelHeadsetMapping
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
+class Assets:
+    # float32 array (channels, features, wavelengths, voxels, simulation_types)
```
mimiliaogo, line 23: @techops-kernel Could we add docstrings clarifying the semantics and shapes of these fields?

operators/volume_renderer/volume_renderer.cpp
tbirdso: Okay sounds good, thanks!

applications/bci_visualization/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+In this Holohub application, we demonstrate a real-time source reconstruction pipeline that runs on a Jetson Thor at the native framerate of the Kernel Flow system (4.75 Hz) and visualizes the 3D data using volume rendering techniques.
+
+## Requirements
+
+This application was developed to run on an NVIDIA Jetson Thor Developer kit. Any Holoscan SDK supported platform should work. 
+
+To run the application you need a streaming Kernel Flow data source. This can be either:
+    - Kernel Flow hardware and SDK
+    - Downloaded `.snirf` files for use with the included data replayer. Example data can be found on [OpenNeuro](https://openneuro.org/datasets/ds006545) and copied locally to be run through the replayer.
```
tbirdso: Nitpick: not rendering as bullet points in the GitHub markdown preview, maybe a spacing issue?
```suggestion
To run the application you need a streaming Kernel Flow data source. This can be either:
- Kernel Flow hardware and SDK
- Downloaded `.snirf` files for use with the included data replayer. Example data can be found on [OpenNeuro](https://openneuro.org/datasets/ds006545) and copied locally to be run through the replayer. 
```

applications/bci_visualization/README.md
```diff
@@ -0,0 +1,231 @@
...
+    
+   ```bash
+   wget -0 data/examples/data.snirf "https://s3.amazonaws.com/openneuro.org/ds006545/sub-bed8fefe/ses-1/nirs/sub-bed8fefe_ses-1_task-audio_nirs.snirf?versionId=sYFJNjlNNlf8xVOMsIde5hpWZE2clsiu"
+   ```
+
+## Quick Start
+
+### 1. Download Required Data
+
+Download the example dataset from [Google Drive](https://drive.google.com/drive/folders/1RpQ6UzjIZAr90FdW9VIbtTFYR6-up7w2) and extract it to `data/bci_visualization` in your holohub directory. The dataset includes:
```
tbirdso: Could we update to automatically download and cache sample data as part of the build step? See [Endoscopy Tool Tracking CMakeLists.txt](https://github.com/nvidia-holoscan/holohub/blob/main/applications/endoscopy_tool_tracking/CMakeLists.txt#L25) for an example

Can keep as-is for now and automate to support testing in a subsequent testing update

applications/bci_visualization/README.md
```diff
@@ -0,0 +1,231 @@
...
+- **`blendingProfile`**: How overlapping components combine
+  
+### Visualization Strategy
+
+The three-component approach creates a layered visualization:
+
+1. **Base layer** (white, 50% opacity): Shows overall brain structure (gray + white matter) throughout the full range [0, 1]
+2. **Blue overlay** (100% opacity): Highlights low values [0, 0.4] representing decreased hemoglobin.
+3. **Red overlay** (100% opacity): Highlights high values [0.6, 1] representing increased hemoglobin.
+4. **Neutral range** [0.4, 0.6]: Only shows the white base layer (no significant change)
```
tbirdso, line 234: In addition to inlined hyperlinks, could we please add a References section with papers and reference links for further reading?

Should we cite the related NeurIPS 2025 demo?

tbirdso (commented):
Reviewed updated README

bhashemian:
Hi @techops-kernel @gabelerner-kernel,
The dataset for this BCI Visualization application is currently hosted on Google Drive, which isn’t ideal for our use case. Google Drive has limited support for handling large datasets — the API is incompatible, and downloading via the web interface splits the data into multiple zip files that need to be manually reassembled.

It would be great if we could host the dataset elsewhere. I suggest using a platform like Hugging Face, but any reliable host with an accessible API for large data downloads would work well. Thanks

applications/bci_visualization/README.md
mimiliaogo: Will add automatically download and testing in the following PR.

mimiliaogo:
@tbirdso @bhashemian 
We should've addressed most of the reviews and confirmed tested on my x86 machine.
The data is already uploaded to Huggingface.
Will add testing in a following PR. Let me know if there's any questions before merging.  
@bhashemian Could you help look at the linting issue? I run `./holohub lint` locally and it shows no errors. Not sure why it keeps failing remotely.

applications/bci_visualization/operators/color_buffer_passthrough/cpp/color_buffer_passthrough.cpp
```diff
@@ -0,0 +1,44 @@
...
+  spec.output<holoscan::gxf::Entity>("color_buffer_out");
+}
+
+void ColorBufferPassthroughOp::compute(InputContext& input, OutputContext& output,
+                                       ExecutionContext& context) {
+  auto color_message = input.receive<holoscan::gxf::Entity>("color_buffer_in");
+  if (!color_message) { throw std::runtime_error("Failed to receive color buffer message"); }
+  
+  auto cuda_streams = input.receive_cuda_streams("color_buffer_in");
```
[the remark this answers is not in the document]
tbirdso: Not a blocker

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
```diff
@@ -0,0 +1,302 @@
...
+    - permute_axis: np.ndarray shape (3,) derived from affine
+    - flip_axes: np.ndarray shape (3,) derived from affine
+    """
+
+    def __init__(self, fragment, *args, **kwargs):
+        self.selected_channel = kwargs.pop("selected_channel", 0)  # Default showing HbO channel
+        self.mask_nifti_path = kwargs.pop("mask_nifti_path", None)  # Anatomy mask NIfTI file
+
+        # Global normalization range
+        # TODO: Based on the data range
```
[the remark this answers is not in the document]
tbirdso: OK for initial contribution

applications/bci_visualization/utils/reconstruction/assets.py
tbirdso: Not a blocker

applications/bci_visualization/streams/snirf.py
tbirdso: Deferring for followup PR

applications/bci_visualization/processing/reconstruction/data_utils.py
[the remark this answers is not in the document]
tbirdso: Discussed offline, leaving as low-priority

applications/bci_visualization/operators/reconstruction/build_rhs_operator.py
```diff
@@ -0,0 +1,195 @@
...
+        # GPU caches (CuPy arrays on the propagated CUDA stream)
+        self._mega_jacobians_gpu = None
+        self._model_optical_properties_gpu = None
+        self._idxs_significant_voxels_gpu = None
+        self._jacobian_cache = None
+        self._last_frame = None  # previous frame (GPU)
+
+    def setup(self, spec: OperatorSpec) -> None:
+        spec.input("moments")
+        spec.output("batch")
```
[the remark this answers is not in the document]
tbirdso: Deferred for followup.

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
```diff
@@ -0,0 +1,341 @@
...
+        op_output.emit({"volume": self.mask_volume_gpu}, "mask_volume")
+        op_output.emit(self.out_spacing, "mask_spacing", "std::array<float, 3>")
+        op_output.emit(self.permute_axis, "mask_permute_axis", "std::array<uint32_t, 3>")
+        op_output.emit(self.flip_axes, "mask_flip_axes", "std::array<bool, 3>")
+
+        # Emit density outputs
+        op_output.emit({"volume": volume_gpu}, "volume")
+        op_output.emit(self.out_spacing, "spacing", "std::array<float, 3>")
+        op_output.emit(self.permute_axis, "permute_axis", "std::array<uint32_t, 3>")
+        op_output.emit(self.flip_axes, "flip_axes", "std::array<bool, 3>")
```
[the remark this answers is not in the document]
tbirdso: Unit tests deferred.

applications/bci_visualization/operators/color_buffer_passthrough/cpp/color_buffer_passthrough.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+  spec.output<holoscan::gxf::Entity>("color_buffer_out");
+}
+
+void ColorBufferPassthroughOp::compute(InputContext& input, OutputContext& output,
+                                       ExecutionContext& context) {
+  auto color_message = input.receive<holoscan::gxf::Entity>("color_buffer_in");
+  if (!color_message) { throw std::runtime_error("Failed to receive color buffer message"); }
+
+  auto cuda_streams = input.receive_cuda_streams("color_buffer_in");
```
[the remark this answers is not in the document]
bhashemian: @tbirdso what do you think about this?

applications/bci_visualization/operators/color_buffer_passthrough/cpp/color_buffer_passthrough.cpp
tbirdso: Looks like this is done 👍

applications/bci_visualization/utils/reconstruction/reg_inv.py
```diff
@@ -0,0 +1,124 @@
...
+SPDX-License-Identifier: Apache-2.0
+"""
+
+import logging
+import cupy as cp
+
+logger = logging.getLogger(__name__)
+
+# Cache holds CuPy arrays when running GPU-only. (Keyed by wavelength index.)
+_HESSIAN_CACHE: dict[int, object] = {}
```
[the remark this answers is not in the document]
tbirdso: @techops-kernel FYI, this comment on global Hession thread-safe access would be good address in a followup. Not a blocker.

applications/bci_visualization/operators/color_buffer_passthrough/cpp/color_buffer_passthrough.cpp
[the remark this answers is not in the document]
tbirdso: I see this CUDA synchronization warning in several review comments, consider for a followup PR

applications/bci_visualization/utils/reconstruction/reg_inv.py
```diff
@@ -0,0 +1,125 @@
...
+    -------
+    NDArray[np.float32]
+        Solution to the linear system Ax = b.
+    """
+
+    # Validate input
+    assert (A.ndim == 2) and (A.shape[0] == A.shape[1])
+    assert b.ndim in {1, 2} and b.shape[0] == A.shape[0]
+    assert cp.all(cp.isfinite(A))
+    assert cp.all(cp.isfinite(b))
```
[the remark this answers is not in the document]
bhashemian: totally agree! never use assert in the main code (only in the tests). committing the change.

applications/bci_visualization/operators/reconstruction/normalize_operator.py
```diff
@@ -0,0 +1,163 @@
...
+
+        assert not cp.any(cp.isnan(row_normalizers))
+        self._hard_coded_row_normalizers_cache = row_normalizers
+        return row_normalizers
+
+    def _normalize_batch(self, batch: BuildRHSOutput) -> Tuple[List[WavelengthSystem], int] | None:
+        num_cols = batch.data_jacobians.shape[-1]
+        num_significant = int(batch.idxs_significant_voxels.size)
+        num_absorbers, remainder = divmod(num_cols, num_significant)
+        assert not remainder
```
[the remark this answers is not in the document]
tbirdso: Deferred, we can keep `assert`s in this contribution for now

applications/bci_visualization/operators/color_buffer_passthrough/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+install(
+  TARGETS color_buffer_passthrough
+  EXPORT holoscan-ops
+)
+
+if(HOLOHUB_BUILD_PYTHON)
+  add_subdirectory(python)
+endif()
```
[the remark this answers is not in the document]
tbirdso: No needed, this operator is tightly coupled with the application

applications/bci_visualization/bci_visualization.py
```diff
@@ -0,0 +1,284 @@
...
+            config_file=self._rendering_config,
+            allocator=volume_allocator,
+            cuda_stream_pool=cuda_stream_pool,
+            **volume_renderer_kwargs,
+        )
+
+        # IMPORTANT changes to avoid deadlocks of volume_renderer and holoviz
+        # when running in multi-threading mode
+        # 1. Set the output port condition to NONE to remove backpressure
+        volume_renderer.spec.outputs["color_buffer_out"].condition(ConditionType.NONE)
```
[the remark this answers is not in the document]
tbirdso: This feedback seems incorrect, the queue size is fixed and queue policy is set correctly

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
```diff
@@ -0,0 +1,361 @@
...
+
+            volume_gpu = cp.transpose(volume_gpu, (2, 1, 0))
+            volume_gpu = cp.ascontiguousarray(volume_gpu, dtype=cp.float32)
+
+        # If we have a mask, emit oriented mask every frame for the renderer
+        if self.mask_volume_gpu is None:
+            with cp.cuda.ExternalStream(cuda_stream):
+                self.mask_volume_gpu = cp.asarray(self.mask_voxel_raw, dtype=cp.uint8)
+                self.mask_volume_gpu = cp.transpose(self.mask_volume_gpu, (2, 1, 0))
+                self.mask_volume_gpu = cp.ascontiguousarray(self.mask_volume_gpu)
```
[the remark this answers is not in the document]
tbirdso: Consider for followup review

applications/bci_visualization/utils/reconstruction/assets.py
tbirdso: Not a blocker

applications/bci_visualization/README.md
tbirdso: Not a blocker

applications/bci_visualization/utils/reconstruction/hbo.py
```diff
@@ -0,0 +1,138 @@
...
+
+        idxs_significant_voxels = cp.asarray(idxs_significant_voxels)
+        sample_mua = cp.zeros((len(wavelengths), num_voxels))
+        sample_mua[:, idxs_significant_voxels] = data_mua[idxs_significant_voxels, :].T
+        sample_hb = cp.linalg.solve(self._cached_coefficients, sample_mua)
+
+        assert sample_hb.shape == (len(wavelengths), num_voxels)
+        data_hbo = sample_hb[0]
+        data_hbr = sample_hb[1]
+        return data_hbo, data_hbr
```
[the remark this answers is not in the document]
tbirdso: FYI @techops-kernel for review. Not a blocker for initial contribution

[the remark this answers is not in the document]
tbirdso: This seems wrong, runtime has been validated

applications/bci_visualization/operators/voxel_stream_to_volume/voxel_stream_to_volume.py
```diff
@@ -0,0 +1,361 @@
...
+        self.mask_shape = None
+
+        # Running statistics for adaptive normalization (initialized from first frame)
+        self.global_min = None
+        self.global_max = None
+        self.frame_count = 0
+
+    def start(self):
+        if not self.mask_nifti_path:
+            raise ValueError("VoxelStreamToVolume: No mask NIfTI path provided")
```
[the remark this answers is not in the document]
tbirdso: @mimiliaogo seems worthwhile for followup to ensure `mask_path` is never `None`

applications/bci_visualization/operators/color_buffer_passthrough/cpp/color_buffer_passthrough.cpp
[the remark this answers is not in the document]
tbirdso: @mimiliaogo @techops-kernel have validated the application runtime under current behavior. Let's move forward for now but plan to address this in a followup PR enhancement.

tbirdso (approved):

applications/bci_visualization/operators/reconstruction/solver_operator.py
```diff
@@ -0,0 +1,92 @@
...
+        num_cols_expected = batch.num_absorbers * num_significant_voxels
+
+        # GPU-only: always use CuPy.
+        result = cp.zeros(
+            (num_cols_expected, num_wavelengths),
+            dtype=batch.systems[0].jacobian.dtype,
+        )
+        for wavelength_idx, system in enumerate(batch.systems):
+            assert system.rhs.ndim == 1
+            assert system.jacobian.shape[1] == num_cols_expected
```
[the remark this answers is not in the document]
bhashemian: makes sense! committing the suggestion.

applications/bci_visualization/utils/reconstruction/reg_inv.py
```diff
@@ -0,0 +1,129 @@
...
+        raise ValueError("Matrix A contains non-finite values")
+    if not cp.all(cp.isfinite(b)):
+        raise ValueError("Vector b contains non-finite values")
+
+    # Ensure symmetry for numerical stability
+    A = 0.5 * (A + A.T)
+
+    # Regular inverse
+    result = cp.linalg.solve(A, b)
+    assert cp.all(cp.isfinite(result))
```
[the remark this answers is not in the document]
bhashemian: makes sense! accepting the suggestion.

applications/bci_visualization/operators/reconstruction/normalize_operator.py
```diff
@@ -0,0 +1,163 @@
...
+            return self._hard_coded_row_normalizers_cache
+
+        row_normalizers = cp.full((num_rows, num_wavelengths), cp.nan)
+        for wavelength_idx in range(num_wavelengths):
+            for idx_feature in range(num_features):
+                row_normalizers[idx_feature::num_features, wavelength_idx] = HARD_CODED_NORMALIZERS[
+                    wavelength_idx
+                ][idx_feature]
+
+        assert not cp.any(cp.isnan(row_normalizers))
```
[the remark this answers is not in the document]
bhashemian: makes sense! accepting the suggestion.

applications/bci_visualization/operators/reconstruction/build_rhs_operator.py
```diff
@@ -0,0 +1,187 @@
...
+                payload.channels.detector_module[channel_idx],
+                payload.channels.detector_number[channel_idx],
+            )
+            for channel_idx in range(len(payload.channels))
+        ]
+
+        # Validate that jacobian features dimension matches realtime moments
+        # 5D jacobian shape: (channels, features, wavelengths, voxels, simulation_types)
+        num_features = realtime_moments.shape[0]
+        assert self._mega_jacobians_cpu.shape[1] == num_features
```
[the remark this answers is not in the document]
bhashemian: worth checking the logic in the followup PR.

applications/bci_visualization/operators/reconstruction/build_rhs_operator.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # Create the CUDA stream at the earliest GPU-producing operator and propagate it downstream.
+        cuda_stream = context.allocate_cuda_stream("reconstruction_stream")
+        with cp.cuda.ExternalStream(cuda_stream):
+            # Host->device copy is enqueued on the current stream (may be sync if host memory isn't pinned).
+            realtime_moments = cp.asarray(payload.data, dtype=cp.float32)
+
+            # take log of moment 0 to convert to optical density
+            # shape is (moments, channels, wavelengths)
+            cp.log(realtime_moments[0], out=realtime_moments[0])
```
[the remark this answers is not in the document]
bhashemian: @mimiliaogo it is worth validate the inputs to log function in the followup PR. Thanks

bhashemian (approved):
LGTM! Thank you so much for contributing this intriguing application.@mimiliaogo @gabelerner-kernel @techops-kernel

Also thank you for uploading the data to Hugging Face. It would be fantastic if we could update the folder structure on HF to align with the expected folder structure in `./holohub/data/bci_visualization`. This way, users can simply download the data without having to reorganize it.
```
holohub/data/bci_visualization/
├── anatomy_labels_high_res.nii.gz      # Brain segmentation
├── data.snirf                          # SNIRF format brain activity data
├── extinction_coefficients_mua.csv     # Absorption coefficients for HbO/HbR
├── flow_channel_map.json               # Sensor-source channel mapping
├── flow_mega_jacobian.npy              # Pre-computed sensitivity matrix (channels → voxels)
└── voxel_info/                         # Voxel geometry and optical properties
    ├── affine.npy                      # 4x4 affine transformation matrix
    ├── idxs_significant_voxels.npy     # Indices of voxels with sufficient sensitivity
    ├── ijk.npy                         # Voxel coordinates in volume space
    ├── mua.npy                         # Absorption coefficient per voxel
    ├── musp.npy                        # Reduced scattering coefficient per voxel
    ├── resolution.npy                  # Voxel resolution (mm)
    ├── wavelengths.npy                 # Measurement wavelengths (690nm, 905nm)
    └── xyz.npy                         # Voxel coordinates in anatomical space (mm)
```

Finally, as you know, we’ve postponed some improvements and testing to a future PR to be able to merge this PR on time for the demo. That follow-up PR is crucial for maintaining code quality, so we’re looking forward to it. Thanks

bhashemian:
@coderabbitai resolve all comments

applications/bci_visualization/operators/color_buffer_passthrough/cpp/color_buffer_passthrough.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+  output.emit(color_message.value(), "color_buffer_out");
```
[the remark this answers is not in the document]
mimiliaogo: This should be `receive_cuda_stream`, will fix it in new PR.

PR #1340, opened 2026-01-13
benchmarks: add build options

kimullaa:
> All dependencies are automatically managed by the Holohub Docker container. You can simply run:
> ./holohub run-container [<application_name>]
> 
https://github.com/nvidia-holoscan/holohub/tree/main/benchmarks/holoscan_flow_benchmarking#holohub-docker-container-recommended

I ran run-container after reading the documentation.

Running benchmark in the container without any options results in errors due to missing packages.

```
$ ./holohub run-container  endoscopy_tool_tracking
...
xxx@yyy:~$ python3 benchmarks/holoscan_flow_benchmarking/benchmark.py -a  endoscopy_tool_tracking  -r 3 -i 3 -m 200     --sched greedy     -d myoutputs
  File "/workspace/holohub/benchmarks/holoscan_flow_benchmarking/benchmark.py", line 27, in <module>
    from nvitop import Device
ModuleNotFoundError: No module named 'nvitop'
```

## Summary by CodeRabbit

* **Documentation**
  * Updated benchmarking guide: new container run command includes build args to enable the benchmarking-setup target by default.
  * Added a note clarifying the container will be built and run with benchmarking tools unless an application provides its own Dockerfile; custom Dockerfiles may omit the benchmarking target—see the application README for details.

### DIFF
--- a/applications/vila_live/README.md
+++ b/applications/vila_live/README.md
@@ -83,6 +83,6 @@ This application downloads a pre-recorded video from [Pexels](https://www.pexels
     ffmpeg -stream_loop -1 -re -i <your_video_path> -pix_fmt yuyv422 -f v4l2 /dev/video3
     ```
 ## 🙌 Acknowledgements
-- Jetson AI Lab, [Live LLaVA](https://www.jetson-ai-lab.com/tutorial_live-llava.html): for the inspiration to create this app
+- Jetson AI Lab, [Live LLaVA](https://www.jetson-ai-lab.com/archive/tutorial_live-llava.html): for the inspiration to create this app
 - [Jetson-Containers](https://github.com/dusty-nv/jetson-containers/tree/master/packages/llm/llamaspeak) repo: For the Flask web-app with WebSockets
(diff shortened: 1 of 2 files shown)


benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -33,7 +33,7 @@ For detailed information, refer to:
 All dependencies are automatically managed by the Holohub Docker container. You can simply run:
 
 ```bash
-./holohub run-container [<application_name>]
+./holohub run-container --build-args "--target benchmarking-setup" [<application_name>]
```
bhashemian, line 36: @kimullaa thanks for the update. However, this only works for applications that does not define their own custom Dockerfile and relies on the default Dockerfile, so the actual command would be:
```suggestion
./holohub run-container --build-args "--target benchmarking-setup"
```

Also it's worth adding a note that if an application defines its own Dockerfile, the user should use:
```
./holohub run-container <application_name>
```

Thanks

bhashemian: I added a note! Thanks

PR #1347, opened 2026-01-14
App: Add find_package(holoscan) to CMakeLists.txt of Python-only apps

jcfr:
## Context

When building HoloHub with a specific application enabled (e.g., `-DAPP_colonoscopy_segmentation=ON`) and an explicit `CMAKE_CUDA_COMPILER` specified (e.g., `-D CMAKE_CUDA_COMPILER=/usr/local/cuda/bin/nvcc`), CMake reports a warning:

```
CMake Warning:
  Manually-specified variables were not used by the project:
    CMAKE_CUDA_COMPILER
```

This occurs because Python-only apps using `project(... NONE)` don't enable any languages, so the compiler variable is never consumed.

## Root Cause

The holoscan package uses `rapids_export()` with `LANGUAGES C CXX CUDA`, which enables these languages when `find_package(holoscan)` is called. This consumes the `CMAKE_CUDA_COMPILER` variable.

Most Python-only apps (e.g., `body_pose_estimation`) already call `find_package(holoscan ...)` and don't trigger the warning. A few apps were missing this call.

## Description

Add `find_package(holoscan ...)` to 5 Python-only apps that were missing it:
- `colonoscopy_segmentation`
- `holochat`
- `isaac_sim_holoscan_bridge`
- `polyp_detection`
- `simple_radar_pipeline` (Python build path only)

This ensures consistency with other HoloHub applications and provides access to holoscan CMake utilities.

## Summary by CodeRabbit

* **Chores**
  * Updated build configurations across multiple applications to ensure required runtime components are discovered and to reduce build-system warnings, improving reliability for Python-only and mixed builds.
  * Updated license header years in a few application files.

### DIFF
--- a/applications/colonoscopy_segmentation/CMakeLists.txt
+++ b/applications/colonoscopy_segmentation/CMakeLists.txt
@@ -15,6 +15,11 @@
 
 project(ColonoscopySegmentation NONE)
 
+# Find holoscan to ensure CMAKE_CUDA_COMPILER is consumed (silences CMake warning
+# about unused variable for Python-only apps with project(NONE))
+find_package(holoscan 1.0 REQUIRED CONFIG
+             PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")
+
 # Default to download datasets
 option(HOLOHUB_DOWNLOAD_DATASETS "Download datasets" ON)
 
--- a/applications/holochat/CMakeLists.txt
+++ b/applications/holochat/CMakeLists.txt
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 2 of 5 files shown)


applications/colonoscopy_segmentation/CMakeLists.txt
```diff
@@ -15,6 +15,11 @@
 
 project(ColonoscopySegmentation NONE)
 
+# Find holoscan to ensure CMAKE_CUDA_COMPILER is consumed (silences CMake warning
+# about unused variable for Python-only apps with project(NONE))
+find_package(holoscan 1.0 REQUIRED CONFIG
+             PATHS "/opt/nvidia/holoscan" "/workspace/holoscan-sdk/install")
```
tbirdso, line 21: I don't think I agree with this approach. The `find_package(holoscan)` call is adding a dependency on the Holoscan SDK C++ development package, which is generally not needed for pure Python apps. This could block developers aiming to use these applications in a Python-only environment.

We should not add new dependencies as a workaround to silence warnings. Could we please explore why the CMAKE_CUDA_COMPILER variable is being provided instead? Maybe we could avoid providing this for pure Python apps? Otherwise, I'd suggest simply suppressing this warning.

tbirdso (changes requested):

jcfr:
Closing in favor of a simpler approach specific to the CI infrastructure.

PR #1353, opened 2026-01-16
Update Documentation for Benchmarking Script

bhashemian:
## Summary by CodeRabbit

* **Documentation**
  * Updated HoloHub Benchmarking guide with restructured content for improved clarity.
  * Revised Docker workflow instructions and expanded guidance on benchmarking environment setup.
  * Enhanced notes regarding custom Dockerfile compatibility and testing considerations.

* **Chores**
  * Removed unused dependency from Python requirements.

### DIFF
--- a/benchmarks/holoscan_flow_benchmarking/README.md
+++ b/benchmarks/holoscan_flow_benchmarking/README.md
@@ -44,8 +44,8 @@ All dependencies are automatically managed by the Holohub Docker container. You 
 ./holohub run-container --extra-script benchmarking [<application_name>]
 ```
 
-  > [!NOTE]
-  > This command sets up the benchmarking environment on top of the default Holohub container image, unless you specify an application with a custom Dockerfile. In that case, the benchmarking environment attempts to extend the application's Dockerfile. Please note that this process has not been extensively tested with all custom Dockerfiles—compatibility depends on how the Dockerfile is configured, so conflicts or issues could occur or the Dockerfile may already have the support for benchamrking and could work without `--extra-script benchmarking`. For specific instructions consult the corresponding application's README.
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -52,7 +60,6 @@ These python dependencies include:
 - numpy: Data processing
 - matplotlib: Graph generation
 - nvitop: GPU monitoring
-- argparse: CLI handling
```
tbirdso, line 55: Why remove? Redundant?

benchmarks/holoscan_flow_benchmarking/README.md
```diff
@@ -33,11 +41,11 @@ For detailed information, refer to:
...
 ```
 
 > [!NOTE]
-> By default, this command will build and run the Holohub Docker container with benchmarking tools enabled, unless the application already provides its own Dockerfile.  If the application has a custom Dockerfile, the `benchmarking-setup` target may not be available within its container image. For further details or special instructions, please consult the README of the respective application.
+> This command sets up the benchmarking environment on top of the default Holohub container image, unless you specify an application with a custom Dockerfile. In that case, the benchmarking environment attempts to extend the application's Dockerfile. However, this process hasn't been fully tested for every possible custom Dockerfile, and success depends on how the Dockerfile is authored. Some Dockerfiles may already include the required benchmarking support and may work even without the `--extra-script benchmarking` option; in other cases, there may be conflicts or additional manual steps needed. Always refer to the relevant application's README for any application-specific benchmarking instructions or compatibility considerations.
```
tbirdso, line 48: Nitpick: could simplify for brevity, your call

```suggestion
> This command sets up the benchmarking environment on top of the default Holohub container image, unless you specify an application with a custom Dockerfile. In that case, the benchmarking environment extends the application's Dockerfile. Note that this is not tested for every possible container; always refer to the relevant application's README for any application-specific benchmarking instructions or compatibility considerations.
```

bhashemian: Thanks @tbirdso. I like your suggestion, but since there aren’t usually other places for users to find information about this benchmarking, I’d prefer to be a bit more descriptive.

benchmarks/holoscan_flow_benchmarking/README.md
bhashemian: correct, it is a built-in module.

PR #1365, opened 2026-01-21
Shorten Kernel BCI App Name For Better Website Presentation

bhashemian:
Currently, the name of this application was going beyond of the expected size for each model card. This PR fix that.

### DIFF
--- a/applications/bci_visualization/metadata.json
+++ b/applications/bci_visualization/metadata.json
@@ -1,6 +1,6 @@
 {
 	"application": {
-		"name": "Kernel Flow BCI Real-Time Reconstruction and Visualization",
+		"name": "Kernel Flow BCI Real-Time Visualization",
 		"authors": [
 			{
 				"name": "Holoscan Team",


tbirdso:
Could we improve text wrapping for website title cards and/or enforce a maximum title length in CI/CD to prevent similar situations in the future?

I also see that the preview card for the recent `ultrasound_postprocessing` contribution displays the preview image in the wrong aspect ratio, could you please help investigate?

https://nvidia-holoscan.github.io/holohub/

image

bhashemian:
> I also see that the preview card for the recent `ultrasound_postprocessing` contribution displays the preview image in the wrong aspect ratio, could you please help investigate?

@tbirdso, the displayed image is consistently square. Do you believe this is an issue? If so, it’s a separate issue unrelated to this PR. Please create a ticket for that. Thanks

Regarding text wrapping, I agree with you. However, @jjomier had a different perspective on this when I was fixing another lengthy name (Live Streaming Data Web Dashboard with NATS). @jjomier could you please share your thoughts on this?

PR #1377, opened 2026-01-27
[Flow Benchmarking] fix cases where there are not enough samples

sohamm17:
MR fixes the case when there are not enough samples (min: 20) in the log file for a particular path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Graceful handling when groups or paths lack samples: metrics show "Not enough samples" and empty groups/paths are skipped in outputs and plots.
  * Per-group aggregation emits a consolidated value only if at least one path has data; avoids empty/invalid CSV/CDash entries.
  * CDF plotting skips empty paths to prevent errors.

* **Chores**
  * Metric printing improved to optionally omit unit suffix for "no data" states, clarifying output formatting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,10 @@ ngc-cli/
 CTestConfig.cmake
 .ngc
 
+# AI coding assistants
+.cursor/
+CLAUDE.md
+
 # Cache and temporary directories
 .cache/
 .cargo/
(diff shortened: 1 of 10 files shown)


bhashemian:
@sohamm17 thanks for this great improvement. Is there any specific use case that could benefit from this, so we can test it with?

sohamm17:
> @sohamm17 thanks for this great improvement. Is there any specific use case that could benefit from this, so we can test it with?

It's mostly adding empty check guards so that we don't throw runtime error when data flow tracking logs don't have enough sample entries. Recently, we have seen such log files, but no specific case to test it with.

vani-nag:
Thanks for the changes @sohamm17! LGTM! Had a minor question about adding `unit` to `print_path_metric_ms` signature

PR #1385, opened 2026-01-30
Fix VSCode Debugging Issues

bhashemian:
## Summary by CodeRabbit

* **Chores**
  * Introduced explicit language-specific build variants (Python / C++) with clearer labels, added matching delay/compound variants, and improved task grouping for multi-language workflows.

* **Bug Fixes**
  * Fixed a minor configuration syntax error and updated several working-directory settings to align with renamed build outputs and ensure consistent launch/task naming.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -189,7 +189,7 @@
                 "debug",
                 "--local",
                 "--language",
-                "cpp"
+                "python"
             ],
             "options": {
                 "cwd": "${env:WORKSPACE_DIR}"
@@ -269,7 +269,7 @@
                 "debug",
                 "--local",
                 "--language",
-                "cpp"
+                "python"
             ],
             "options": {
                 "cwd": "${env:WORKSPACE_DIR}"
@@ -788,9 +788,7 @@
                 "build",
                 "${input:app_name}",
                 "--benchmark",
-                "--local",
-                "--language",
-                "cpp"
+                "--local"
             ],
             "options": {
                 "cwd": "${env:WORKSPACE_DIR}"


.vscode/tasks.json
```diff
@@ -528,7 +628,9 @@
                 "nvidia_nim_nvclip",
                 "--build-type",
                 "debug",
-                "--local"
+                "--local",
+                "--language",
+                "cpp"
```
tbirdso: NVClip app is Python

```suggestion
                "python"
```

.vscode/tasks.json
```diff
@@ -131,7 +187,9 @@
                 "colonoscopy_segmentation",
                 "--build-type",
                 "debug",
-                "--local"
+                "--local",
+                "--language",
+                "cpp"
```
tbirdso: `colonoscopy_segmentation` is Python

.vscode/tasks.json
```diff
@@ -149,7 +207,9 @@
                 "cuda_quantum",
                 "--build-type",
                 "debug",
-                "--local"
+                "--local",
+                "--language",
+                "cpp"
```
tbirdso: `cuda_quantum` is Python

```suggestion
                "python"
```

.vscode/tasks.json
```diff
@@ -113,7 +167,9 @@
                 "sam2",
                 "--build-type",
                 "debug",
-                "--local"
+                "--local",
+                "--language",
+                "cpp"
```
tbirdso: `sam2` is Python
```suggestion
                "python"
```

.vscode/tasks.json
```diff
@@ -95,7 +147,9 @@
                 "asr_to_llm",
                 "--build-type",
                 "debug",
-                "--local"
+                "--local",
+                "--language",
+                "cpp"
```
tbirdso: `asr_to_llm` is Python

```suggestion
                "python"
```

tbirdso (commented):
I see a couple of issues where `python` apps are explicitly marked with `--language cpp`.

Could you please help review and verify for any other language mismatches?

Alternatively, if there is only one language option for an app, should we default to omitting the `--language` parameter?

bhashemian:
> I see a couple of issues where `python` apps are explicitly marked with `--language cpp`.
> 
> Could you please help review and verify for any other language mismatches?
> 
> Alternatively, if there is only one language option for an app, should we default to omitting the `--language` parameter?

Thanks for catching these mismatches, @tbirdso! The reason behind these mismatches is that I mistakenly assumed that changing `—language cpp` would simply alter the default settings, which is obviously not! I will correct them, and review the rest.

bhashemian:
@tbirdso, I’ve reviewed all the tasks and verified if the build command can locate the application with the specified language. I kept the language parameter even if it was just one language to prevent any future confusion if we expand the supported languages for a particular application. The only place I removed the language was for benchmarking purposes. Thanks again for your review and feedback.

PR #1390, opened 2026-02-04
cdash: support non-nightly dashboard models

agirault:
https://cmake.org/cmake/help/latest/command/ctest_start.html

Want to be able to configure ctest runs to push Experimental or Continuous builds too, not just Nightly

## Summary by CodeRabbit

* **Chores**
  * Testing pipeline updated to use a configurable dashboard model with an automatic fallback to the default model, improving flexibility and ensuring tests run with a valid configuration when no explicit model is provided.

### DIFF
--- a/utilities/testing/holohub.container.ctest
+++ b/utilities/testing/holohub.container.ctest
@@ -108,7 +108,12 @@ endif()
 # Default nightly start time
 set(CTEST_NIGHTLY_START_TIME "06:00:00 UTC")
 
-ctest_start(Nightly)
+# Dashboard model: Nightly (scheduled), Experimental (manual), or Continuous (CI)
+if(NOT DASHBOARD_MODEL)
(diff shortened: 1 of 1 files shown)


utilities/testing/holohub.container.ctest
```diff
@@ -108,7 +108,12 @@ endif()
 # Default nightly start time
 set(CTEST_NIGHTLY_START_TIME "06:00:00 UTC")
 
-ctest_start(Nightly)
+# Dashboard model: Nightly (scheduled), Experimental (manual), or Continuous (CI)
```
wyli, line 111: ‘Model’ vs ‘mode’, looks like a typo here..

agirault: no, it it called model: https://cmake.org/cmake/help/latest/command/ctest_start.html

wyli: ah good to know

PR #1398, opened 2026-02-09
fix(cli): Skip sccache compiler launchers for Python apps

jcfr:
Python-only applications don't compile C/C++/CUDA code, so setting `CMAKE_*_COMPILER_LAUNCHER` variables causes CMake warnings about unused variables. Only set these for C++ builds where compilation actually occurs.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to skip compiler-launcher options for Python builds so launcher flags are only applied for non-Python targets, ensuring Python build behavior no longer uses the compiler launcher while other languages continue to do so.


utilities/cli/holohub.py
```diff
@@ -1059,13 +1059,14 @@ def build_project_locally(
...
                 )
             # Set CMake compiler launchers with -D
-            cmake_args.extend(
-                [
-                    f"-DCMAKE_C_COMPILER_LAUNCHER={sccache_bin}",
-                    f"-DCMAKE_CXX_COMPILER_LAUNCHER={sccache_bin}",
-                    f"-DCMAKE_CUDA_COMPILER_LAUNCHER={sccache_bin}",
-                ]
-            )
+            if language == "cpp":
```
bhashemian: The normalized language can be "", "cpp", and "python", so maybe it is safer to disable it only for python.

```suggestion
            if language != "python":
```

PR #1406, opened 2026-02-15
Don't use SCCACHE_MEMCACHED_ENDPOINT if it is not valid

wendell-hom:
Don't pass the sccache memcached endpoint env variable if the server is not reachable, otherwise compilation will fail.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added runtime validation for the cache endpoint so unreachable endpoints are no longer forwarded to the environment, preventing misconfiguration.
  * Emits a warning and falls back to local caching when the endpoint cannot be reached, improving reliability and startup behavior.


utilities/cli/container.py
```diff
@@ -757,7 +778,7 @@ def get_environment_args(self) -> List[str]:
             args.extend(["-e", f"SCCACHE_DIR={SCCACHE_CONTAINER_DIR}"])
             # Forward other SCCACHE_* environment variables present on host
             for k in sccache_keys:
-                if k != "SCCACHE_DIR":
+                if k != "SCCACHE_MEMCACHED_ENDPOINT" or self.is_valid_endpoint():
```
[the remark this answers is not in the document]
wendell-hom: take a look again

wyli (approved):

utilities/cli/container.py
```diff
@@ -757,7 +779,7 @@ def get_environment_args(self) -> List[str]:
             args.extend(["-e", f"SCCACHE_DIR={SCCACHE_CONTAINER_DIR}"])
             # Forward other SCCACHE_* environment variables present on host
             for k in sccache_keys:
-                if k != "SCCACHE_DIR":
+                if (k != "SCCACHE_DIR") and (k != "SCCACHE_MEMCACHED_ENDPOINT" or self.is_valid_endpoint()):
```
wyli: looks good to me, minor suggestions:
- add the env var to the printing util https://github.com/nvidia-holoscan/holohub/blob/2ac7c840a9b97ee31b478df2d53064be47b06971/utilities/cli/util.py#L1308
- add usage to readme: https://github.com/nvidia-holoscan/holohub/blob/2ac7c840a9b97ee31b478df2d53064be47b06971/utilities/cli/README.md?plain=1#L445

bhashemian:
@wendell-hom what do you think about Wenqi's comment? Could you please respond? Thanks
https://github.com/nvidia-holoscan/holohub/pull/1406#discussion_r2811847495

bhashemian:
@wendell-hom Could you please address the comments on this PR? Thanks.

bhashemian:
Hi @wendell-hom, this behavior may still be useful, but it now belongs in `nvidia-holoscan/holoscan-cli`. Would be okay if we close it here?

bhashemian (dismissed):

utilities/cli/container.py
```diff
@@ -762,7 +784,9 @@ def get_environment_args(self) -> List[str]:
             args.extend(["-e", f"SCCACHE_DIR={SCCACHE_CONTAINER_DIR}"])
             # Forward other SCCACHE_* environment variables present on host
             for k in sccache_keys:
```
bhashemian, line 786: Blocking: This predicate calls `is_valid_endpoint()`, which performs network I/O while Docker args are being assembled. Could the dry-run state be passed into endpoint validation so `--dry-run` only checks the endpoint format and never opens a socket?

bhashemian (changes requested):
Thank you for the sccache fallback fix.

Blocking:
- `utilities/cli/container.py` performs a real TCP probe while assembling Docker args, including dry-run paths. `--dry-run` should not perform network I/O or wait up to 5 seconds; please thread the dry-run state into endpoint validation and validate only host/port format when dry-run is active.
- The branch currently has merge conflicts; these need to be resolved before merge.

wendell-hom:
> Hi @wendell-hom, this behavior may still be useful, but it now belongs in `nvidia-holoscan/holoscan-cli`. Would be okay if we close it here?

sure, closing, thanks!

PR #1520, opened 2026-04-20
feat: Add Atracsys Visualizer for spryTrack 300 RGB (Visible, IR, Structured Light) and Marker Tracking

artrittelaku-wayland:
## Description
This PR introduces the `atracsys_visualizer`, a HoloHub application designed to showcase the imaging and submillimetric tracking capabilities of the Atracsys **spryTrack 300 RGB** camera. 

To ensure continuous CI reliability without hardware dependencies, the application defaults to a `replayer` mode using a high-fidelity dataset hosted on the Wayland Technologies GitHub Releases page.

### Features Highlighted:
- **Visible Mode**: Raw RGB stream output.
- **Infrared (IR) Mode**: Native hardware infrared capabilities, which process simultaneously with marker tracking.
- **Structured-Light Mode**: Disparity map generation using structured light points.
- **High-Accuracy Tracking**: Real-time rendering of instrument poses with submillimetric accuracy. Includes `geometry10.ini` as a custom reflective fiducial reference.

### Notes for Reviewers:
- **Default Mode (`replayer`)**: The dataset (.zip) is downloaded dynamically in CMake (`file(DOWNLOAD...)`) natively for CI tests and quick-starts.
- **Live Camera Support (`live_camera`)**: Fully configured operator, but requires proprietary SDKs (spryTrack SDK v4.9.0 and S3DK) provided by `support@atracsys.com`.
- **Dependencies**: Explicitly requires `--cuda 12` specified in the Docker build due to underlying OpenCV `4.10.0` behavior with CUDA 13 headers.

## Checklist
- [x] I have run `./holohub lint` locally
- [x] I have tested the application successfully locally in `replayer` mode
- [x] Required copyright headers (`SPDX-FileCopyrightText: Copyright (c) 2026 Wayland Technologies.`) included
- [x] README is fully populated with setup, SDK, and controls instructions
- [x] `metadata.json` conforms to HoloHub schemas

## Summary by CodeRabbit

* **New Features**
  * Atracsys Visualizer app: replay-first viewer with optional live-camera mode, GPU-accelerated processing, and Docker-ready runtime.
  * Interactive mode switching (Visible / Infrared / Structured-Light / Tracking) with keyboard controls, overlays, marker projection, and point-cloud rendering.
  * Automatic demo dataset fetch, configurable recordings path, and runtime calibration validation; headless test run for replay mode.

* **Documentation**
  * Detailed README covering setup, proprietary SDK notes, calibration/registration steps, dataset/recording guidance, build/run commands, and keyboard controls.


bhashemian:
Hi @artrittelaku-wayland, thanks for your contribution. Could you please provide me write access to this PR's branch of your fork, so we can help updating and manage the PR merge? Thanks

operators/atracsys_camera/master_source_op.cpp
```diff
@@ -0,0 +1,812 @@
...
+    is_running_ = true;
+    async_cond_->event_state(
+        holoscan::AsynchronousEventState::EVENT_WAITING);
+    capture_thread_ =
+        std::thread(&AtracsysMasterSourceOp::capture_loop, this);
+  } catch (...) {
+    destroy_frame();
+    AtracsysDevice::instance().shutdown();
+    throw;
+  }
```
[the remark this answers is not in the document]
artrittelaku-wayland: I don’t think the specific `s3dk_` restart concern applies here. `RealS3DKWrapper` is a stateless adapter, and `start()` calls `initializeDeviceHelper()` on every retry, so the wrapper itself is not left in a partially initialized state. The catch block already destroys `frame_`; the only narrower cleanup gap would be the raw S3DK pointers (`stereo_params_`, `engine_`, `gpu_frame_`) if an exception occurs after they are created.

operators/atracsys_camera/master_source_op.cpp
```diff
@@ -0,0 +1,811 @@
...
+
+    is_running_ = true;
+    async_cond_->event_state(
+        holoscan::AsynchronousEventState::EVENT_WAITING);
+    capture_thread_ =
+        std::thread(&AtracsysMasterSourceOp::capture_loop, this);
+  } catch (...) {
+    AtracsysDevice::instance().shutdown();
+    throw;
+  }
```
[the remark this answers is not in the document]
artrittelaku-wayland: Addressed. `start()` now calls `destroy_frame()` in the exception path before shutting down the device, and `destroy_frame()` destroys the frame and nulls `frame_`.

operators/atracsys_camera/sdk/sdk_wrapper.hpp
```diff
@@ -0,0 +1,89 @@
...
+    ftkLibrary lib = ftkInitExt(config, buffer);
+    if (!lib) {
+      if (buffer && buffer->data) {
+        throw std::runtime_error(std::string("ftkInitExt failed: ") +
+                                 reinterpret_cast<const char*>(buffer->data));
+      }
+      throw std::runtime_error("ftkInitExt failed (unknown error)");
+    }
+    return ftkError::FTK_OK;
+  }
```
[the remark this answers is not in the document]
artrittelaku-wayland: Partially addressed.

The unbounded-read issue is fixed: `initExt()` now builds the error message using `buffer->size` rather than assuming NUL termination.

operators/atracsys_camera/metadata.json
```diff
@@ -0,0 +1,50 @@
+{
+  "operator": {
+    "name": "AtracsysCameraOp",
```
bhashemian: Is this the correct name for this operator? The actual operators that this metadata represets seems to be `AtracsysMasterSourceOp` and `PointCloudFilterOp`, right?
I suggest the following options:
1. Either change the name to "Atracsys Camera Operators" (note the "s" at the end)
2. OR separate each operator in its own subdirectory.

Thanks

applications/atracsys_visualizer/CMakeLists.txt
```diff
@@ -0,0 +1,86 @@
...
+    COMMAND ${CMAKE_COMMAND} -E tar xf "${CMAKE_CURRENT_BINARY_DIR}/atracsys_visualizer.zip"
+    WORKING_DIRECTORY "${HOLOHUB_DATA_DIR}"
+    RESULT_VARIABLE EXTRACT_STATUS
+  )
+  if(NOT EXTRACT_STATUS EQUAL 0)
+    message(FATAL_ERROR "Failed to extract dataset.")
+  endif()
+  file(REMOVE "${CMAKE_CURRENT_BINARY_DIR}/atracsys_visualizer.zip")
+  endif()
+endif()
```
bhashemian, line 60: Can this download/extraction logic be replaced by `holoscan_download_data`

Like these examples:

https://github.com/nvidia-holoscan/holohub/blob/09cfd82b5bf7471bac27403429865393fd0fb21f/applications/imaging_ai_segmentator/CMakeLists.txt#L53-L58

https://github.com/nvidia-holoscan/holohub/blob/09cfd82b5bf7471bac27403429865393fd0fb21f/workflows/ai_surgical_video/CMakeLists.txt#L38-L69

applications/atracsys_visualizer/cpp/atracsys_visualizer.cpp
```diff
@@ -0,0 +1,290 @@
...
+  }
+  for (size_t i = 0; i < distortion.size(); ++i) {
+    calibration->distortion[i] = static_cast<float>(distortion[i]);
+  }
+
+  if (!calibration->valid()) {
+    throw std::runtime_error("Atracsys visualizer calibration is invalid");
+  }
+  return calibration;
+}
```
bhashemian, line 124: Could we use Holoscan's `from_config` like every other parameter, removing the explicit yaml-cpp::yaml-cpp link here? Thanks

applications/atracsys_visualizer/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+The application is split into:
+
+- a required reusable `AtracsysModeSwitcherOp` operator
+- an optional `atracsys_camera` operator package for live hardware input
+- a single C++ application entrypoint with configuration via `atracsys_visualizer.yaml`
+
+## Visual Showcase
+
+![Atracsys Visualizer Demo](demo.gif)
```
bhashemian: Can we move this upper, like right before `## Architecture`?

applications/atracsys_visualizer/demo.gif
bhashemian, line 1: Is it possible to reduce the size of tissue gif? Thanks

bhashemian (commented):
Thanks @artrittelaku-wayland for this interesting contribution. I reviewed and added a few comments inline.

bhashemian:
@artrittelaku-wayland, considering that building OpenCV 4.10 from source incurs significant container build time, it would be nice to explicitly mention this as a known cost in the README, thereby setting the user’s expectations. Thanks

operators/atracsys_camera/metadata.json
artrittelaku-wayland: I updated the metadata name to Atracsys Camera Operators since this package contains both AtracsysMasterSourceOp and PointCloudFilterOp.

applications/atracsys_visualizer/CMakeLists.txt
artrittelaku-wayland: I replaced the custom download/extraction logic with holoscan_download_data and verified the replay dataset downloads correctly and ends up under the expected data/atracsys_visualizer layout.

applications/atracsys_visualizer/cpp/atracsys_visualizer.cpp
artrittelaku-wayland: I switched the calibration loading over to Holoscan config access instead of YAML::LoadFile and removed the explicit yaml-cpp link. I also verified the replay app still builds and runs.

applications/atracsys_visualizer/README.md
artrittelaku-wayland: I moved the visual showcase section up so it appears right before ## Architecture.

applications/atracsys_visualizer/demo.gif
artrittelaku-wayland: I replaced the demo GIF with a smaller version. The new file is half size of the previous one.

applications/atracsys_visualizer/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+              "-DDATASET_DIR=${HOLOHUB_DATA_DIR}/atracsys_visualizer"
+              "-DSTAMP_FILE=${HOLOHUB_DATA_DIR}/atracsys_visualizer/.layout-normalized.stamp"
+              -P "${CMAKE_CURRENT_SOURCE_DIR}/normalize_dataset_layout.cmake"
+      DEPENDS atracsys_visualizer_data
+      VERBATIM
+    )
+
+    add_custom_target(atracsys_visualizer_dataset ALL
+      DEPENDS "${HOLOHUB_DATA_DIR}/atracsys_visualizer/.layout-normalized.stamp"
+    )
```
bhashemian: May I know why do we still need this? `holoscan_download_data` automatically creates `atracsys_visualizer_data`, isn't it enough for your purpose?

operators/atracsys_camera/master_source_op.cpp
```diff
@@ -0,0 +1,812 @@
...
+  is_running_ = false;
+  if (capture_thread_.joinable()) {
+    capture_thread_.join();
+  }
+
+  reset_state();
+  destroy_frame();
+  gpu_frame_ = nullptr;
+  engine_ = nullptr;
+  stereo_params_ = nullptr;
```
bhashemian: **Resource leak across start/stop cycles.**

`stereo_params_`, `engine_`, and `gpu_frame_` are created at lines 231-233 via the S3DK factory functions (`createStereoParameters`, `createDefaultEngine`, `createGpu3DFrame`). They are raw owning pointers (declared as such in the header). `stop()` here only assigns `nullptr` — there is no matching `s3dk_->destroyStereoParameters(...)` / `destroyDefaultEngine(...)` / `destroyGpu3DFrame(...)` call anywhere in the file (`grep` for `destroy` returns only `destroyFrame`, which is unrelated).

Every `start()` → `stop()` cycle (e.g. mode reconfiguration, app restart) leaks the underlying GPU buffers and engine state. With structured-light buffers this is significant GPU memory.

Suggested fix: wrap each pointer in a `std::unique_ptr` with a custom deleter that calls the matching `s3dk_->destroy*` function, or call the destroy functions explicitly here before nulling. The header declarations at `master_source_op.hpp:108-110` should change in lockstep.

operators/atracsys_camera/master_source_op.cpp
```diff
@@ -0,0 +1,812 @@
...
+void AtracsysMasterSourceOp::load_geometries() {
+  const auto& dev = AtracsysDevice::instance();
+  const int ref_status = sdk_.loadBody(dev.lib(), geometry_path_.get(), geometry_);
+  if (ref_status == 0 || ref_status == 1) {
+    sdk_.setRigidBody(dev.lib(), dev.serial(), &geometry_);
+  } else {
+    HOLOSCAN_LOG_WARN("AtracsysMasterSourceOp: failed to load geometry from {}",
+                      geometry_path_.get());
+  }
+}
```
bhashemian, line 413: **Silent tracking degradation when geometry fails to load.**

If `loadBody` fails (returns anything other than 0 or 1 — e.g. file not found, corrupted file, parse error), this function logs a `WARN` and returns. `start()` then proceeds, `setRigidBody` is never called, and the operator runs forever emitting empty `marker_poses` because `frame_->markersCount` stays 0 without a registered body. The user sees no fiducials in the visualizer, no error in the logs at ERROR level, and a single WARN that is easy to miss in the firehose of `"waiting for frames"` warnings logged elsewhere in this file.

Recommendations:
1. Promote this to `HOLOSCAN_LOG_ERROR` (or `throw` from `start()`). For a tracking app this is a fatal misconfiguration, not a warning condition.
2. Surface the actual SDK error code — `geometry_helper.cpp:43-47` collapses every distinct `ftkError` (file-not-found, buffer-load-failed, rigid-body-parse-failed) to the same return value `2`, so the user can't distinguish a typo'd path from a corrupt geometry file. Either pass the original `ftkError` through, or branch on it here.

Without this, a typo in `geometry_path` deploys silently to the field and tracking fiducials never appear, with no actionable diagnostic.

operators/atracsys_camera/master_source_op.cpp
```diff
@@ -0,0 +1,812 @@
...
+      ftkError::FTK_OK) {
+    throw std::runtime_error("AtracsysMasterSourceOp: failed to set scheduler pattern");
+  }
+
+  current_pattern_ = pattern;
+}
+
+void AtracsysMasterSourceOp::set_exclusive_pattern(atracsys::HardwareMode mode) {
+  set_scheduler_pattern(std::string(54, slot_for_mode(mode)));
+}
```
bhashemian, line 771: **An SDK failure during a user-driven mode switch will tear down the whole application.**

Call chain when the user presses a key to switch mode:

```
compute()                                    // master_source_op.cpp:301
  → apply_pending_command(hw_cmd)
    → set_exclusive_pattern(mode)            // here, line 716-718
      → set_scheduler_pattern(...)           // throws std::runtime_error
                                             // ("failed to set scheduler pattern") at line 710
                                             // when sdk_.setData() returns non-OK
```

Mode switching is the central feature of this application, so this throw fires on a normal user action (press '1'/'2'/'3'/'4'). Because the throw escapes from `compute()`, Holoscan's executor terminates the operator and the surgeon-facing visualizer disappears, replaced by a single line: `failed to set scheduler pattern` — no SDK error code, no target mode, no recovery path.

Recommended fix: catch the failure at the boundary where the user request is consumed (i.e. wrap the `apply_pending_command(...)` call in `compute()` at line 307 with a try/catch), log the SDK error code and the requested mode at `HOLOSCAN_LOG_ERROR`, and refuse the mode change without exiting `compute()`. The previously-applied pattern remains in effect, the user can try again or fall back to a working mode, and the app stays alive.

Same hazard applies to the `dev.options().at("Image Scheduler Pattern")` lookup further down (and the five other unchecked `.at(...)` sites in this file) — older firmware that doesn't expose this option will throw `std::out_of_range("map::at")` from the same call path with no actionable message. Use `opts.count(...)` first.

applications/atracsys_visualizer/README.md
bhashemian: Thanks for this change and sorry for not being so clear. I meant to move the image to the top section only, and not the whole section. An splash image is needed to be on the top to adhere to the template: https://github.com/nvidia-holoscan/holohub/blob/main/applications/template/README.md.template

Thanks

bhashemian (commented):
Thanks @artrittelaku-wayland for addressing the feedback. The PR looks great now, but there are just a few comments that would be great if you could take a look. Once addressed or justifiably dismissed, we can merge this PR immediately. Thanks

applications/atracsys_visualizer/CMakeLists.txt
artrittelaku-wayland: Our original source zip file contained an unnecessary root folder which was causing nested extractions. I've re-packaged and updated the dataset zip on our release page so that holoscan_download_data handles it perfectly out of the box. I have now removed the custom normalization step and script.

operators/atracsys_camera/master_source_op.cpp
artrittelaku-wayland: I noticed this as well after bot comment and have explicitly added and called the missing destroyStereoParameters, destroyDefaultEngine, and destroyGpu3DFrame methods during stop() and failure conditions, effectively safely clearing the memory before assigning nullptr.

operators/atracsys_camera/master_source_op.cpp
artrittelaku-wayland: I've updated the geometry loaders to return explicit ftkError codes instead of hiding them. Now, if geometry loading fails inside load_geometries(), the operator logs the exact failure code and immediately throws a std::runtime_error to halt the pipeline, rather than starting up with silent tracking failures.

operators/atracsys_camera/master_source_op.cpp
artrittelaku-wayland: I've wrapped the mode switch in a try/catch block within compute() so it just logs the error and keeps the app alive if a switch fails. I also updated all the .at(...) calls to check .count() first, which prevents those silent crashes on older firmware.

bhashemian (approved):
LGTM! Thanks @artrittelaku-wayland for your contribution. Just could you please review the Greptile's comments and either address or dismiss them. When all marked as resolved, the PR will be merged automatically. Thanks

operators/atracsys_camera/master_source_op.cpp
```diff
@@ -0,0 +1,812 @@
...
+
+    is_running_ = true;
+    async_cond_->event_state(
+        holoscan::AsynchronousEventState::EVENT_WAITING);
+    capture_thread_ =
+        std::thread(&AtracsysMasterSourceOp::capture_loop, this);
+  } catch (...) {
+    destroy_frame();
+    AtracsysDevice::instance().shutdown();
+    throw;
```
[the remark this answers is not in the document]
artrittelaku-wayland: Addressed. The IS3DKInterface wrapper was updated to expose the proper explicit destruction methods (destroyStereoParameters, destroyDefaultEngine, destroyGpu3DFrame), and they are now safely invoked in both the start() exception catch-block and the stop() lifecycle method to prevent any memory leaks.

operators/atracsys_camera/master_source_op.cpp
```diff
@@ -0,0 +1,839 @@
...
+    emit_marker_poses(op_output, context, cuda_stream);
+  }
+  if (is_structured_format(format)) {
+    emit_structured_points(op_output, context, cuda_stream);
+  }
+
+  if (async_cond_) {
+    async_cond_->event_state(holoscan::AsynchronousEventState::EVENT_WAITING);
+  }
+}
```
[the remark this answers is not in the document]
artrittelaku-wayland: Dismissing as a false positive. Because the SDK uses standard, unpinned host memory, cudaMemcpy2DAsync automatically blocks the CPU until the data is safely copied to an internal staging area. There is no actual race condition here.

operators/atracsys_camera/master_source_op.cpp
```diff
@@ -0,0 +1,865 @@
...
+  const int height = static_cast<int>(frame_->imageHeader->height);
+  auto& roi = gpu_frame_->_ROIbox;
+
+  const bool roi_invalid = (roi.x_min < 0) || (roi.y_min < 0) || (roi.x_max <= roi.x_min) ||
+                           (roi.y_max <= roi.y_min) || (roi.x_max > height) || (roi.y_max > width);
+  if (roi_invalid) {
+    roi.x_min = 0;
+    roi.y_min = 0;
+    roi.x_max = height;
+    roi.y_max = width;
```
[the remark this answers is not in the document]
artrittelaku-wayland: Dismissing as a false positive. The Atracsys sensor buffer at this stage is transposed (column-major layout), meaning the logical x coordinate actually maps to the physical image height in memory, and y maps to width. The check is intentionally written this way to match the hardware's raw buffer orientation.

PR #1522, opened 2026-04-21
cli: strip ANSI color codes when stdout is not a TTY

agirault:
## Summary

The `Color` class in `utilities/cli/util.py` unconditionally emitted ANSI escape codes, which leaked into scripts parsing CLI output (e.g., `./holohub list | awk ...`).

This PR makes color emission conditional on the destination stream and environment:
- **`NO_COLOR=<non-empty>`** - always strip colors ([no-color.org](https://no-color.org/))
- **`FORCE_COLOR=<non-empty>`** - always emit colors, even if piped
- Otherwise, emit only when the destination stream is a TTY

`Color.format()` and the color methods (`Color.red`, `Color.yellow`, etc.) now accept an optional `stream` parameter. `warn()` and `fatal()` pass `stream=sys.stderr` so the TTY check targets the correct destination (e.g., when stdout is piped but stderr still goes to the terminal).

This matches what `colorama` / `click` do and follows the standard Unix convention.

## Before / After

Before:
```
$ ./holohub list | head -3
\x1b[37m\x1b[1m== APPLICATIONS =================\x1b[0m

aja_source (cpp)
```

After:
```
$ ./holohub list | head -3
== APPLICATIONS =================

aja_source (cpp)
```

Interactive terminal output is unchanged.

## Test plan

- [x] `./holohub list` in a terminal - colors still appear
- [x] `./holohub list | cat` - no ANSI codes
- [x] `./holohub list | awk ...` from a script - clean output parseable with awk/sed
- [x] `NO_COLOR=1 ./holohub list` in a terminal - no colors
- [x] `FORCE_COLOR=1 ./holohub list | cat` - colors even when piped
- [x] Pipe stdout only (e.g. `./holohub foo | cat`) - warnings/errors on stderr still colored if stderr is a TTY

### DIFF
--- a/utilities/cli/util.py
+++ b/utilities/cli/util.py
@@ -55,7 +55,11 @@ BUILD_TYPES = {
 
 
 class Color:
-    """Utility class for terminal color formatting"""
+    """Utility class for terminal color formatting.
+
+    ANSI codes are emitted only when stdout is a TTY. When piped or redirected,
+    all formatting is stripped so scripts can parse the output cleanly.
+    """
 
     # ANSI color codes
     RED = "\033[31m"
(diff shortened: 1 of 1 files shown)


wyli (approved):

bhashemian (approved):
LGTM! Thanks @agirault for this update.

PR #1533, opened 2026-05-03
feat(cli): auto-detect X11/Wayland display forwarding

agirault:
## Summary

- Replace flag-driven display forwarding with environment-based auto-detection
- Drop `xhost +local:docker` in favor of cookie-based xauth (per-run mktemp, mode 0600, ro mount, FamilyWild rewrite)
- Add Wayland support (mount only the socket, not the whole runtime dir)
- Deprecate `--enable-x11` and `--ssh-x11` flags (kept as no-ops with a stderr warning, removal in a follow-up release)

## Why

The container CLI previously required `--ssh-x11` to handle SSH-forwarded displays and called `xhost +local:docker`, which opens X access to every process in the host docker group. It also used a stable `/tmp/.docker.xauth` path that races between concurrent users on the same host, and did not handle Wayland sessions.

The new code path detects the environment from `DISPLAY` and `WAYLAND_DISPLAY` and supports native local X11, SSH-forwarded TCP X11, Wayland, and Xwayland in a single code path with cookie-based auth scoped to the calling user:

- per-run mktemp xauth file (mode 0600, mounted read-only)
- SSH-forwarded displays (`localhost:N`, `127.0.0.1:N`, `[::1]:N`) skip the `/tmp/.X11-unix` mount
- Wayland: bind-mount only `$XDG_RUNTIME_DIR/$WAYLAND_DISPLAY`, not the whole runtime dir
- cookie temp file removed after `docker run` returns
- no flag, no manual `xhost`, no manual `xauth merge`

Also adds `WAYLAND_DISPLAY` passthrough to the devcontainer environment so VS Code Remote Containers picks up the new path.

## Test plan

- [x] `python3 -m unittest utilities.cli.tests.test_container` (23 tests pass)
- [x] `./holohub run-container` on a native X11 desktop (`DISPLAY=:0`): GUI renders
- [ ] `./holohub run-container` over SSH `-X` (`DISPLAY=localhost:NN`): GUI renders
- [ ] `./holohub run-container` on a Wayland session (Ubuntu 24.04+): GUI renders via Xwayland
- [x] `./holohub run-container --ssh-x11` prints deprecation warning and still works
- [x] `./holohub run-container --enable-x11` prints deprecation warning and still works
- [x] Headless (no `DISPLAY`/`WAYLAND_DISPLAY`): runs without errors, info message logged

## Summary by CodeRabbit

* **New Features**
  * Graphics forwarding now auto-detects and forwards X11 or Wayland from host DISPLAY or WAYLAND_DISPLAY (including devcontainer support).

* **Documentation**
  * Updated container graphics docs to reflect automatic detection.
  * Marked `--ssh-x11` and `--enable-x11` CLI options as deprecated.

* **Tests**
  * Added tests for no-display, native X11, TCP X11, and Wayland forwarding scenarios.

### DIFF
--- a/utilities/cli/container.py
+++ b/utilities/cli/container.py
@@ -99,6 +99,9 @@ class HoloHubContainer:
     DEFAULT_DOCKER_BUILD_ARGS = os.environ.get("HOLOHUB_DEFAULT_DOCKER_BUILD_ARGS", "")
     # Additional Default run arguments for docker run command
     DEFAULT_DOCKER_RUN_ARGS = os.environ.get("HOLOHUB_DEFAULT_DOCKER_RUN_ARGS", "")
+    DISPLAY_FORWARDING_DISABLED_MESSAGE = (
+        "No DISPLAY or WAYLAND_DISPLAY set; skipping display forwarding."
+    )
 
     @classmethod
     def default_base_image(cls, cuda_version: Optional[Union[str, int]] = None) -> str:
@@ -452,6 +455,7 @@ class HoloHubContainer:
         self.cuda_version = None  # None means use default from get_cuda_tag
         self.dryrun = False
         self.verbose = False
+        self._display_temp_files: List[Path] = []
 
     def build(
         self,
@@ -610,6 +614,7 @@ class HoloHubContainer:
         cmd.extend(self.ucx_args())
         cmd.extend(self.get_device_mounts())
         cmd.extend(self.group_args())
+        self._display_temp_files = []
         cmd.extend(self.get_display_options(enable_x11, ssh_x11))
         cmd.extend(self.get_nsys_options(nsys_profile, nsys_location))
         cmd.extend(self.get_pythonpath_options(local_sdk_root, img))
@@ -632,7 +637,10 @@ class HoloHubContainer:
             cmd_list = [f'"{arg}"' if " " in str(arg) else str(arg) for arg in cmd]
             print(f"Launch command: {' '.join(cmd_list)}")
 
-        run_command(cmd, dry_run=self.dryrun)
+        try:
+            run_command(cmd, dry_run=self.dryrun)
+        finally:
+            self._cleanup_display_temp_files()
 
     def get_basic_args(self) -> List[str]:
         """Basic container runtime arguments"""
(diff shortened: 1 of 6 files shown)


agirault:
Force-pushed `auto-display-forwarding` to address review feedback. New commits:

- `2cb742113` feat(cli): auto-detect X11/Wayland display forwarding (Signed-off-by)
- `eeb1b95f9` chore(cli): deprecate --enable-x11 and --ssh-x11 flags (Signed-off-by)

Addressed:

- @greptile-apps / @coderabbitai (`@staticmethod` with `self` on `enable_x11_access`): deleted the dead stub. The method had no remaining callers; the deprecation lives at the CLI flag layer instead.
- @coderabbitai (short `xauth nlist` lines): added a `len(line) >= 4` guard so malformed/short lines are skipped rather than producing `ffff` followed by an empty tail.
- @greptile-apps (temp file created during dry-run): added an early return in `_get_xauth_options` when `self.dryrun` is true, returning a placeholder path so the printed docker command still demonstrates the intended mounts and no real tempfile is created on disk.
- @github-code-quality (empty `except OSError`): added an explanatory comment ("best-effort cleanup; ignore if the file is already gone").
- DCO: both commits now carry `Signed-off-by` trailers.

Also folded in the parity fix from the holoscan-sdk MR review: `warn()` is now emitted when `xauth nlist` returns no entries or `xauth nmerge` fails, instead of silently dropping the cookie mount.

Tests: `python3 -m unittest utilities.cli.tests.test_container` -> 23/23 pass.

agirault:
Round 2 follow-up. Force-pushed:

- `5309157da` feat(cli): auto-detect X11/Wayland display forwarding (Signed-off-by)
- `aa843ffbb` chore(cli): deprecate --enable-x11 and --ssh-x11 flags (Signed-off-by)

Addressed:

- @coderabbitai (line 848 - new): added `warn()` when `shutil.which("xauth")` returns None, so users see why GUI apps fail to authenticate when xauth is not installed on the host. Same fix applied symmetrically in the holoscan-sdk MR.

wyli (approved):

utilities/cli/container.py
```diff
@@ -772,75 +793,116 @@ def get_environment_args(self) -> List[str]:
...
+
+        self._display_temp_files.append(xauth_path)
+        return ["-v", f"{xauth_path}:{xauth_path}:ro", "-e", f"XAUTHORITY={xauth_path}"]
+
+    def _cleanup_display_temp_files(self) -> None:
+        for path in self._display_temp_files:
+            try:
+                path.unlink(missing_ok=True)
+            except OSError:
+                # Best-effort cleanup; ignore if the file is already gone.
```
bhashemian: "if the file is already gone." is being taken care in `missing_ok=True`, so this one accounts for other failures. Since `_cleanup_display_temp_files` is being called in `finally`, this maybe more informative:
```suggestion
                # Suppress I/O errors so cleanup doesn't mask the original exception.
```

bhashemian (approved):

utilities/cli/container.py
agirault: Good catch - applied verbatim in 49fb5b411.

agirault:
Fyi @kimullaa in case you run into issues with this, I saw you had a related MR.

utilities/cli/container.py
```diff
@@ -772,75 +793,100 @@ def get_environment_args(self) -> List[str]:
...
 
-        # Handle X11 forwarding
-        if enable_x11 or ssh_x11:
-            # Enable X11 access for Docker containers
-            self.enable_x11_access()
-            options.extend(["-v", "/tmp/.X11-unix:/tmp/.X11-unix", "-e", "DISPLAY"])
+        # Required by Vulkan, dconf, pipewire, etc. on both X11 and Wayland.
+        runtime_dir = os.environ.get("XDG_RUNTIME_DIR")
+        if runtime_dir and Path(runtime_dir).is_dir():
+            options.extend(["-e", "XDG_RUNTIME_DIR", "-v", f"{runtime_dir}:{runtime_dir}"])
```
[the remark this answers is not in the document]
agirault: Yes we use holoviz which is on vulkan which needs it.

PR #1563, opened 2026-05-18
Modules: Extend Module schema with presentation metadata

tbirdso:
## Context

Piloting presentation of Holoscan Module domain-specific libraries leveraging existing Holoscan ecosystem tooling
    
## Changes
    - Add sensor_hardware, homepage, documentation, and quality_reporting fields to module.schema.json
    - Extend binary_packages with install_commands support
    - Add module_with_documentation_and_quality.json fixture to cover new fields in schema regression tests

## Results

Unit tests pass locally

## Not In Scope (Planned for Later)

- Fetch and validate external module information

Assisted-by: Claude:claude-sonnet-4-6[1m]
Signed-off-by: Tom Birdsong <tbirdsong@nvidia.com>

### DIFF
--- a/utilities/metadata/module.schema.json
+++ b/utilities/metadata/module.schema.json
@@ -79,6 +79,7 @@
         },
         "date": {
           "type": "string",
+          "format": "date",
           "description": "ISO 8601 date of the test run (e.g. '2026-05-01')."
         },
         "holoscan_sdk_version": {
@@ -215,6 +216,7 @@
         },
         "homepage": {
           "type": "string",
+          "format": "uri",
           "description": "URL of the module's primary landing page or repository."
         },
         "source_repository": {
--- a/utilities/metadata/module.schema.json
+++ b/utilities/metadata/module.schema.json
@@ -65,7 +65,7 @@
           "items": {
             "type": "string"
           },
-          "description": "Relative paths to images, GIFs, or other assets the documentation builder must fetch for webpage rendering."
+          "description": "Relative paths to images, GIFs, or other assets the documentation builder must fetch for webpage rendering. Assumed relative to the project root."
         }
       },
       "additionalProperties": false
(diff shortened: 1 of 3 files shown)


utilities/metadata/module.schema.json
```diff
@@ -45,6 +52,79 @@
...
+          }
+        },
+        "tests_passing": {
+          "type": "integer",
+          "minimum": 0
+        },
+        "tests_failing": {
+          "type": "integer",
+          "minimum": 0
+        },
```
tbirdso, line 101: @bhashemian I recall we discussed useful test metrics but I don't recall the resolution. Do you have suggestions as to how we can support self-reported test results with meaningful fields?

wyli (approved):

utilities/metadata/module.schema.json
```diff
@@ -45,6 +52,80 @@
...
+        "readme": {
+          "type": "string",
+          "description": "Relative path from this metadata.json to the module README. Defaults to README.md in the same directory when absent."
+        },
+        "assets": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "Relative paths to images, GIFs, or other assets the documentation builder must fetch for webpage rendering."
```
wyli: relative to the readme or the metadata.json?

tbirdso: Good point. Let's assume relative to the project root, which is where we expect to find the metadata.json file.

Note that the tooling for this is not implemented today, but agreed that it is good to set the contract.

```suggestion
          "description": "Relative paths to images, GIFs, or other assets the documentation builder must fetch for webpage rendering. Assumed relative to the project root."
```

utilities/metadata/module.schema.json
tbirdso: Moving forward to unblock dependent tasks. Not a blocker, can easily revisit these optional fields.

PR #1583, opened 2026-05-29
Migrate the HoloHub CLI to the standalone holoscan-cli package

wyli:
## What

Migrate HoloHub from the in-tree `utilities/cli` implementation to the standalone [`holoscan-cli`](https://github.com/nvidia-holoscan/holoscan-cli) package.

`./holohub` is now a thin wrapper that bootstraps and delegates to `holoscan-cli`, while still supporting local CLI development through `HOLOSCAN_CLI_SOURCE`. The superseded in-tree Python CLI implementation and its unit tests are removed from HoloHub.

Supersedes #1552 with the same migration organized as smaller commits.

## Scope

- Update the root `./holohub` wrapper, generated module wrapper, and external-application tutorial wrapper to install/delegate to `holoscan-cli`.
- Use `HOLOSCAN_CLI_INSTALL_ARGS` as the single install knob for the CLI package spec plus pip options.
- Bootstrap `holoscan-cli` in CI and the relevant Docker layers so wrapper-driven builds and standalone container builds use the same CLI.
- Preserve `HOLOSCAN_CLI_SOURCE` for developing HoloHub against a local `holoscan-cli` checkout.
- Rename affected integration-test environment variables from `HOLOHUB_*` to `HOLOSCAN_CLI_*`.
- Update module templates, docs, benchmark helpers, and setup scripts for the standalone CLI layout.
- Keep HoloHub-owned check dependencies in `utilities/requirements.check-cli.txt`.
- Add wrapper/CTest smoke coverage for both the released `holoscan-cli` package path and a source checkout path.

## Install Behavior

The default install args are:

```bash
--pre --extra-index-url https://pypi.nvidia.com holoscan-cli>4.2.0
```

This intentionally allows the current 4.3 prerelease from NVIDIA PyPI. A `>=4.3.0` spec would not select release candidates because prereleases sort lower than the final release under PEP 440.

## Setup Scripts

`utilities/setup` stays in this PR. In the current `holoscan-cli` prerelease, setup script resolution is directory-level: either the repo-local `utilities/setup` directory is used, or the bundled upstream setup scripts are used when the repo-local directory is absent.

The local setup scripts were aligned with that behavior:

- `requirements.template.txt` now lives beside `template.sh` under `utilities/setup/`.
- `template.sh` reads the sibling requirements file.
- `Dockerfile.util` copies the setup-script directory context so scripts with adjacent files work in extra-script Docker layers.

Removing `utilities/setup` entirely can be a follow-up once HoloHub is ready to rely fully on bundled setup scripts, or after `holoscan-cli` supports merged repo-local plus bundled setup-script lookup.

## Validation

- Required GitHub checks are passing on the latest head.
- PR is mergeable and approved.
- `check-cli-py` passes across Python 3.10, 3.11, 3.12, and 3.13.
- `check-cli-ctest` and `check-cli-source` pass.
- `check-lint`, `check-metadata`, `check-urls`, `compliance-check`, `CodeQL`, `DCO`, `pre-commit.ci`, and docs build/deploy checks pass.
- Focused local validation covered:

```bash
pytest -q --noconftest \
    utilities/cli/tests/test_holoscan_cli_consolidation.py \
    utilities/cli/tests/test_create_module.py
```

- CMake/CTest smoke suite: `100% tests passed, 0 tests failed out of 10`.
- `git diff --check`, shell syntax checks, and markdownlint checks passed for touched files.

AI-assisted: Created with Codex/GPT at the user's request.


Dockerfile
```diff
@@ -15,69 +15,83 @@
...
+# one that can run the HoloHub CLI. Each step is a no-op when the base
+# already provides the requirement.
+############################################################
+FROM ${BASE_IMAGE} AS holohub-cli-prerequisites
 
-# Install python3 if not present (needed for holohub CLI)
+ARG DEBIAN_FRONTEND=noninteractive
 ARG PYTHON_VERSION=python3
+
+# 1. Ensure python3 + pip exist. Install pip via get-pip.py, NOT apt
```
wyli: the root `Dockerfile` is simplified per your previous comment @tbirdso. now the advanced usages are only included in `/utilities/docker/Dockerfile.holohub-base` (maybe this file could be upstreamed to holoscan-cli package as well), let me know your thoughts

holohub
```diff
@@ -14,6 +14,61 @@
...
+PYTHON_BIN="${HOLOHUB_PYTHON_BIN:-python3}"
+
+# Pinned holoscan-cli build. Env vars take precedence so users can opt into a
+# different version without editing.
+export HOLOSCAN_CLI_VERSION="${HOLOSCAN_CLI_VERSION:-holoscan-cli==4.3.0a26390596878}"
+
+# A local holoscan-cli checkout takes precedence over the pinned install.
+if [[ -n "${HOLOSCAN_CLI_SOURCE:-}" ]]; then
+    export PYTHONPATH="${HOLOSCAN_CLI_SOURCE%/}/src${PYTHONPATH:+:${PYTHONPATH}}"
+fi
```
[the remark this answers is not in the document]
wyli: Addressed in f534db1: the wrapper now normalizes a bare version or comparator (`4.3.0`, `>=4.3`) to a valid `holoscan-cli` requirement via `_holohub_normalize_cli_spec` before passing it to `pip install` and before forwarding it as the `HOLOSCAN_CLI_INSTALL_SPEC` build-arg.

modules/template/{{cookiecutter.module_repo_name}}/holohub
```diff
@@ -3,120 +3,88 @@
...
-export HOLOHUB_BUILD_PARENT_DIR="${HOLOHUB_BUILD_PARENT_DIR:-${SCRIPT_DIR}/build}"
-export HOLOHUB_DATA_DIR="${HOLOHUB_DATA_DIR:-${SCRIPT_DIR}/data}"
-export HOLOHUB_SETUP_SCRIPTS_DIR="${HOLOHUB_SETUP_SCRIPTS_DIR:-${CLI_DIR}/utilities/setup}"
+PYTHON_BIN="${HOLOHUB_PYTHON_BIN:-python3}"
 
-# ==============================================================================
-# Build defaults
-# ==============================================================================
+# Pinned holoscan-cli build (matches the HoloHub root wrapper).
+export HOLOSCAN_CLI_VERSION="${HOLOSCAN_CLI_VERSION:-holoscan-cli==4.3.0a26390596878}"
```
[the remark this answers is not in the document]
wyli: Addressed in 9b12465: the generated module wrapper now normalizes bare `HOLOSCAN_CLI_VERSION` values (`4.3.0`, `>=4.3`) to a valid pip requirement, matching the tutorial wrapper.

utilities/cli/README.md
```diff
@@ -19,9 +31,36 @@ A command-line interface for managing Holoscan-based applications and workflows.
...
+
+   | Env var                | Behavior                                                                       |
+   | ---------------------- | ------------------------------------------------------------------------------ |
+   | `HOLOSCAN_CLI_SOURCE`  | Path to a local checkout. Wins over `HOLOSCAN_CLI_VERSION`.                    |
+   | `HOLOSCAN_CLI_VERSION` | Pip spec — `4.3.0`, `>=4.3`, `git+https://...`. Defaults to `holoscan-cli`.    |
+
+3. Exports the `HOLOSCAN_CLI_*` environment variables that configure the CLI
+   for this repo, then delegates the user's argv to
+   `python3 -m holoscan_cli`.
+
```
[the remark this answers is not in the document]
wyli: Addressed in c33a764: the README now describes the actual command-surface probe (`python3 -m holoscan_cli --help | grep -qw build`), the pinned default spec (`holoscan-cli==4.3.0rc1`), and that bare `HOLOSCAN_CLI_VERSION` values are normalized.

utilities/docker/Dockerfile.holohub-base
```diff
@@ -0,0 +1,100 @@
...
+        curl -sS https://bootstrap.pypa.io/get-pip.py | ${PYTHON_VERSION} \
+    ; fi
+
+# Install `holoscan-cli`. Two-step (matches CI): the wheel from
+# TestPyPI (--no-deps), then deps from PyPI so a TestPyPI mirror can't shadow a
+# PyPI dep. Escape hatches for iterating on the CLI itself:
+#   * `--build-context holoscan-cli-src=<path>` installs a local checkout (wins).
+#   * HOLOSCAN_CLI_INSTALL_SPEC pins the version.
+ARG HOLOSCAN_CLI_INSTALL_SPEC=holoscan-cli
+
```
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: `HOLOSCAN_CLI_INSTALL_SPEC` now defaults to the pinned `holoscan-cli==4.3.0rc1` (matching the wrapper and CI), so a standalone build of this base image is reproducible; the build-arg remains the override escape hatch.

utilities/docker/Dockerfile.holohub-base
```diff
@@ -0,0 +1,100 @@
...
+        && apt purge -y \
+            python3-pip \
+            software-properties-common \
+        && apt-get autoremove --purge -y \
+        && rm -rf /var/lib/apt/lists/* \
+        && update-alternatives --install /usr/bin/python python /usr/bin/${PYTHON_VERSION} 100 \
+        && if [ "${PYTHON_VERSION}" != "python3" ]; then \
+            update-alternatives --install /usr/bin/python3 python3 /usr/bin/${PYTHON_VERSION} 100 \
+            ; fi \
+    ; fi
```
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: the prerequisite stage now guards on `${PYTHON_VERSION}` rather than bare `python3`, so a `PYTHON_VERSION` override still installs the requested interpreter.

Dockerfile
```diff
@@ -15,69 +15,83 @@
...
+#    `${PYTHON_VERSION}-pip`: the debian-managed pip lacks a RECORD file, so a
+#    transitive `pip>25.1.0` upgrade from holoscan-cli aborts with
+#    "Cannot uninstall pip 24.0, RECORD file not found".
 RUN if ! command -v python3 >/dev/null 2>&1; then \
         apt-get update \
```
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: the guard now uses `! command -v "${PYTHON_VERSION}"` (not bare `python3`), so overriding `PYTHON_VERSION` on a base that already ships python3 still installs the requested interpreter.

operators/gstreamer/Dockerfile
```diff
@@ -18,56 +18,70 @@
...
-# Install python3 if not present (needed for holohub CLI)
+ARG DEBIAN_FRONTEND=noninteractive
 ARG PYTHON_VERSION=python3
+
+# Do NOT apt-install ${PYTHON_VERSION}-pip on Ubuntu 24.04+: that pip is
+# debian-managed and missing the RECORD file, so a transitive
+# `pip>25.1.0` upgrade from holoscan-cli aborts with `Cannot uninstall
+# pip 24.0, RECORD file not found`. Install pip via get-pip.py instead.
 RUN if ! command -v python3 >/dev/null 2>&1; then \
         apt-get update \
```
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: the guard now uses `! command -v "${PYTHON_VERSION}"` rather than bare `python3`.

Dockerfile
```diff
@@ -15,69 +15,83 @@
...
-    ; fi
+            update-alternatives --install /usr/bin/python3 python3 "/usr/bin/${PYTHON_VERSION}" 100; \
+        fi; \
+    fi
 ENV PIP_BREAK_SYSTEM_PACKAGES=1
 RUN if ! python3 -m pip --version >/dev/null 2>&1; then \
-        curl -sS https://bootstrap.pypa.io/get-pip.py | ${PYTHON_VERSION} \
-    ; fi
+        curl -sS https://bootstrap.pypa.io/get-pip.py | ${PYTHON_VERSION}; \
+    fi
```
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: the guard now also fires when `curl` is missing (`! command -v "${PYTHON_VERSION}" || ! command -v curl`) and installs `curl ca-certificates` alongside python, so the get-pip.py bootstrap has curl + CA roots even when python is already present.

operators/gstreamer/Dockerfile
```diff
@@ -18,56 +18,70 @@
...
-    ; fi
+            update-alternatives --install /usr/bin/python3 python3 "/usr/bin/${PYTHON_VERSION}" 100; \
+        fi; \
+    fi
 ENV PIP_BREAK_SYSTEM_PACKAGES=1
 RUN if ! python3 -m pip --version >/dev/null 2>&1; then \
-        curl -sS https://bootstrap.pypa.io/get-pip.py | ${PYTHON_VERSION} \
-    ; fi
+        curl -sS https://bootstrap.pypa.io/get-pip.py | ${PYTHON_VERSION}; \
+    fi
```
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: the guard now also fires when `curl` is missing, installing `curl ca-certificates` so the get-pip.py bootstrap works even when python is already present.

utilities/docker/Dockerfile.holohub-base
```diff
@@ -0,0 +1,100 @@
...
+
+ENV PIP_BREAK_SYSTEM_PACKAGES=1
+RUN if ! python3 -m pip --version >/dev/null 2>&1; then \
+        curl -sS https://bootstrap.pypa.io/get-pip.py | ${PYTHON_VERSION} \
+    ; fi
```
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: a dedicated guard now installs `curl` + `ca-certificates` before the get-pip.py bootstrap for the case where the base already had python but lacks curl/CA roots.

utilities/docker/Dockerfile.holohub-base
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: the default is now pinned to `holoscan-cli==4.3.0rc1`, matching the repo-root wrapper and CI.

holohub
```diff
@@ -14,6 +14,69 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
+export HOLOSCAN_CLI_CMD_NAME="${BASH_SOURCE[0]}"
+export HOLOSCAN_CLI_ROOT="${SCRIPT_DIR}"
+
+PYTHON_BIN="${HOLOHUB_PYTHON_BIN:-python3}"
+
+# Pinned holoscan-cli build. Env vars take precedence so users can opt into a
+# different version without editing.
+export HOLOSCAN_CLI_VERSION="${HOLOSCAN_CLI_VERSION:-holoscan-cli==4.3.0a26885294575}"
+
```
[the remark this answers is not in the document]
wyli: Addressed: every pin is now `holoscan-cli==4.3.0rc1` (wrapper, Dockerfiles, cookiecutter template, and CI), and the PR description has been updated to match.

utilities/docker/Dockerfile.holohub-base
```diff
@@ -0,0 +1,100 @@
...
+FROM base AS holohub-base
+
+ARG DEBIAN_FRONTEND=noninteractive
+ARG PYTHON_VERSION=python3
+
+# Install python3 if not present (needed for holohub CLI)
+RUN if ! command -v python3 >/dev/null 2>&1; then \
+        apt-get update \
+        && apt-get install --no-install-recommends -y \
+            software-properties-common curl gpg-agent \
```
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: `ca-certificates` is now installed in the python bootstrap apt step, and a dedicated `curl` + `ca-certificates` guard runs before the HTTPS get-pip.py bootstrap so minimal bases have CA roots.

utilities/docker/Dockerfile.holohub-base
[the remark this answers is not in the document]
wyli: Addressed in 1ff63a4: `HOLOSCAN_CLI_INSTALL_SPEC` now defaults to the pinned `holoscan-cli==4.3.0rc1`, so a direct `docker build -f utilities/docker/Dockerfile.holohub-base` is reproducible.

holohub
```diff
@@ -14,6 +14,69 @@
...
+if [[ -n "${HOLOSCAN_CLI_SOURCE:-}" ]]; then
+    export PYTHONPATH="${HOLOSCAN_CLI_SOURCE%/}/src${PYTHONPATH:+:${PYTHONPATH}}"
+fi
+
+# Install holoscan-cli unless a source-project-capable one is already present.
+# Probe the command surface (not just importability), so a stale legacy
+# holoscan-cli (<= 4.2.0, which lacks `build`/`list`) doesn't satisfy the check.
+# Two-step install keeps TestPyPI scoped to the holoscan-cli wheel so a TestPyPI
+# mirror of a common dep can't shadow the PyPI one.
+if ! "${PYTHON_BIN}" -m holoscan_cli --help 2>/dev/null | grep -qw build; then
```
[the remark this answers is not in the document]
wyli: Addressed in f534db1: the wrapper now bootstraps pip via `ensurepip --upgrade` (and exits with a clear instruction if pip is still unavailable) before attempting `pip install`.

tutorials/holohub_operators_external_applications/holohubCLI_wrapper
```diff
@@ -16,93 +16,77 @@
...
+# cwd (otherwise `python -m holoscan_cli` scans the caller's cwd and builds the
+# wrong tree). Mirrors the repo-root `./holohub` wrapper.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
+export HOLOSCAN_CLI_CMD_NAME="${BASH_SOURCE[0]}"
+export HOLOSCAN_CLI_ROOT="${SCRIPT_DIR}"
 
-    cd "${TMP_HOLOHUB_DIR}/holohub"
-    git sparse-checkout init --cone >/dev/null 2>&1
-    git sparse-checkout set utilities/cli utilities/metadata utilities/testing utilities/setup cmake >/dev/null 2>&1
+PYTHON_BIN="${HOLOSCAN_CLI_PYTHON_BIN:-python3}"
```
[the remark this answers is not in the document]
wyli: Addressed in c33a764: the example wrapper now exports `HOLOSCAN_CLI_BASE_SDK_VERSION` (default `4.3.0`) so `run`/`build`/`package` keep working when the wrapper is copied into an external project.

utilities/cli/README.md
```diff
@@ -19,9 +31,36 @@ A command-line interface for managing Holoscan-based applications and workflows.
...
 ```
 
-## Documentation
+## How the wrapper resolves the CLI
+
+The `./holohub` script at the repo root does three things in order:
+
+1. Checks that `holoscan_cli` is importable. The right version comes from the
+   pinned install spec (below), so the wrapper just bootstraps when nothing is
+   installed rather than probing the command surface itself.
```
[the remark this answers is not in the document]
wyli: Addressed in c33a764: the README now documents the command-surface probe (`python3 -m holoscan_cli --help | grep -qw build`) rather than a plain importability check, so a legacy holoscan-cli (<= 4.2.0) won't satisfy it.

.github/workflows/check_cli.yml
```diff
@@ -63,3 +83,42 @@ jobs:
         run: |
           cd build
           ctest --verbose
+
+  # Exercise the same smoke test against a *source checkout*
+  # of holoscan-cli (https://github.com/nvidia-holoscan/holoscan-cli) rather
+  # than the pinned NVIDIA PyPI wheel. This catches breakage when an in-flight
+  # holoscan-cli change lands ahead of a new release.
```
tbirdso: Good idea!

.github/workflows/check_lint.yml
```diff
@@ -17,8 +25,44 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v4
 
-      - name: Install Dependencies
-        run: sudo ./holohub lint --install-dependencies
+      # Pre-install the pinned holoscan-cli from NVIDIA PyPI so the
+      # ./holohub bootstrap doesn't reach public PyPI for a stale 4.2.0
+      # build that predates the ``lint`` subcommand. Revert this step to
```
tbirdso: Shouldn't we guard against stale builds with ">=v4.3.0" version pinning?

.github/workflows/check_lint.yml
```diff
@@ -9,6 +9,14 @@ on:
 permissions:
   contents: read
 
+# Pinned holoscan-cli build used by CI while the release is still on
+# NVIDIA PyPI. Bump this single value to roll the version across all jobs;
+# drop the index-url overrides below once the release is promoted to
+# public PyPI.
+env:
+  HOLOSCAN_CLI_SPEC: "holoscan-cli==4.3.0rc2"
```
tbirdso: What about `holoscan-cli>=4.3.0 --pre` so that we can pick up latest RCs?

.github/workflows/check_lint.yml
```diff
@@ -17,8 +25,44 @@ jobs:
...
+      #
+      # Use a single interpreter across non-sudo / sudo / ./holohub bootstrap.
+      # On GitHub-hosted ubuntu-22.04 the runner exposes `python` only via a
+      # PATH shim that sudo strips (sudo's secure_path doesn't include the
+      # hostedtoolcache dirs); meanwhile the wrapper exec's `python3`. Resolve
+      # the absolute interpreter path once and reuse it everywhere so the
+      # pinned wheel lands in the same site-packages the wrapper imports from.
+      - name: Resolve Python interpreter
+        id: pybin
+        run: echo "path=$(command -v python3)" >> "${GITHUB_OUTPUT}"
```
tbirdso: Rather than a custom reference, could we standardize to `python3` everywhere or otherwise use `update-alternatives` to set "python" as an alias for the "python3" executable? Seems simpler and less prone to breakage

doc/developer.md
```diff
@@ -95,6 +95,26 @@ Where:
...
+The HoloHub default container layers application support on top of the
+prepared base image at `utilities/docker/Dockerfile.holohub-base`. That base
+image exposes several knobs (local checkout, install spec, extra pip flags) to
+support development against an unreleased `holoscan-cli`; when writing a brand
+new Dockerfile you usually do not need any of that.
+
+The smallest fragment that gives a custom Dockerfile the `holoscan` command is:
+
+```dockerfile
+FROM nvcr.io/nvidia/clara-holoscan/holoscan:v<sdk-version>-<gpu-type>
```
tbirdso: Wheel is sufficient, Holoscan container not needed for `holoscan` command

```suggestion
FROM ubuntu:24.04
```

operators/gstreamer/Dockerfile
```diff
@@ -18,58 +18,72 @@
...
         && apt-get install --no-install-recommends -y \
-            software-properties-common curl gpg-agent \
-        && add-apt-repository ppa:deadsnakes/ppa \
-        && apt-get update \
-        && apt-get install --no-install-recommends -y \
-            ${PYTHON_VERSION} ${PYTHON_VERSION}-dev \
-        # Conditionally install -distutils only if available (for older python versions)
-        && (apt-cache show ${PYTHON_VERSION}-distutils >/dev/null 2>&1 && \
-            apt-get install --no-install-recommends -y ${PYTHON_VERSION}-distutils || true) \
-        && apt purge -y \
```
tbirdso, line 46: Seems like this drops installation for `<python>-dev`, `<python>-distutils` packages. Can we restore?

operators/gstreamer/Dockerfile
```diff
@@ -18,58 +18,72 @@
...
-# --------------------------------------------------------------------------
-#
-# Set up GStreamer-specific dependencies
-#
-# --------------------------------------------------------------------------
+# Ensure the `holoscan` CLI exists. The legacy packaging-only CLI (holoscan-cli
+# <= 4.2.0) is also on PATH, so probe `holoscan --help` for a source-project
+# command (`build`) only the new CLI has. Use NVIDIA PyPI as an extra index
+# until the pinned release is promoted to public PyPI. HOLOSCAN_CLI_INSTALL_SPEC (the
+# wrapper forwards it) selects the version; default is the pinned build.
```
tbirdso: > The legacy packaging-only CLI (holoscan-cli <= 4.2.0) is also on PATH

Not sure this is correct, was this reflective of some intermediate testing environment maybe? I don't think the stale holoscan-cli <= 4.2.0 should be present at all


```suggestion
# Ensure the `holoscan` CLI exists. Use NVIDIA PyPI as an extra index
# until the pinned release is promoted to public PyPI. HOLOSCAN_CLI_INSTALL_SPEC (the
# wrapper forwards it) selects the version; default is the pinned build.
```

operators/gstreamer/Dockerfile
```diff
@@ -18,58 +18,72 @@
...
+# HoloHub CLI: stage the in-tree wrapper and smoke-test.
+############################################################
 FROM holohub-cli-prerequisites AS holohub-cli
 
-RUN mkdir -p /tmp/scripts
-COPY holohub /tmp/scripts/
+ARG CMAKE_BUILD_TYPE=Release
+
+# Stage the `holohub` wrapper + utilities/ helpers `holohub setup` needs, so
+# the raw-base `holohub setup` fallback below can run.
```
tbirdso: Do we need to install the `holoscan-cli` explicitly at all, or could we rely on the `./holohub` entrypoint script for default setup?

utilities/cli/README.md
```diff
@@ -19,9 +31,41 @@ A command-line interface for managing Holoscan-based applications and workflows.
...
 ```
 
-## Documentation
+## How the wrapper resolves the CLI
+
+The `./holohub` script at the repo root does three things in order:
+
+1. Probes the installed command surface — `python3 -m holoscan_cli --help |
+   grep -qw build` — rather than just checking importability. This ensures a
+   source-project-capable CLI is present, so a stale legacy `holoscan-cli`
```
tbirdso, line 40: Why not directly check version rather than probing command surfaces?

utilities/docker/Dockerfile.holohub-base
```diff
@@ -0,0 +1,111 @@
+# syntax=docker/dockerfile:1
+
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# ----------------------------------------------------------------------------
+# HoloHub prepared base image: stacks python+pip and `holoscan-cli` onto the
+# Holoscan SDK runtime and runs `holohub setup`, so downstream Dockerfiles can
+# `FROM ${BASE_IMAGE}` without re-provisioning.
```
tbirdso: What is the purpose of this Dockerfile? Seems like we still have the root-level Dockerfile default?

Dockerfile
```diff
@@ -15,69 +15,84 @@
...
-            software-properties-common curl gpg-agent \
-        && add-apt-repository ppa:deadsnakes/ppa \
-        && apt-get update \
-        && apt-get install --no-install-recommends -y \
-            ${PYTHON_VERSION} \
-        && apt purge -y \
-            python3-pip \
-            software-properties-common \
-        && apt-get autoremove --purge -y \
+            ${PYTHON_VERSION} curl ca-certificates \
```
tbirdso: `deadsnakes` PPA may be necessary for some Python installations, suggest we restore the previous approach

modules/template/{{cookiecutter.module_repo_name}}/holohub
tbirdso, line 1: Is this file fully aligned to the root level "holohub"? If so we could remove here and simply copy from the HoloHub script at template creation time. I can help in a follow-up as needed.

tbirdso (commented):
Thanks @wyli , changes are looking good! Left a few comments and clarification requests

utilities/docker/Dockerfile.holohub-base
wyli: this is an example if the user wants to support source of cli other than the standard pypi.

modules/template/{{cookiecutter.module_repo_name}}/holohub
wyli: this is not fully aligned as we may need some module specific env var settings. I think we should revisit this in separate PRs.

utilities/cli/README.md
wyli: main reason is that if we mount cli from source we may not have proper version strings

doc/developer.md
wyli: sure, note that if we use ubuntu, we'll need extra steps for pip
```
FROM ubuntu:24.04
RUN apt-get update \
    && apt-get install --no-install-recommends -y python3 python3-pip ca-certificates \
    && rm -rf /var/lib/apt/lists/*
ENV PIP_BREAK_SYSTEM_PACKAGES=1
```

.github/workflows/check_lint.yml
wyli: updated to `HOLOSCAN_CLI_INSTALL_ARGS: "--pre --extra-index-url https://pypi.nvidia.com holoscan-cli>4.2.0"`

operators/gstreamer/Dockerfile
wyli: Addressed in 4c42cf38a: `operators/gstreamer/Dockerfile` now installs the selected Python development headers independently of whether the interpreter binary already existed. For `PYTHON_VERSION=python3`, it installs `python3-dev`; for custom versions, it installs `${PYTHON_VERSION}-dev` and optional `${PYTHON_VERSION}-distutils` via deadsnakes.

operators/gstreamer/Dockerfile
wyli: Keeping the explicit `holoscan-cli` install here. This Dockerfile supports standalone `docker build` of the GStreamer image, before the in-tree `./holohub` wrapper is staged, and it keeps standalone builds aligned with wrapper-driven builds through the same `HOLOSCAN_CLI_INSTALL_ARGS` build arg. The wrapper bootstrap still remains as a fallback for normal host-side usage.

doc/developer.md
tbirdso: Good point, maybe https://hub.docker.com/_/python or https://catalog.ngc.nvidia.com/orgs/nvidia/containers/cuda would be better? Up to you, not a blocker

utilities/cli/README.md
tbirdso: Could we consider making a version string available in the source-only HoloHub CLI module as a future enhancement? Not a blocker

utilities/docker/Dockerfile.holohub-base
tbirdso: How should the user discover this file to use as an example? Wondering if a short tutorial or similar would be more discoverable, or else if this should move to examples in the Holoscan CLI repo?

operators/gstreamer/Dockerfile
```diff
@@ -43,26 +43,43 @@ RUN if ! command -v "${PYTHON_VERSION}" >/dev/null 2>&1; then \
...
-            update-alternatives --install /usr/bin/python3 python3 "/usr/bin/${PYTHON_VERSION}" 100; \
-        fi; \
+        && rm -rf /var/lib/apt/lists/*; \
+    fi
+RUN if [ "${PYTHON_VERSION}" != "python3" ]; then \
+        apt-get update \
+        && apt-get install --no-install-recommends -y \
+            software-properties-common ca-certificates gpg-agent \
+        && add-apt-repository ppa:deadsnakes/ppa \
+        && apt-get update \
```
tbirdso: Seems duplicated with previous layer?

operators/gstreamer/Dockerfile
```diff
@@ -43,26 +43,43 @@ RUN if ! command -v "${PYTHON_VERSION}" >/dev/null 2>&1; then \
...
+        update-alternatives --install /usr/bin/python3 python3 "/usr/bin/${PYTHON_VERSION}" 100; \
     fi
 RUN if ! command -v curl >/dev/null 2>&1; then \
         apt-get update \
         && apt-get install --no-install-recommends -y curl ca-certificates \
         && rm -rf /var/lib/apt/lists/*; \
     fi
 ENV PIP_BREAK_SYSTEM_PACKAGES=1
-RUN if ! python3 -m pip --version >/dev/null 2>&1; then \
+RUN if ! "${PYTHON_VERSION}" -m pip --version >/dev/null 2>&1; then \
```
tbirdso: Change seems unnecessary, shouldn't `update-alternatives` have already set `python3` to alias `${PYTHON_VERSION}`?

utilities/docker/Dockerfile.holohub-base
wyli: make sense, I'll remove and upstream this one as a tutorial

utilities/cli/README.md
wyli: yes, I'll follow up on this as we are aligned on the versioning

doc/developer.md
wyli: changing to nvcr.io/nvidia/cuda:13.3.0-devel-ubuntu24.04

operators/gstreamer/Dockerfile
wyli: thanks, simplified the Dockerfile related changes

tbirdso (approved):
Reviewed updates diff since last review. LGTM!

PR #1591, opened 2026-06-01
test: make CuPy/Holoscan optional in test fixtures

wyli:
## Summary

Several test `conftest.py` files imported `cupy` and `holoscan.core` at
module-import time, so collecting tests failed whenever those optional
dependencies were missing or unimportable (for example, CuPy installed
without a working GPU/driver, or a NumPy ABI mismatch). Because the
repo-root `conftest.py` is loaded for the entire session, this aborted
collection of **every** test in the repo — even tests unrelated to CuPy.

This change defers/guards those imports so that only tests which actually
request a CuPy- or Holoscan-backed fixture skip when the dependency is
unavailable:

- **`conftest.py` (root)** and the **`video_streaming_{server,client}`
  conftests** (near-identical copies of the root fixtures):
  - `app` / `fragment` import `holoscan.core` lazily via
    `pytest.importorskip`.
  - `mock_image` imports `cupy` only for the `cupy` backend and skips on
    any import failure (not just `ImportError`); its default `dtype` is
    now the `"uint8"` string instead of `cupy.uint8`, so the fixture no
    longer references CuPy at definition time.
  - The root `mock_image` float branch uses
    `rng.uniform(...).astype(dtype)`, which works on both the NumPy and
    CuPy `Generator` APIs (NumPy's `Generator.uniform` does not accept a
    `dtype` argument).
- **Ultrasound tests** (`test_display`, `test_pipeline_core`,
  `test_wrap_op`): they already guarded their own `cupy` import, but
  still failed at collection because the `ultra_post.*` modules they
  import pull in CuPy. They now skip at module level when CuPy (or
  Holoscan, for `FuncOp`) is unavailable, before importing those
  GPU-only modules.

## Testing

- `ruff check` / `ruff format --check` on all changed files.
- `python -m compileall -q conftest.py`; `python -c "import conftest"`
  succeeds with CuPy unimportable.
- `pytest` with CuPy unavailable:
  - NumPy-backed `mock_image` (uint8 + float32) passes; CuPy-backed
    `mock_image` skips cleanly.
  - `video_streaming` server/client tests: collect and skip (89 skipped)
    instead of erroring at collection.
  - ultrasound `test_display`/`test_pipeline_core`/`test_wrap_op`: 3
    modules skip at collection instead of erroring.

## Out of scope

A few GPU-operator/app test modules (`pixelator`, `sam2`, `vila_live`)
still fail collection without CuPy because they transitively import CuPy
through the operator/application code under test. Those are a separate,
larger cleanup and are not addressed here.

## Summary by CodeRabbit

* **Tests**
  * Test suites now lazily check optional GPU and core dependencies and skip affected modules when unavailable, improving reliability across environments.
  * Image test fixtures use string-based dtype defaults, defer GPU backend import until needed, and ensure generated floating-point data is cast to the requested dtype.

### DIFF
--- a/applications/ultrasound_postprocessing/tests/test_display.py
+++ b/applications/ultrasound_postprocessing/tests/test_display.py
@@ -17,21 +17,22 @@ from __future__ import annotations
 
 import unittest
 
+import pytest
+
 try:
     import cupy as cp  # type: ignore
-except ImportError:  # pragma: no cover - optional dependency
-    cp = None  # type: ignore
 
-from ultra_post.core.display import (
-    DisplayCompressionSettings,
-    compress_color,
-    compress_grayscale,
-    ensure_rgba,
-    tensor_to_display,
-)
+    from ultra_post.core.display import (
+        DisplayCompressionSettings,
+        compress_color,
+        compress_grayscale,
+        ensure_rgba,
+        tensor_to_display,
+    )
+except Exception as exc:  # pragma: no cover - optional GPU dependency
+    pytest.skip(f"CuPy is unavailable: {exc}", allow_module_level=True)
 
 
-@unittest.skipIf(cp is None, "CuPy is required for display tests.")
 class DisplayTests(unittest.TestCase):
     def test_ensure_rgba_shapes(self) -> None:
         gray = cp.ones((2, 3), dtype=cp.float32)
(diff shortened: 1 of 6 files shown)


wyli:
Superseded by #1592, which contains the same change rebased into a single signed-off commit (fixes the DCO and copyright-header CI failures here) and starts with a clean review history.

PR #1626, opened 2026-06-24
Module template: add pre-commit lint configuration

tbirdso:
## Overview

Update the Holoscan Module template so that linting tools are available and configured out of the box for Holoscan CLI.

## Details

Add a pre-commit configuration so a freshly generated module is lint-clean and stays that way:
    
- .pre-commit-config.yaml wiring ruff (lint + format), codespell, cpplint,
        markdownlint, cmakelint, basic hygiene hooks, and a copyright-header check
- .markdownlint.yaml / .markdownlintignore, codespell allowlists
- .github/workflows/scripts/{check_copyright.py,gitutils.py,copyright_excludes.txt}
    
The config is tailored to a standalone module: it drops the HoloHub-monorepo
    excludes carried by the source config, uses ruff (the template's existing
    Python toolchain) instead of black/isort, excludes the vendored cmake/ helpers,
    and disables cpplint's line-length check since clang-format (ColumnLimit 100)
    owns width.

### Additional Changes

- Fix the metadata schema check so that it uses the Holoscan CLI vendored schema files rather than the HoloHub `utilities/` path
- Ran `./holohub lint` on a newly created module from template and applied fixes back to the template

## Result

`./holohub lint` runs out of the box in a new project and applies _minimal_ fixes (updated copyright header, line length adjustments based on length of module name)
    
Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added template files for pre-commit, markdown linting, codespell, and copyright exclusions to give new projects ready-to-use quality checks.
  * Added automated metadata validation in template CI, plus a reusable validation script for project metadata.
  * Generated module projects now copy required workflow scripts during creation so checks work out of the box.

* **Bug Fixes**
  * Improved template path exclusions so linting skips the intended generated and template directories.
  * Refined several template and documentation formatting details for cleaner generated output.


codespell_ignore_words.txt
```diff
@@ -7,6 +7,7 @@ assertIn
 ba
 bu
 dne
+DNE
```
bhashemian: I think you have already replaced DNE with does not exist, right?

modules/template/{{cookiecutter.module_repo_name}}/.pre-commit-config.yaml
bhashemian, line 1: Isn't there any way [in cookiecutter] to just copy ``.pre-commit-config.yaml‎` on the root, and modify the license? or maybe just create a simlink here with the same license? Thanks

codespell_ignore_words.txt
tbirdso: Good catch — you're right. The  entry was added redundantly since  (lowercase, line 9) already covers it, and the original comment that used 'DNE' was already updated to 'does not exist' in this PR. Removed in the latest commit.

modules/template/{{cookiecutter.module_repo_name}}/.github/workflows/scripts/validate_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+registry (operator, application, module, package, …).  The correct
+schema is chosen automatically from the top-level envelope key.
+"""
+
+import json
+import os
+import sys
+from pathlib import Path
+
+_EXCLUDE_DIRS = {"build", ".local", ".cache", "_CPack_Packages"}
```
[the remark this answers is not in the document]
tbirdso: Fixed in the latest commit — added `"install"` to `_EXCLUDE_DIRS`.

modules/template/{{cookiecutter.module_repo_name}}/.github/workflows/scripts/validate_metadata.py
```diff
@@ -0,0 +1,74 @@
...
+        sys.exit(1)
+
+    repo_root = Path(__file__).resolve().parents[3]
+    failed = False
+
+    for path in sorted(iter_metadata_files(repo_root)):
+        rel = path.relative_to(repo_root)
+        try:
+            data = json.loads(path.read_text())
+        except json.JSONDecodeError as exc:
```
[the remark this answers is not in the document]
tbirdso: Fixed in the latest commit — added `encoding="utf-8"` to `path.read_text()`.

modules/template/{{cookiecutter.module_repo_name}}/.pre-commit-config.yaml
tbirdso: The template config is intentionally distinct from the root `.pre-commit-config.yaml` — it drops the HoloHub monorepo excludes, uses `parents[3]` for schema path resolution (relative to the standalone module layout), disables cpplint line-length (since clang-format owns width), and uses ruff instead of black/isort. A symlink wouldn't work since the generated module is standalone (no HoloHub root available), and a straight copy of the root config would need significant editing to work in a generated module. So a purpose-built template config is the most maintainable approach here.

tbirdso: To your initial point -- yes, there is a post-install hook script that we use for copying several files from HoloHub verbatim or with minor tweaks, this file is intentionally tailored for modules here

bhashemian: That makes sense, but it means that we need to update the precommit hooks in two locations. This isn’t a blocker for this PR, but perhaps we could use the same hooks for both and figure out how to define the exclude paths to make them portable.

bhashemian (approved):

tbirdso:
Rebased on `main` and fixed merge conflicts

PR #1652, opened 2026-07-09
Add unified top navigation for NVIDIA Holoscan websites

jjomier:
This commit introduces a new `tabs.html` file that provides a consistent top navigation menu for the NVIDIA Holoscan GitHub Pages sites. The menu is designed to be identical across both the main site and HoloHub, ensuring a seamless user experience. It includes links to key sections such as the Holoscan SDK, HoloHub applications, and support resources, with dropdowns for additional options.

The navigation is styled to enhance usability and accessibility, with features like dropdown panels and active link highlighting.

## Summary by CodeRabbit

* **New Features**
  * Added a unified top navigation tab bar across the website.
  * Introduced dropdown submenus for top-level tabs, with clear active-state highlighting.
  * External links in the navigation now open safely in a new tab and show an indicator.
  * Updated navigation styling for smoother dropdown behavior and improved layout.

PR #1653, opened 2026-07-09
Refactor website menu into shared partial and add mobile drawer nav

jjomier:
Follow-up to the unified top navigation (#1652):

- Extract the menu into overrides/partials/hs-nav.html as the single source of truth, consumed by both the desktop tab bar (tabs.html) and the mobile drawer (nav.html).
- Rearrange the menu to: Home, Holoscan SDK, HoloHub, Holoscan Sensor Bridge, Support, Blog (with Documentation/GitHub, NVAIE/Forum submenus).
- Add a native, expandable mobile drawer that mirrors the desktop menu; the injected items are drawer-only so they don't duplicate on the desktop sidebar.
- Keep HoloHub's own deep applications/operators tree in the mobile drawer via the hs_drawer_native_nav flag, below the shared menu.
- Fix desktop dropdowns being clipped by Material's overflow/contain on the tabs.

## Summary by CodeRabbit

* **New Features**
  * Added a unified Holoscan navigation menu across the documentation site.
  * Improved mobile navigation by showing the shared menu in the drawer while keeping the site’s own section tree below it.

* **Style**
  * Updated navigation styling for cleaner mobile-only visibility and a more subtle divider between menu sections.
  * Refined desktop tab/dropdown navigation behavior and presentation.

* **Bug Fixes**
  * Made navigation state more accurate so the current page is highlighted consistently.


bhashemian (approved):
Thanks for these fixes! LGTM!

